PR #17, opened 2015-10-09
Fix: a keyword may have duplicated keys so it is not strictly a dict

Tica2:
This is from Keyword help module:

A keyword may have duplicated keys so it is not strictly a dictionary. However
most of the functions in this module behave exactly as a dictionary and mimic
the API defined by the Dict behaviour.


doomspork:
Good catch!  Thank you for the contribution @Tica2 :+1:

PR #18, opened 2015-10-30
Mostly grammar/spelling error fixes.

jinyeow:
As I was going through elixir-school I noticed a few grammar/spelling mistakes. I the project really useful, so I decided to do what I could to contribute and fixed those errors I could find. I also added a note that the booleans true and false are also atoms in the Basics lesson. Just as an FYI I suppose.


doomspork:
Awesome @jinyeow, thank you so much!  Contributions are always welcome.  I'm actually working on some more lessons so this is great motivation!

PR #29, opened 2015-12-02
Fix TOC of basics/basics.md

3100:
Hi, the order of 'table of contents' in `basics/basics.md` is not the same as its contents.


doomspork:
Thanks for catching that @3100 :+1:

PR #36, opened 2015-12-10
Fix some typos.

3100:


doomspork:
Thanks @3100, I need to do a better job of spell checking! :+1:

PR #40, opened 2015-12-13
Fix a small markdown error on basic/enum.md

tiagoamaro:
Hi! This PR is quite simple. Its goal is to fix up a small markdown formatting error:

## Before

before

## After

after

> PS: Thank you very much for this tutorial! It's an awesome resource to learn Elixir!


doomspork:
Thanks @tiagoamaro, much appreciated :+1:

PR #45, opened 2015-12-16
fix composition samples

upamune:


doomspork:
Thanks again @upamune :+1:

PR #56, opened 2015-12-28
Fix typo for supervisor/3

milmazz:


doomspork:
I need to go back over my earlier lessons with a spellchecker it seems :grinning: 

Thanks @milmazz :+1:

PR #58, opened 2015-12-29
Use Ecto.Schema instead of Ecto.Model

milmazz:
This PR include the following improvements:

* Fix the anchor links in the ToC section
* Include new anchor links in the ToC section
* Fix some typos
* Include the key `hostname` in the `config/config.exs` sample
* Update the sample in the *Model* section, now we use `Ecto.Schema` instead of `Ecto.Model`, the latter is deprecated
* Include the import for the `ExampleApp.{Repo,User}` modules
* Update the samples for the *Basics*, *Group by* and *Order by* subsections
* Replace the `created_at` field for `inserted_at`, which is produced by timestamps
* Replace `phoenix-examples/ecto_query_library` link for the official reference of the `Ecto.Query.API`


doomspork:
Thank you @milmazz :+1:

PR #66, opened 2016-01-16
fix small typo

tiagoamaro:
Don't know if it was a typo, or it was wanted to say that `Mix is a smart build tool...`. Just removed the `a` for brevity, but if you prefer the later, I'll update the PR :)


doomspork:
Thank you @tiagoamaro :+1:

PR #68, opened 2016-01-17
Correct string interop section

### LINKED ISSUE #67: Advanced section, lesson 1 - Strings
The section says that Erlang strings are represented with single quotes. This is unfortunately incorrect. Strings in Erlang are represented with double quotes, the same as in Elixir.
 
As for single quotes, they are used to represent atoms that contain spaces, uppercase letters or any other characters that are "unconventional" for atoms. 

You can find the documentation for it [here](http://www.erlang.org/doc/reference_manual/data_types.html#id64703).

This part needs be updated.

doomspork:
Thanks again @viktorot.  How does this sound to you now?

Closes #67

### DIFF
--- a/lessons/advanced/erlang.md
+++ b/lessons/advanced/erlang.md
@@ -81,18 +81,30 @@ example.
 
 ### Strings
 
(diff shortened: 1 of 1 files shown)


viktorot:
Great. Glad to see a fast fix :)

Cheers.

doomspork:
Thanks for bringing it to my attention :+1:

PR #78, opened 2016-01-25
fix: update the link to Chinese version

zhyu:


doomspork:
Not sure how I screwed that up :disappointed:

Thanks @zhyu! :+1:

PR #83, opened 2016-02-02
New Error Handling lesson

doomspork:
Feedback welcomed.

### DIFF
--- a/lessons/advanced/error-handling.md
+++ b/lessons/advanced/error-handling.md
@@ -1,0 +1,158 @@
+---
+layout: page
+title: Error Handling
+category: advanced
+order: 7
+lang: en
+---
+
+Although more common to return the `{:error, reason}` tuple, Elixir supports exceptions and in this lesson we'll look at how to handle errors and the different mechanisms available to us.
+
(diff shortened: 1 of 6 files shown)


lessons/advanced/error-handling.md
```diff
@@ -0,0 +1,160 @@
...
+  |> File.read!
+rescue
+  e in KeyError -> IO.puts "missing :source_file option"
+  e in File.Error -> IO.puts "unable to read source file"
+end
+```
+
+## After
+
+Sometimes it may be necessary to perform some action after our `try/rescue`, regardless of whether an error occurrs or not. To handle this Elixir supports `try/after` which can be used inconjuntion with `try/rescue`.
```
doomspork: occurrs

doomspork:
@pjhampton on the topic of re-ordering lessons, what do you think about this one?  I think Error Handling should come before the OTP topics, thoughts?

pjhampton:
This is good stuff! I just read through with no comments.

I think you are right. Running with the assumption that the reader is consuming the lessons from Basic - Specific, 1..n respectively, then this lessons position is Advanced 2 or 3 ~ I don't think which matters but definitely before OTP and Metaprogramming.

doomspork:
Thanks @pjhampton, I went with Adv 2:

image

Issue #50, opened 2015-12-24
Lesson Suggestions


doomspork:
This can serve as a place for folks to suggest new lessons.  

Suggested lessons:

**Advanced:**
- [ ] OTP Architecture
- [ ] Distribution
- [ ] GenStage

**Specifics:**
- [ ] Creating a DSL
- [ ] Leex & Yecc
- [ ] Dialyzer / Debugging (created #566 )
- [ ] Hex Docs

**Libraries:**
- [ ] Guardian (in-progress #431)
- [ ] exrm 
- [ ] Bamboo

doomspork:
Two additional lesson that could either be lessons or just a section in another lesson: sigils and comprehensions

pjhampton:
Do you mean a lesson on existing sigils or creating custom sigils?

doomspork:
I would like to cover both topics actually!  They could certainly be in the same lesson.

pjhampton:
Ok, cool. I can certainly try and tackle this one!

doomspork:
That's great! 

@pjhampton as I've done with other contributors I'll add you to the repo so you can assign issues to yourself and update them.  I do ask that big changes go in via PR so they can be reviewed and we can ping folks doing translations.  Sound good?

Master is deployed on merge.

pjhampton:
Sounds good! I will get started with Sigils this week coming.

doomspork:
I started on "Creating a DSL" lesson but I'm not entirely sure _what_ DSL to build.  Most of the existing examples create new types of assertions but I'd like to do something different.  Any thoughts @pjhampton?

pjhampton:
@doomspork I will get back to you on this once I have done some thinking.

doomspork:
Go for it @pjhampton :grinning:

pragmaticivan:
Are `nodes` inside `Distribution`? Should we have a lesson just for `nodes`? I think it's a good one to have after processes.

pjhampton:
Hey @pragmaticivan. Nodes were mentioned in `Mnesia` which I wasn't happy about, but I needed to give context as it is a distributed database. But yeah, from what knowledge I have, nodes fall under distribution.

doomspork:
I'm going to close this issue in favor of conversations we have in Slack.

ybur-yug:
@doomspork do you want to reopen this per our discussion earlier today or should I just join the slack group?

doomspork:
[...]

PR #85, opened 2016-02-03
Adding Sigils lesson.

pjhampton:
As discussed in #50. Let me know if there is any issues and I will get on them.


doomspork:
This looks good @pjhampton :+1: Thanks again

PR #94, opened 2016-02-18
Vietnamese Translation

joneslee85:
Xin chào!

It will take few more weeks until Elixir Vietnam usergroup can translate all pages to Vietnamese. For the time being, I'd like to suggest to merge for whatever we have.


doomspork:
This great, thank you very much @joneslee85 :+1:

Issue #89, opened 2016-02-07
Portuguese Translation: Basic Lessons 🇧🇷


tiagobarreto:
- [x] index.md
- Basics
 - [x]  basics.md
 - [x]  collections.md
 - [x]  enum.md
 - [x]  pattern-matching.md
 - [x]  control structures.md
 - [x] pipe-operator.md
 - [x]  functions.md
 - [x]  composition.md
 - [x]  mix.md
 - [x]  sigils.md
 - [x]  documentation.md
 - [x]  testing.md

pragmaticivan:
Update this?

doomspork:
@pragmaticivan if you're interested in continuing your contributions, I'd be glad to add you to the repo so you can help with these things.  Only request is that big changes go through PR so there's an opportunity for comment (24-48 hours).  Thanks again!

doomspork:
Updated :+1:

pragmaticivan:
I do have plans in continuing my contributions.

doomspork:
Great, I will add you now.  Thank so much @pragmaticivan :grinning:

pragmaticivan:
Just for tracking, I'm working on testing.md. The last basic one.

pragmaticivan:
Basic is ready :D Can you tweet about the basic section in Portuguese @doomspork Would like to attract more Brazilians to contribute to this project.

doomspork:
@pragmaticivan I will do that shortly!  Thanks @pragmaticivan

pragmaticivan:
WIP: pipe-operator.md (@pragmaticivan)

pragmaticivan:
The last one is in #126 .

pragmaticivan:
Close this one as well? @doomspork

doomspork:
:+1:

Issue #71, opened 2016-01-19
Translation version control


cizixs:
Right now I am translating into tutorial into Chinese, and find out a issue: **How to keep translation up to date?**

There are updates on previous lessons, which makes my translation out of date. The obvious way to solve this is to go through commit logs and find out the updated content, then add another patch.

Do you guys have a better way to work on this one?

doomspork:
@cizixs this is a great question!  I don't have a good solution so I'm open to any suggestions you have.  So far I've been trying to keep up with changes in #49 for @3100 but I don't know if that's working out well.

As the folks handling the translations, what would make it easiest on you?

ping @milmazz, @erickgnavar, @3100

milmazz:
@doomspork I think it's really hard to keep synchronized all the content, for example, in some translations of the official pages about [Debian](https://www.debian.org/) they offer a note that says something like:

> Note: The original document (link to the original document in English) is newer than this translation. 

So, with this in mind, we can set a new variable in the frontmatter, let's say `version` and the link to the original document in English, if both versions differ then we put the note in the translated content.

By the way, I used [Transifex](https://www.transifex.com/) a long time ago in other open source projects, looking at their page they now offer something called *Transifex Live*, and in this [section](https://www.transifex.com/pricing/), they said that they support open source projects for free in their new pricing schema.

doomspork:
I like the idea of adding a note to the translated versions.  Great suggestion @milmazz.

I'll take a look at Transifex later this week.

doomspork:
@milmazz do you recall which other OSS projects used Transifex?  I want to see how other people are using it.

milmazz:
@doomspork Right now I recall [Django](https://www.transifex.com/django/django/) and [Turpial](https://github.com/satanas/Turpial) (a Twitter client written in Python). But both projects use Transifex to offer translations for their source code base, not for their sites. Maybe we need to find better examples for our use case.

doomspork:
I think our current approach with GitHub issues is working.  We can investigate an alternative service/product/solution if we continue to grow and feel pains keeping synced.

Thanks for all the help guys! :+1:

PR #98, opened 2016-02-19
Russian translation

brain-geek:
I've started translating the project to russian. Here is the first lesson.


pjhampton:
Awesome! Thanks @brain-geek :+1: 

It may be helpful to create an issue like #89 so we can all follow the progress. There is a lot of translations happening simultaneously.

brain-geek:
@pjhampton Thanks for idea! 

I'll create one.

pjhampton:
Maybe @joneslee85 this approach might be useful for the Vietnamese lessons as well? It might be useful to establish this protocol across all languages. What you think?

doomspork:
Take a peek at #71 if you guys don't mind, any suggestions on how to manage these different translations would be greatly appreciated!

PR #103, opened 2016-02-20
Enhance Colletion page [vi]

joneslee85:
There are many English terms that could not be 1:1 translated to Vietnamese. Most of the cases, translated terms make no senses whatsoever. Because most of Vietnamese developer uses English terms in their conversations, I don't see why we have to introduce awkward translated Vietnamese terms. I'd like to propose to keep the English word but introducing Vietnamese explanation clauses.

### DIFF
--- a/vi/lessons/basics/collections.md
+++ b/vi/lessons/basics/collections.md
@@ -6,28 +6,28 @@ order: 2
 lang: vi
 ---
 
-Danh sách (list), tuple, danh sách từ khoá (keyword), map, từ điển (dict) và sự kết hợp hàm số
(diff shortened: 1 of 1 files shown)


vnnoder:
Agree with these changes. This seems to be more practical.

For those english terms that come with vietnamese explanation, should we only show explanation the first time the term appears in a lesson? I think most reader will read from the beginning to the end of a lesson and is already familiar with the term.

joneslee85:
@vnnoder 

> should we only show explanation the first time the term appears in a lesson?

Yes and No, I depends on the concept, some concepts are common in programming but some are unique in Elixir. Thus IMHO your suggestion would go well with the first.

joneslee85:
@trexnix thanks much for the comments, I've updated the PR

PR #114, opened 2016-02-23
Fixed 'specifics' numbering to reflect 'en' lesson changes

pjhampton:
There were only two languages that needed updated (en & vi) regarding the specifics lessons. The other languages are yet to add any content to this section.


doomspork:
Thank you @pjhampton

PR #117, opened 2016-02-25
Pipe Operator Lesson, Reordered Basic Lessons

pjhampton:
- Added Pipe Operator Lesson for Review
 - Reordered basic lessons

### DIFF
--- a/lessons/basics/composition.md
+++ b/lessons/basics/composition.md
@@ -2,7 +2,7 @@
 layout: page
 title: Composition
 category: basics
-order: 7
+order: 8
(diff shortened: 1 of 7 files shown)


doomspork:
This is great @pjhampton.  Do any of the other languages need to be re-ordered?

pjhampton:
I imagine so @doomspork. Would you prefer if I baked it into this PR or a separate one?

doomspork:
Could you add another commit to this one with those changes?  Then it'll all deploy together.

Issue #118, opened 2016-02-26
Best Practices in Testing


pragmaticivan:
If there's no `Best Practices` content, why are we keeping this area in `testing.md` ? should we remove until someone send this data?

doomspork:
That's a good point @pragmaticivan.  @pjhampton any thoughts on whether we should come up with some content or remove this?  If either of you have suggestions for topics I can sit down and work on it.

pjhampton:
@pragmaticivan @doomspork let's remove it _for now_ and we'll come back to this later. I say this because we discussed adding Best Practices / FAQs at the end of each lesson.

doomspork:
That sounds good to me.  Do either of you want to tackle removing it across the languages?  Please use a PR.

pjhampton:
@doomspork @pragmaticivan On it! :smile:

pragmaticivan:
I think we can close this one? right @doomspork ?

doomspork:
Correct @pragmaticivan :+1:

PR #119, opened 2016-02-27
Reordered lessons / Removed Best Practices

pjhampton:
References #118

### DIFF
--- a/cn/lessons/basics/composition.md
+++ b/cn/lessons/basics/composition.md
@@ -2,7 +2,7 @@
 layout: page
 title: 组合
 category: basics
-order: 7
+order: 8
 lang: cn
 ---
 
--- a/cn/lessons/basics/documentation.md
+++ b/cn/lessons/basics/documentation.md
@@ -1,0 +1,20 @@
+---
+layout: page
+title: Documentation
+category: basics
+order: 11
+lang: cn 
+---
+
+Documenting Elixir code.
+
+## Table of Contents.
+
+- [Annotation](#annotation)
+  - [Inline Documentation](#inline-documentation)
+  - [Documenting Modules](#documenting-modules)
+  - [Documenting Functions](#documenting-functions)
+- [ExDoc](#exdoc)
+  - [Installing](#installing)
+  - [Generating Documentation](#generating-documentation)
+- [Best Practice](#best-practice)
--- a/cn/lessons/basics/functions.md
+++ b/cn/lessons/basics/functions.md
@@ -2,7 +2,7 @@
 layout: page
 title: 函数
 category: basics
-order: 6
+order: 7
 lang: cn
 ---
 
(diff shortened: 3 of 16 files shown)


doomspork:
Hi @joneslee85, @pragmaticivan, @jesusangelm, and @cizixs!  First and foremost, thank you all for your help with translations.  In an effort to keep everyone in the loop, I wanted to call your attention to this PR.  It re-orders the basic lessons in your languages and introduces a new Sigil lesson.

Any questions or concerns with these changes?

pragmaticivan:
My considerations about the PR: the translations where we don't have the `pipe` lesson translated seems to be wrong, and also you added a file for the `pipe lesson` without translated data inside the folder `/pt`

The Best Practices update seems to be solid.

Not sure how we supposed to solve this order problem in case we don't have the `pipe` lesson.

pragmaticivan:
What about instead of using the order from the list here, just using a count variable?

```

doomspork:
@pragmaticivan if you want to test out that change and open a PR I'd be glad to look at it, sounds like a solid suggestion :+1:

doomspork:
@pragmaticivan do you want to remove the empty pt/pipe lesson for now?  I believe some of the other translators were using English placeholders which may have been @pjhampton's rationale.

pjhampton:
I agree with @doomspork, @pragmaticivan. The problem is translators from different languages have different ways of doing things. For example, the vietnamese translators duplicate all the english lessons and translate over them. However, if you would like an exception for pt, feel free to let me know :smile_cat:

pragmaticivan:
In this case I would suggest having a flag such as ` to be translated` with a link for this file on github, or something like that.

doomspork:
That's an interesting idea @pragmaticivan.  I think a placeholder with a notice that the file needs translation and a link to the repo would be a good solution.  Thoughts @pjhampton?

doomspork:
@joneslee85 you've used English lesson placeholders, do you have any thoughts?

pjhampton:
I'm impartial guys to be honest as I don't really ever deal with the translations, just new English lesson content. 

It may be best to let the invisible hand guide this one, and let the language translators work to the workflow they wish within their language, rather than unifying an approach? If we are creating an approach it could be good to update
https://github.com/doomspork/elixir-school/wiki/Translating - with a more refined setup process for a new language.

A placeholder statement on an English lesson might be a good idea, but again, I'm neutral on this enhancement.

doomspork:
Alright.  @pjhampton for the time being will you remove the pt lesson?  We can leave the others and make changes once we incorporate the count change that @pragmaticivan suggested.

pjhampton:
@pragmaticivan @doomspork: I reverted any changes to pt. Let me know if any issues.

doomspork:
Looks good to me @pjhampton.  @pragmaticivan unless you have any addition thoughts I'm going to merge this :grinning:

pragmaticivan:
@doomspork go for :D We're going to add the last one on basic, so we can update the order again.

PR #131, opened 2016-02-29
Changed erl func representation to ex style for consistency.

pjhampton:
Minor update to prevent any possible confusion from any readers.

### DIFF
--- a/lessons/specifics/mnesia.md
+++ b/lessons/specifics/mnesia.md
@@ -109,7 +109,7 @@ iex> Mnesia.start()
 :ok
 ```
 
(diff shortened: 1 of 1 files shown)


doomspork:
@pjhampton what about the examples for `Mnesia.dirty_read/1`, they're using `:mnesia`?

pjhampton:
Jeez, well spotted @doomspork! I think the consistency is quite important in this lesson as we are using an alternative syntax to the traditional `:erlang.func/x` representation in Elixir.

doomspork:
:+1:

PR #134, opened 2016-03-02
Fix pluralization in the 'Interactive' section of 'Mix'.

timlew:
This just replaces 'loads' with 'load'.


doomspork:
Thanks @timlew :+1:

PR #137, opened 2016-03-03
Fixed navigation in russian 'collections' translation.

brain-geek:


pjhampton:
@brain-geek LFTM :+1: 

Before I merge this, is there any others that need to be changed as well?

brain-geek:
@pjhampton Navigation is OK in russian.

But looks like we need to have glossaries for each translation. That's due to lot of terms we usually don't use in native language e.g. 'tuple', 'keyword list', etc.

In that way it would be easier for new translators to join and it'll improve overall translations quality.

doomspork:
@brain-geek could you elaborate on your glossary idea?

Issue #96, opened 2016-02-18
Better mobile menu


KaoruDev:
![image](https://cloud.githubusercontent.com/assets/5273755/13156753/e3229d48-d638-11e5-9d33-f1a09c86bf36.png)

it's frustrating to be viewing this on a phone and see the massive navbar take up the entire page.

would be nice to have a slide menu ( responsive only )

https://codyhouse.co/gem/3d-bold-navigation/
might be cool.

doomspork:
Closing this for now.  We've collapsed the Advanced and Specific sections by default and added additional styles to make the links easier to see.

PR #141, opened 2016-03-05
add mobile styles to nav

jessejanderson:
image

### DIFF
--- a/_assets/stylesheets/custom.scss
+++ b/_assets/stylesheets/custom.scss
@@ -38,17 +38,26 @@ a.sidebar-lesson:focus {
 }
 
 .sidebar-nav-item {
-  padding-left: 5px;  
+  border-radius: 3px;
+  margin: 0 0 5px;
+  text-align: left;
+  background: rgba(255, 255, 255, 0.05);
+  padding: 0 15px;
 }
 
 .sidebar-nav-group {
-  padding-bottom: 10px;
+  padding-bottom: 0;
 }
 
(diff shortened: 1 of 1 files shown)


jessejanderson:
Not sure the best way to do this, but it would be great if the navigation was toggled close by default on mobile.

doomspork:
I agree, the menu is getting a bit long these days.  Any suggestions?  One option might be to have all sections closed by default for everyone.

doomspork:
Collapsing the menu may a good solution to #96

doomspork:
@pjhampton and @pragmaticivan, what do you think about collapsing the menus by default?  Right now they're a bit long, this might serve to buy us some time building a whole new site.

Thoughts?

pjhampton:
This looks swell. 

**All devices** - Maybe it could paginated? I say this because I think it maybe a good idea to show at least the basic lessons in the menu my default, until maybe they move to another section? 

**Only mobile** - If this just applies to mobile, then I think that is good :+1: 

thanks @jessejanderson

pragmaticivan:
I like the new just mobile menu, just tested here locally, but would collapse the menu and just show the basics area. And also need to detect when an advanced area is active, so it needs to show the advanced area opened instead the basic area.

doomspork:
Thanks for your thoughts @pragmaticivan and @pjhampton.  I'm going to merge this as-is but I'm open to a PR to improve collapsing the menus.

Thanks again @jessejanderson this is great :+1:

PR #149, opened 2016-03-12
Removed undefined password struct field.

inchingforward:


pjhampton:
You're right! Well spotted @inchingforward :+1:

doomspork:
Great catch :+1:

PR #151, opened 2016-03-12
Update README - Instructions for description data

pragmaticivan:


pragmaticivan:
cc @doomspork

doomspork:
Looks good @pragmaticivan, thanks! :+1: 

Can you also make this change to the Wiki?  I'd like to eventually direct everyone there and reduce the README down to a table of contents.

pragmaticivan:
Done @doomspork :)

pjhampton:
@pragmaticivan @doomspork 

Maybe we could should send a broadcast out to all the languages devs? I just realised the only lang where the description is translated is **pt**! Great job btw

pragmaticivan:
I also think that's a good time to think about organization. Would be really coo to mention `team/portuguese` in a Pull Request Revision. And in this case a Broadcast such as `language_managers` etc...

pjhampton:
Maybe we're too small for an org right now @pragmaticivan ? ~289 stars isn't an awful lot. @doomspork do you have any analytics on traffic from either google or heroku?

doomspork:
@pjhampton there are analytics.  We can discuss them in Slack if you message me.

I'm thinking about the move to an org.  I'll make a decision and let everyone know.  There's pros and cons to both approaches.  I appreciate that people care enough to have the conversation at all :grinning:

PR #157, opened 2016-03-13
Changed sidebar language links to be relative.

brain-geek:
Right now there is a very confusing behaviour when running the project locally - when you change language via link in the sidebar, you are redirected to the https://elixirschool.com instead of your local version of the site.

This PR should not change anything in production, but will simplify usability for translators a bit.


doomspork:
Thanks @brain-geek :+1:

Issue #162, opened 2016-03-14
Glossary proposal


brain-geek:
The problem: right now there are some wording issues related to Elixir-specific (or other non-widely used) terms when translating. Also, there is a problem whether some things should be translated at all. And I'm not the only one having this problem:

> There are many English terms that could not be 1:1 translated to Vietnamese. Most of the cases, translated terms make no senses whatsoever. 

And there's even small discussion [in this topic](https://github.com/doomspork/elixir-school/pull/103).

> Note: I couldn't find a good translation for callback, so I leave as it is but in italic.
> Note 2: The word handler is not easy to translate, so I used the word "processador" which is something similar.

[Source](https://github.com/doomspork/elixir-school/pull/153)

Also, for all terms, we should have stable translations in one language to improve reading experience.

A possible solution for this problem is having some kind of glossary per language. [Example I made for Russian language](https://github.com/brain-geek/elixir-school/blob/ru_glossary/ru/GLOSSARY.md)

This will improve translation quality and simplify terminology translation lookups.

Let's discuss this.

pjhampton:
It sounds like a good idea to me, but then again... I'm not a translator. Thoughts @pragmaticivan?

erickgnavar:
Good idea, I have the same problem when I translate terms like "pattern matching" or "function clause" to spanish.

pragmaticivan:
I like the idea, in Portuguese some of the terms are really hard to translate. Some of them I just don't translate because they're common terms in programming.

doomspork:
I'm :+1: for a glossary.  I think it would be useful even for English speakers.

brain-geek:
I've added the first version of glossary in #174 . Feedback is welcome.

setzer777:
Do we have an spanish glossary at this moment?

doomspork:
No we do not @setzer777.  Feel free to check out #174 and leave your feedback/suggestions :grinning:

setzer777:
I think this idea is super great! 
Sometimes we don't have to translate every term into our language, because of two things.

1.  The translation effort grows so bad, because we need to find adequate words to make our translation work.
2.  Specially with the technical terms like Pattern Matching, at least in Spanish, when we try to translate that, it turns to be a really poor translation, and it's not because of the translator, it's because there isn't really a synonym or a single way to make the translation fully match the real term. And this is something I've been claiming a lot in technology translated books, the translated terms are not an easy way to google the real topic, and that makes the process slower and unclear.

[...]

PR #174, opened 2016-03-16
First version of russian glossary

brain-geek:
In the discussion of issue #162 there was a lot of positive feedback about creating glossary. This is the first version of it with all terms I've found yet.

I'm not sure whether it should be separate file per language or one file for everything in the root folder - that needs to be discussed.


doomspork:
This is awesome @brain-geek.  Is it possible for you to squash some of these commits?

Do you think this is strictly useful for translators or would this be something a reader might want to peruse?  If the latter, we should figure out a place to link to this from the website.

brain-geek:
@doomspork That's very good question. Actually, readers of translated versions may be interested in what are original terms used. But that is not a major use case, as those experienced people will be reading original english version anyway. 

So, I don't think why would we link it on the site.

I'll squash last commits.

doomspork:
Sounds good @brain-geek :+1:

doomspork:
@pragmaticivan and @pjhampton, any thoughts?

pjhampton:
This LGTM :+1:. Can we change this PR to reflect that it is the Russian Glossary in the title?

doomspork:
Thinking about this a little more, I want to offer some suggestions and see what everyone thinks.

First thing, do we need this under source control?  If not, this might be better in the Wiki.  There we can setup individual glossaries for each language.

If we do need it under source control, should we be maintaining one large glossary for all languages or individual files per language?  (eg `glossary.md` -> `/glossary/ru.md`)

pragmaticivan:
Once it will be always one page, maybe we can "create an example" of how it would fit there first.

brain-geek:
> First thing, do we need this under source control? 
As for me, that makes sense under version control due to the fact that better term translations pop up - and this leads to glossary change.

Some terms (like 'pattern matching') changed their russian translation 3-4 times before getting to current state.

> If we do need it under source control, should we be maintaining one large glossary for all languages or individual files per language?
That's a good question and I don't know an answer to it. My suggestion is to start with global file and see how it goes.

brain-geek:
> Can we change this PR to reflect that it is the Russian Glossary in the title?

Done

> Is it possible for you to squash some of these commits?

Done

doomspork:
@brain-geek sounds good.  We'll start with this glossary and go from there.

Thanks again!

PR #180, opened 2016-03-19
Fix anchor

philss:
Anchors needs quotes. 

It fixes this page: https://elixirschool.com/pt/lessons/advanced/umbrella-projects/


pragmaticivan:
Nice catch @philss , thanks!

PR #183, opened 2016-03-20
Add a section on `with`

doomspork:
RFC @pjhampton & @pragmaticivan (and anyone else) — Added a section on `with` to the control structure lesson.

### DIFF
--- a/lessons/basics/control-structures.md
+++ b/lessons/basics/control-structures.md
@@ -13,6 +13,7 @@ In this lesson we will look at the control structures available to us in Elixir.
 - [`if` and `unless`](#if-and-unless)
 - [`case`](#case)
 - [`cond`](#cond)
+- [`with`](#with)
 
 ## `if` and `unless`
 
@@ -100,7 +101,6 @@ iex> case {1, 2, 3} do
 
 Check the official docs for [Expressions allowed in guard clauses](http://elixir-lang.org/getting-started/case-cond-and-if.html#expressions-in-guard-clauses).
 
-
 ## `cond`
 
 When we need to match conditions, and not values, we can turn to `cond`; this is akin to `else if` or `elsif` from other languages:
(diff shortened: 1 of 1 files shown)


lessons/basics/control-structures.md
```diff
@@ -128,3 +128,36 @@ iex> cond do
...
+%{first: "doomspork"}
+iex> with {:ok, first} <- Map.fetch(user, :first),
+...>      {:ok, last} <- Map.fetch(user, :last),
+...>      do: last <> ", " <> first
+:error
+```
+
+Now let's look a larger example without `with` and then see how we can refactor it:
+
+```elixir
```
doomspork: I need a good complex example.

lessons/basics/control-structures.md
```diff
@@ -128,3 +128,36 @@ iex> cond do
 ...> end
 "Catch all"
 ```
+
+## `with`
+
+The special form `with` is useful when you might us a nested `case` statement or situations that cannot cleanly be piped together. We'll start with a simple example of `with` and then look at something more significant:
```
pragmaticivan: `might use a nested`

doomspork: Good catch :+1:

lessons/basics/control-structures.md
```diff
@@ -128,3 +128,49 @@ iex> cond do
...
+%{first: "doomspork"}
+iex> with {:ok, first} <- Map.fetch(user, :first),
+...>      {:ok, last} <- Map.fetch(user, :last),
+...>      do: last <> ", " <> first
+:error
+```
+
+Now let's look a larger example without `with` and then see how we can refactor it:
+
+```elixir
```
doomspork: @pjhampton / @pragmaticivan what do you think of this example?

lessons/basics/control-structures.md
```diff
@@ -128,3 +128,49 @@ iex> cond do
...
+```
+
+In the event that an expression fails to match, the non-matching value will be returned:
+
+```elixir
+iex> user = %{first: "doomspork"}
+%{first: "doomspork"}
+iex> with {:ok, first} <- Map.fetch(user, :first),
+...>      {:ok, last} <- Map.fetch(user, :last),
+...>      do: last <> ", " <> first
```
pragmaticivan: correct if I'm wrong, but the expression fails to match here should not just have `first`? where is the `last` coming from ?

lessons/basics/control-structures.md
```diff
@@ -128,3 +128,49 @@ iex> cond do
...
+:error
```
pragmaticivan: I see we get an error here but would be good showing also an "real alternative without errors"

lessons/basics/control-structures.md
doomspork: I'm not sure I follow.  Have you run the example?

lessons/basics/control-structures.md
doomspork: That's the examples below.

pragmaticivan: Sorry, my bad, I executed but was expecting a different concept for not having the error, thanks.

lessons/basics/control-structures.md
```diff
@@ -128,3 +128,49 @@ iex> cond do
 ...> end
 "Catch all"
 ```
+
+## `with`
+
+The special form `with` is useful when you might use a nested `case` statement or situations that cannot cleanly be piped together. We'll start with a simple example of `with` and then look at something more significant:
```
pragmaticivan: What about:

```
The special form `with` is useful when you might use a nested `case` or situations that cannot cleanly be piped together. It's quite similar to `for` comprehensions. It retrieve the value returned by the right-side of `<-` and uses pattern match on the left side. If the value matches the pattern, `with` moves on to the next expression. This way we can identify more than one expression in order to perform the result. We'll start with a simple example of `with` and then look at something more significant:
```

doomspork: Thanks @pragmaticivan!  I'm going to add more to this section based on your suggestion.  I want to refrain from referencing `for` here though since it has not been covered yet in the lessons by this point.

doomspork:
I updated some of the wording in the `with` section, any thoughts?

ping @pragmaticivan / @pjhampton / @setzer777

PR #228, opened 2016-03-27
Hide sidebar scrool

pragmaticivan:
I created this Pull Request because seems to be annoying the current version looks like we're using an `iframe`. You guys can check, it's just a suggestion.


doomspork:
@pragmaticivan it may take me a few days to get to this PR, I'm currently out-of-town; I want to make sure UX changes are well tested so we don't inadvertently impact UX for others.

doomspork:
Alrighty, had a chance to check this out :+1:

PR #233, opened 2016-03-28
[RU] Added note from #178 to russian translation.

brain-geek:


brain-geek:
Link to the PR from title #178

PR #236, opened 2016-03-28
Mix lesson changes

brain-geek:
Moved 'interaction' section before the 'compilation'.
Removed reference to project compilation requirement to use iex.

Current wording is a bit confusing. ( _With our application compiled we can start a new `iex` session:		_ ) From this one might tell that this is caused by the fact app is compiled, while really it's working without compilation too.

Also, I've moved the 'iex' part before the compilation, because it's very widely used.


setzer777:
:+1:

pjhampton:
LGTM :+1:

PR #267, opened 2016-04-07
Add Heroku app.json

doomspork:


doomspork:
We need this for the review apps to work.

Issue #270, opened 2016-04-09
Remove Line of available languages from index of all languages but english


pragmaticivan:
`Available in Việt ngữ, 汉语, Español, 日本語 and Português.`

It's creating complexity to update that every time a new language comes to the website.

Now we have `en jp cn es pt vi ru sk fr gr it` and it doesn't represents the list in the index.

My suggestion is to have this `Available in Việt ngữ, 汉语, Español, 日本語 and Português.` in the english version.

doomspork:
Thoughts @jesusangelm / @joneslee85 / @setzer777  / @brain-geek and others?

PR #274, opened 2016-04-11
Update Chinese index page

cizixs:


cn/index.md
```diff
@@ -8,6 +8,14 @@ lang: cn
 
 Elixir 基础知识教程，受 Twitter 的 [Scala School](http://twitter.github.io/scala_school/) 启发编写。
 
+目前有多种语言版本： [Việt ngữ][vi], [汉语][cn], [Español][es], [日本語][jp] and [Português][pt].
```
doomspork: Per #270, we're thinking about removing theses languages from all indexes _except_ English.  This way you (nor other translators) need to keep track of need languages and updating your indexes.

  Any thoughts @cizixs?

cizixs:
@doomspork I think this is okay, as long as readers can jump to any translation version from any page. Right now it is available from the bottom-left corner, so remove these languages is convenient for translators.

doomspork:
@cizixs I am going to close since we are removing this from the translated indexes.

Thank you for your help :grinning:

Issue #281, opened 2016-04-13
Design Feedback


jakefolio:
I just wanted to point out a few things I noticed while reading the site:

1. Default link color is the same as text and doesn't have an underline to make it obvious it's a link
2. The site name/logo links to the current page and doesn't bring you back to the homepage.

doomspork:
Hi @jakefolio, thanks for the feedback!  Would you open to contributing fixes for these?  PRs are welcome :grinning:

doomspork:
Closed via #282, thanks again @jakefolio :+1:

PR #282, opened 2016-04-13
Design feedback

jakefolio:
This PR fixes the missing baseurl in the _config.yml file and defaults it to /. While browsing the site, I found it incredibly hard to see links in the content body. I've added an underline to all content body links. 

The only place I've noticed "odd" behavior is when the link contains code block, in which I question whether a code tag should even be in a link.

Issue Reference: #281

### DIFF
--- a/_assets/stylesheets/custom.scss
+++ b/_assets/stylesheets/custom.scss
@@ -9,6 +9,9 @@
 .theme-custom-purple .related-posts li a:hover {
   color: #4A365F;
 }
+.theme-custom-purple .content a {
+  text-decoration: underline;
+}
 
 h2 {
  color: #F07E6B;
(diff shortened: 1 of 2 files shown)


doomspork:
This is great, thank you @jakefolio!  Improvements to the UI/UX are always welcomed.

@pragmaticivan / @pjhampton — any thoughts before I merge?

pjhampton:
LGTM :+1:

pragmaticivan:
Something tells me that we should go using `baseurl` instead the one we're using now.

Not clear but looks like we can avoid the problem change urls just updating the README and passing a flag to tell the server to use localhost:

screen shot 2016-04-13 at 5 25 02 pm


We need a little bit more of research.

doomspork:
Thank @jakefolio.  @pragmaticivan if you want to look into that, I'd be grateful!

PR #283, opened 2016-04-14
Fix a wording in plug.md

AquarHEAD:
I believe it should be "GET requests"?


doomspork:
Good catch @AquarHEAD :+1:

PR #289, opened 2016-04-17
Fix wrong numeration of lessons

erickgnavar:


doomspork:
Thanks @erickgnavar 👍

PR #291, opened 2016-04-17
Applied the fix from PR #290 to every translation.

nscyclone:


doomspork:
Great catch @nscyclone, I meant to ask the original PR to get these.

PR #306, opened 2016-04-25
Make a general view for sigils list

nscyclone:
Every noun of the sigils list besides `strings` was singular.
I also think that it might be useful to point that `~W` and  `~w` sigils generate word lists but not character lists or anything else. It should look more clearly this way.


doomspork:
This is a great change, thank you @nscyclone 👍

PR #309, opened 2016-04-25
[PL] Moving files to proper directories

Koziolek:
Fix of my mistake with putting files to wrong paths. @doomspork could you merge it?


doomspork:
Good catch @Koziolek, sorry I missed that before.  👍

Issue #275, opened 2016-04-11
Upgrade Elixir version in examples?


AquarHEAD:
For example in [Basic](https://elixirschool.com/lessons/basics/basics/), the first code example is 2 minor version behind latest Elixir version.

```
Erlang/OTP 17 [erts-6.4] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Interactive Elixir (1.0.4) - press Ctrl+C to exit (type h() ENTER for help)
iex>
```

doomspork:
@AquarHEAD can you take the lead on updating all of the examples?  One PR should be sufficient for all of the changes.

AquarHEAD:
@doomspork Sure will do as soon as I find enough time to go over everything :)

doomspork:
Great!  Thank you @AquarHEAD

nscyclone:
Dear @AquarHEAD, you were gone for a long time, so I made a PR for this issue on my own: #321. I hope I didn't have done any harm by this.

AquarHEAD:
@nscyclone Thanks, was on a business trip last 2 weeks, and now just started a vacation, so barely have time to do things on my own( I'm happy that someone else could help, that's part of the reason why I opened an issue first :D

doomspork:
Thank you @AquarHEAD and @nscyclone 👍

PR #321, opened 2016-04-27
Bump Elixir version in examples (#275)

nscyclone:
I just made another key in Jekyll's config so it will be easy to update Elixir version later. For now, I've set it to 1.2.3.
Changes were made in every existing translation of the following lessons:
- lessons/advanced/umbrella-projects.md;
- lessons/basics/basics.md;
- lessons/specifics/mnesia.md.

I also think we should note the use of  `{{site.elixir.version}}` in CONTRIBUTING.md or somewhere else.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -100,3 +100,6 @@ sitemap:
       - '/index.html'
   change_frequency_name: 'change_frequency'
   priority_name: 'priority'
+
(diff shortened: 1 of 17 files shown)


doomspork:
Thank you @nscyclone, this looks great!

Issue #314, opened 2016-04-26
Basic menu section is always open


eksperimental:
Regardless if you are reading an article in the other sections.

Browser: Chrome and Firefox
OS: Linux

doomspork:
Thanks for the report @eksperimental!

nscyclone:
I'm going to fix this within the next few days.

doomspork:
Thank you @nscyclone, that'd be great!

doomspork:
Fixed by @nscyclone in #330

PR #330, opened 2016-04-29
Open only current section in the sidebar (#314)

nscyclone:
Earlier sidebar sections were toggled via JS: onload the `click` event was manually triggered on `advanced` and `specifics`.
Now we do not use any JS to set the sections' initial state when the page loads: appropriate CSS is applied in Liquid depending on the current page category. If the page has no category (e.g., we've just visited the index when we heard about Elixir School first time), `basics` is open by default.


doomspork:
This is great @nscyclone, thank you!

eksperimental:
:+1:

PR #339, opened 2016-05-01
Fix a few typos in ETS examples

jc00ke:
These examples are super handy, thanks!


doomspork:
Good catch @jc00ke, thank you 👍

jc00ke:
Thank you!

Issue #336, opened 2016-05-01
Automatic 'table of contents' building


brain-geek:
Hey!

I think it would be very useful to have 'table of contents' build automatically in a lesson. The way we do it right now is error-prone.

All data is already there (we always link to all h* elements in a lesson), we just need to use it somehow. 

For example, there is a [js script](https://github.com/ghiculescu/jekyll-table-of-contents) to do that.

What do you think?

doomspork:
@brain-geek if you want to set this up, I'm open to using it 👍

Koziolek:
Remember guys that we need to introduce this in all files in all translations (lot of testing).

doomspork:
Agreed @Koziolek

nscyclone:
I am already working on this. (Just in case if someone wanted to start.)

doomspork:
Thank you @nscyclone 💜

nscyclone:
It would be nice to have some sort of rules to share with the authors of pending translations because they do not know about this feature yet and still continue creating manual TOCs.
I think we could update `README` or `CONTRIBUTING` to reflect the usage of the TOC generator template and then close this issue. Recommendations for translations are somehow located in `README` so I can't decide where should we place our new advice. There may also be something else worthy to be pointed for contributors.
#194 is still a work in progress.
Do you have any suggestions, @doomspork?

doomspork:
Very good point @nscyclone.  I'm not sure people read either of those documents but that is certainly a start.  We should probably also ask the existing PR authors to update their changes, right?

nscyclone:
@doomspork, here is [an interesting thing](https://github.com/blog/2111-issue-and-pull-request-templates) we could look at. This may help us with newer PRs until we release new `CONTRIBUTING` and `README` (actually it may help after that too). Anyway, I couldn't find any other solution so I think we could do nothing else but leave a comment in every existing PR manually.

nscyclone:
@doomspork, I've just left a copy of advice in every open PR with translations: #229, #244, #264, #316, #319, #320, #326, #343, #345, #347, #350, #352, #353 and #359.

doomspork:
Thank you @nscyclone, that is tremendously helpful!

doomspork:
Huge thanks to @nscyclone and @brain-geek for this 👍

PR #342, opened 2016-05-03
Set a proper id for the header of social links block

nscyclone:
It seems like the old id was just copied from the `basics` lesson.
The new correct id may also become useful for generation of the TOC (#336).

### DIFF
--- a/_includes/social.html
+++ b/_includes/social.html
@@ -1,5 +1,5 @@
 <hr />
-<h3 id="string-concatenation">Share This Page</h3>
+<h3 id="social">Share This Page</h3>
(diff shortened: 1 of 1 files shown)


doomspork:
Thank you @nscyclone

doomspork:
@nscyclone what is your twitter handle?  Apologizes if I've asked you already.

nscyclone:
@doomspork, no, you didn't ask this earlier. I actually have one (twitter.com/nscyclone) but I'm afraid I don't use it enough to worry. Some sort of silent readonly, I do not even follow anyone.

Issue #344, opened 2016-05-03
Linking in String lesson not works


Koziolek:
Some of links in table of content in String lesson not works. 
List:
 * Graphemes and Codepoints
 * all functions subsections

doomspork:
This should have been fixed via #346

PR #346, opened 2016-05-03
Fix links from String lesson

jeanmatheussouto:
Fix issue #344 

Fixes:
 - Resolved links from String lesson.
 - Add link to `Split` function

### DIFF
--- a/lessons/basics/strings.md
+++ b/lessons/basics/strings.md
@@ -12,11 +12,12 @@ What are Strings in Elixir, Char lists, Graphemes and Codepoints.
 
 - [Strings in Elixir](#strings-in-elixir)
 - [Char lists](#char-lists)
-- [Graphemes and Codepoints](#graphemes)
+- [Graphemes and Codepoints](#graphemes-and-codepoints)
 - [String functions](#string-functions)
-  - [Length](#length)
-  - [Replace](#replace)
-  - [Duplicate](#duplicate)
+  - [Length](#length1)
+  - [Replace](#replace4)
+  - [Duplicate](#duplicate2)
+  - [Split](#split2)
 - [Exercises](#exercises)
   - [Anagrams](#anagrams)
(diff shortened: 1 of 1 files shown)


lessons/basics/strings.md
```diff
@@ -12,11 +12,12 @@ What are Strings in Elixir, Char lists, Graphemes and Codepoints.
 
 - [Strings in Elixir](#strings-in-elixir)
 - [Char lists](#char-lists)
-- [Graphemes and Codepoints](#graphemes)
+- [Graphemes and Codepoints](#graphemes-and-codepoints)
 - [String functions](#string-functions)
-  - [Length](#length)
-  - [Replace](#replace)
-  - [Duplicate](#duplicate)
+  - [Length](#length1)
```
doomspork: @jeanmatheussouto instead of changing these, how about we update the headers to no longer be code instead?

For instance, changing line 72 from `##### length/1` to `##### Length`.

jeanmatheussouto: I wanted to keep the default to display the method definition and its parameters, but I think valid change to become more friendly.

I do it!

lessons/basics/strings.md
```diff
@@ -69,36 +69,36 @@ iex> String.graphemes string
 
 Let's review some of the most important and useful function the String module has for us.
 
-##### `length/1`
+##### Length
 
-Returns the number of Graphemes in the string.
+Returns the number of Graphemes in the string. Function definition is `length/1`.
```
jeanmatheussouto: @doomspork what do you think of `Function definition is `length/1` to show function definition unless just show function?

doomspork: That works for me @jeanmatheussouto.  Could you end each of them with a colon?

```
Function definition is `length/1`:
```

Then this is good to merge 👍

PR #348, opened 2016-05-05
Add missing returned map from initializing a map.

IamCharlesHan:


doomspork:
Thanks @IamCharlesHan 👍

Issue #202, opened 2016-03-23
Polish Translations


pjhampton:
## Get Involved!

We are seeking **Polish** translators for this project. If you would like to get involved with translating a lesson. Please folk the project and start translating.

- **Set up**
  - [x] _config.yml
  - [x] index.md 

- **Basics**
  - [x] Collections
  - [x] Enum
  - [x] Pattern Matching
  - [x] Control Structures
  - [x] Pipe Operator
  - [x] Functions
  - [x] Composition
  - [x] Mix
  - [x] Sigils
  - [x] Documentation
  - [x] Testing
  - [x] Comprehensions
  - [x] Strings
  - [x] Custom Mix Tasks

- **Advanced**
  - [x] concurrency.md
  - [x] errors.md
  - [x] erlang.md
  - [x] escripts.md
  - [x] otp-concurrency.md
  - [x] otp-supervisors.md
  - [x] metaprogramming.md
  - [x] umbrella projects.md

- **Specifics**
 - [x]  ecto.md
 - [x]  eex.md
 - [x]  ets.md
 - [x]  mnesia.md
 - [x]  plug.md

**Note**
Those new to Elixir and / or Github are encouraged to get involved :smile:. 1 [Pull Request](https://help.github.com/articles/using-pull-requests/) per lesson.

Koziolek:
I could do part of this.

Koziolek:
Forked to https://github.com/Koziolek/elixir-school branch issue-202
If anyone else would like to help please send me message we will coordinate our work

pjhampton:
Yep, sounds good @Koziolek! You lead this and we will send out a contributors later!

Thanks :smile:

doomspork:
That is great news @Koziolek!  @kdebowski already opened our first Polish PR if you want to take a peek at it: #271

kdebowski:
@Koziolek, I've already started translating files to polish language. Files are divided into chapters and I'm in the beginning os basics.md - you can easily pick some file from the rest if you want to evade potential conflicts

Koziolek:
@kdebowski 
Basics done ;) we need to coordinate it. Do you have trello account to coordinate translation (and make subtrasks list)?
Now I start to translate eex.md from special

kdebowski:
@Koziolek I've created Trello board for polish translations. If you could provide me your email account for Trello I can add you as a member.

doomspork:
@kdebowski and @Koziolek, what are you twitter handles? 😀

[...]

PR #350, opened 2016-05-05
[PL] add translation of metaprogramming

Koziolek:
In scope of issue #202 

@kdebowski :)


nscyclone:
We have implemented a new feature that automatically generates the table of contents: #356.
This may help you avoid copypaste, typos and broken anchors. In order to use it you can replace the manually written TOC with `{% include toc.html %}`.

PR #356, opened 2016-05-09
Implement automatic TOC generation

nscyclone:
First of all, I want to thank @brain-geek for this great idea. During the work over this PR I've found that it will fix a lot of different shortcomings like non-working anchors, typos and so forth.
I had to test everything manually: I was looking for visual differences between the local version with generated TOC and the original one hosted at http://elixirschool.com/ and also ensured that the new anchors are working well. I think there is actually almost nothing to be worried about: if the lessons have passed the localizers' checks before merge then the headers were named correctly and the new anchors must be correct as well.
Here are the things to notice:
* general:
    * if we assume this PR viable and merge it then we need to use `{{ include toc.html }}` in the new lessons and **pending translations** &mdash; this could be reflected somewhere in `CONTRIBUTING`;
    * `h5` headers were replaced with `h3` in `basics/strings` because they were not displaying correctly: `h3` is expected inside `h2`;
    * native `<a/>` anchors before headers were removed in Portuguese and French lessons: they are not needed anymore because the script works fine with automatically applied `id` regardless of diacritics and that kind of stuff;
    * the `basics/compositions` lesson lacked the `filtering` section: it was not included in the original lesson (I think it was a mistake since this is a header too) but now it is present in every translation;
* localizations:
    * Malay (@alkaha, could you, please, check the following?):
        * there are 2 different translations used for the `Table of Contents` phrase: `Kandungan` and `Isi kandungan`. Which one should we use? I've chosen `Kandungan` for now;
        * there was no space between the long words in the title of `advanced/erlang` and some part of it was [outside of the sidebar box](http://i.imgur.com/rC3IMD9.png);
        * a typo was found in `basics/documentation`: `pencatatatan` → `pencatatan`;
        * in the `basics/basics` lesson the first anchor was translated (`Penyediaan`) but the header itself was named `Setup` so I renamed it according to the anchor.

I am looking forward to any feedback on this.

@doomspork, I am also not sure whether should this be rebased or not.


doomspork:
This is great @nscyclone!  I'm going to pull this down and test it out locally today.

It looks like this needs to be rebased in order to merge, can you take care of that?  Thank you!

nscyclone:
Thanks for a quick reply, @doomspork!
I've just squashed the branch into a single commit (rebased it onto `1feebf3`).

alkaha:
1. I fixed the issues mentioned.
2. You can use Isi Kandungan for TOC auto generation.

nscyclone:
Thanks, @alkaha !
Actually I didn't understand what did you fix because I've already fixed everything from the `localization` list of the PR description here. I just asked you to check if those fixes were correct.
Now using `Isi Kandungan`, the branch was rebased again.

alkaha:
@nscyclone my bad. sorry. :)

brain-geek:
That's awesome! 👍

doomspork:
@nscyclone I had a chance to play with this and it works great!  It does look like you need to rebase once more, sorry!

Once this is merge-able we're good to go!

nscyclone:
@doomspork, I'm glad to inform you that we're finally ready to merge this. Thanks!

doomspork:
This is so great, thank you very much @nscyclone 🎉

PR #360, opened 2016-05-12
Apply the TOC generator to the remaining translations

nscyclone:
Added automatic TOC generation to the translations that were added during our work over it.


doomspork:
Thank you @nscyclone, I will keep an eye out for this on PRs.  We should probably ping the current pending PRs and ask them to update, what do you think?

PR #369, opened 2016-05-17
[KO] Translation of Control Structures Lesson

Dalgona:
#201 any comments are welcome 😅


Dalgona:
@devleoper, @rijgndqw012

rijgndqw012:
pretty nice!

devleoper:
Pretty cool. 👍 

Plus, everybody has their own flavors in translating, (That doesn't mean someone does better than others.) When Korean Translation of this documentation is done entirely, We need to match our delivery. Let's go, guys! :blush:

PR #371, opened 2016-05-17
[KO] Translation of Pattern Matching Lesson

Dalgona:
#201 :sweat_smile: Any improvements or suggestions are welcome.


Dalgona:
@devleoper, @rijgndqw012

rijgndqw012:
No problem. very nice.

---------------------------------------------
While translating some words to korean word, i found few things which is ambiguous to translate. For example, some people say list comprehension as '리스트 해석', other people say '리스트 통합'
So, we searched those keywords in google, other functional programming language reference, and erlang programming language book(all of these are translated into korean).  

It may remains inconsistency of tone, or wrong translated words.
To avoid awkward translation and share how to translate some ambiguous words, we made trello board which is related to translation. if you see something awkward and report us, it would be very appreciated.

PR #378, opened 2016-05-20
Don't use sigils in a chapter that comes before sigils

kyledecot:


doomspork:
Sounds like a good idea @kyledecot, I'll look this over and let you know if I've got any feedback.  Thank you 👍

doomspork:
Thank you again @kyledecot 😀

PR #383, opened 2016-05-21
[JP] Fix wrong result of example code

hinamiyagk:


doomspork:
Thank you @hinamiyagk!  @3100 if you have a moment, could you look this one?

PR #386, opened 2016-05-22
[JP] Update translation in 'basics/control-structures'

hinamiyagk:
Elixir 1.2 updates(`with` expression)


doomspork:
Thank you @hinamiyagk!  @3100 can you take a look at this when you have some time?

3100:
Nice work. 
I suggest some points below, but  I will leave the choices up to you, @hinamiyagk .

- It would be better to translate the phrase `The special form` as `スペシャルフォーム` or `特殊形式` because it is a well-known technical term in some programming languages. 
- `知っておくだけで良いです` seems to be more natural translation of `only need to know`.

hinamiyagk:
Updated

3100:
Thank you so much :+1:

doomspork:
Thank you @hinamiyagk and @3100 👍

Issue #366, opened 2016-05-16
Probably missed description in Ecto.counting section


Koziolek:
In lesson about Ecto in section `Counting` we have some code only. Maybe we need some small description like:

```
If we want to count number users that has confirmed account we could use `count/2`:
```

Changes made to the original lesson could be found [here](https://github.com/doomspork/elixir-school/pull/401/files).
## Translation

Translation progress:
- [x] en
- [x]  ru /cc @nscyclone
- [x]  ko /cc @devleoper
- [x]  jp /cc @3100
- [ ]  no /cc @therobinsj
- [x]  pl /cc @Koziolek
- [ ]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [x]  cn /cc @cizixs
- [ ]  es /cc @aleandros / @jonahoffline / @eze1981
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @apeacox / @sformisano
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima

nscyclone:
I'll take care of updating the Russian translation.

doomspork:
@Koziolek could you open a PR for English & Polish?  We'll use that to get the other translations completed. 👍

nscyclone:
@Koziolek, how do you look on updating this issue with [this](https://github.com/doomspork/elixir-school/issues/598) format? `en`, `ru` and `pl` translations are already completed.

Koziolek:
It's OK for me :)

hinamiyagk:
Japanese translation is done(#444)

cizixs:
Chinese translation is done at #748.

PR #401, opened 2016-05-24
Add description to count function

Koziolek:
In scope of #366. Added description of `count/1`, and information about `count/2` function. 

@doomspork

### DIFF
--- a/lessons/specifics/ecto.md
+++ b/lessons/specifics/ecto.md
@@ -187,11 +187,23 @@ In addition to `all/2`, Repo provides a number of callbacks including `one/2`, `
 
 ### Count
 
+If we want to count number users that has confirmed account we could use `count/1`:
+
 ```elixir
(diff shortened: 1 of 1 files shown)


lessons/specifics/ecto.md
```diff
@@ -187,12 +187,24 @@ In addition to `all/2`, Repo provides a number of callbacks including `one/2`, `
 
 ### Count
 
+If we want to count number users that has confirmed account we could use `count/1`:
```
nscyclone: It looks like there is a missing preposition: `to count number users` → `to count the number of users`.

Koziolek: done

doomspork:
Sounds good @Koziolek, thank you! 👍 

We should take note of which translations need to be updated now

Koziolek:
all.

PR #413, opened 2016-05-28
Correct case and typoe and fix/improve wording

yyyc514:


doomspork:
Looks good, thank you again @yyyc514 👍

PR #416, opened 2016-05-28
Replace only takes 3 arguments here

yyyc514:


yyyc514:
If this types of small fixes are welcome (see other PRs) I can take a little time and review the entire guide and submit a larger PR.  I see all sorts of small typos and minor grammar issues that could be cleaned up.

doomspork:
Please do @yyyc514!!  That would be _fantastic_

PR #419, opened 2016-05-28
Language links on the index page should redirect to configured URL

nscyclone:


doomspork:
Thank you @nscyclone 👍

PR #428, opened 2016-05-31
Japanese translation 1

HirofumiTamori:
Translated Strings session into Japanese.

### Opinion
This time, I just translated as same as the original, however I think multibyte character languages including Japanese should have additional description / examples / notes for the native users.
For example (Japanese small kana):
```elixir
iex> string = "\uFF76\uFF9E"
"ｶﾞ"
iex> String.codepoints string
["ｶ", "ﾞ"]
iex> String.graphemes string
["ｶﾞ"]
iex> String.length string
1
```
(note: I suspect that there is still no de facto standard of Japanese small kana's graphemes/codepoints translations, but Elixir just do it in this way)

### DIFF
--- a/jp/lessons/basics/sigils.md
+++ b/jp/lessons/basics/sigils.md
@@ -1,0 +1,168 @@
+---
+layout: page
+title: Sigils
+category: basics
+order: 10
+lang: ja
+---
+
+シギル(sigil)を使う/作る
+
+{% include toc.html %}
+
+## シギルの概要
+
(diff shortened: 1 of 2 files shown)


doomspork:
Thank you @HirofumiTamori 👍 

@3100 could you take a look at these changes?

3100:
@HirofumiTamori I agree with your opinion. Please feel free to add some changes if you would like to apply them.

3100:
Thank you, @HirofumiTamori!

PR #424, opened 2016-05-30
various textual, grammar, clarity improvements to the lessons

yyyc514:

PR #429, opened 2016-05-31
Swap the page order of `functions.md` and `pipe-operator.md`

nscyclone:
Applied the order swap from #424 to every language.


doomspork:
Good catch, thank you @nscyclone!  You're on top of it 👍

PR #430, opened 2016-06-01
Replace `rem(x, 2)` with `is_even(x)`

nscyclone:
Applied the replacement from #424 to every language.


doomspork:
I'm sorry I missed all of these, distracted by my move I guess.  Thank you for the help @nscyclone, it's greatly appreciated.

PR #433, opened 2016-06-02
Fix the headings and arity of functions in `basics/strings`

nscyclone:


doomspork:
Thank you @nscyclone!

PR #435, opened 2016-06-03
removing concation project references

Zahidul-Islam:
I have removed concation project references from all translations


doomspork:
Thank you very much @Zahidul-Islam, much appreciated 👍

PR #434, opened 2016-06-02
Arabic translation

jaksz:
Can someone render the Arabic `basics.md` file and see if the right-to-left text is displaying properly in Jekyll so far?

I added a `dir: rtl` property into the Arabic `basics.md` file, hopefully it helps somehow.

PR #442, opened 2016-06-08
Enable RTL languages support

nscyclone:


doomspork:
@nscyclone you are awesome, this is fantastic.  I take it you've tested this out and it's safe to merge and deploy?

nscyclone:
@doomspork, thanks! There is not that much to test I guess: layout should be fine in both mobile and desktop view, the headers are generated correctly and the links work as expected. But I think nothing bad will happen if we check this once more. 
I also suppose we could do something with #231 because the RTL setup is the subject of this PR and translations are now separated as well.

@jaksz, [this](http://imgur.com/a/oFVeK) is how the full updated translation looks like. Could you please take a peek at it? Unless something has been broken we're ready to merge.

jaksz:
@nscyclone, everything is correct except one small thing:



In Arabic, the `و` (meaning `and`) is connected to the next word, but in this case the new word is supposed to be bolded to reflect the English original, and the `و` prevents that. Can you add a space between the `و` and the `_`? This should fix it

jaksz:
@nscyclone, there was a small formatting issue which I have fixed in ce4657985b30e848a9d5bb8e2177912b0117b614

doomspork:
@nscyclone good point RE: #231, I think most of it is addressed in #434 so we should be good.

nscyclone:
@doomspork, I am wondering if my conclusion is expected to merge this one. If so I am happy to say that it works fine as we see it with the translation of the first lesson from @jaksz.

doomspork:
@nscyclone no, my bad!  I tested it locally and hadn't gotten back to merging.  Thank you again 👍

PR #445, opened 2016-06-12
[JP] Fix wrong order of articles

hinamiyagk:


3100:
@hinamiyagk Good catch! Thank you.

PR #446, opened 2016-06-12
Links to the official doc should always refer to its stable version

3100:
There are many links to the official elixir-lang.org/docs, but they refer to several different versions. How about if we always use elixir-lang.org/docs/stable for less maintenance?


doomspork:
This is a great idea, thank you @3100!!

PR #452, opened 2016-06-14
Explain syntax for updating maps and access atom keys. Fix #402

### LINKED ISSUE #402: Never mentioned how to update a map before demonstrating it on struct
In Basics -> Composition -> structs there is an explanation of how to update a struct:

> We can update our struct just like we would a map:

However, update a map was never introduce before.

Nagasaki45:
I took the text for the official [elixir getting-started guid](http://elixir-lang.org/getting-started/keywords-and-maps.html#maps), and simplified the example a bit.

### DIFF
--- a/lessons/basics/collections.md
+++ b/lessons/basics/collections.md
@@ -148,3 +148,14 @@ iex> %{foo: "bar", hello: "world"}
 iex> %{foo: "bar", hello: "world"} == %{:foo => "bar", :hello => "world"}
 true
 ```
+
+Another interesting property of maps is that they provide their own syntax for updating and accessing atom keys:
+
+```elixir
+iex> map = %{foo: "bar", hello: "world"}
+%{foo: "bar", hello: "world"}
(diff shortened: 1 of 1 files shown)


doomspork:
Thank you @Nagasaki45!  I think this is something we will want added to the translations as well.

Nagasaki45:
@doomspork, how can I help with that?

doomspork:
@Nagasaki45 I'm wondering that myself!  These little changes are a bit tougher to track and make sure they're translated.  Before I merge this I want to ping a few of the translators and get some thoughts on tracking these, sound good?

Nagasaki45:
Sounds great!
@meyvain, my significant other, showed interest in helping to translate this to spanish.

doomspork:
That would be fantastic @Nagasaki45 & @meyvain!

doomspork:
Thank you @Nagasaki45 👍

PR #463, opened 2016-06-20
[KO] Translation of Documentation lesson

malkoG:
related to #201 

just pruned branches :)


devleoper:
Since there were some branch-related troubles with PR #388.

PR #468, opened 2016-06-21
[KO] Translation of Executables Lesson

devleoper:
For comparing original text to translated, refer to devleoper/elixir-school#2.

It might have some outdated content, I'm going to fix it.

Related: #201


malkoG:
LGTM, :smile_cat:

PR #469, opened 2016-06-21
Add a share button for `vk.com`

nscyclone:
[VK](https://new.vk.com/) is [the biggest social network](https://en.wikipedia.org/wiki/VK_(social_networking)) in the CIS countries.
Along with adding the share button for it I also removed the repeating inline style and fixed the formatting in `social.scss`.


doomspork:
Thank you @nscyclone, I really appreciate all of your help 👍

PR #476, opened 2016-06-29
Add arity to functions in EEx lesson

brain-geek:
Right now we have arity in all function mentions. This PR adds arity to function names in EEx lesson.

### DIFF
--- a/es/lessons/specifics/eex.md
+++ b/es/lessons/specifics/eex.md
@@ -16,7 +16,7 @@ La API EEx soporta trabajar con cadenas y archivos directamente. El API se divid
 
 ### Evaluación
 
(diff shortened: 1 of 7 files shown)


brain-geek:
@nscyclone Can you look through this one? Just for verification.

nscyclone:
This is a nice catch, @brain-geek! It's totally fine but I'm afraid the fix should applied to every existing translation (`es`, `id`, `ko`, `pl`, `pt`, `vi`) and the pending `ru` translation as well.

brain-geek:
@nscyclone Thanks for noting! I've fixed all existing translations + ru in separate PR.

doomspork:
This is great, thank you @brain-geek and @nscyclone!

PR #479, opened 2016-06-30
Correct the `basics/strings` lesson

nscyclone:


doomspork:
Thank you @nscyclone!  I doubt this is something we need to concern translators with, I'm assuming (hopefully not wrongfully) that these sorts of things would be resolved in the translation process?

nscyclone:
@doomspork, I suppose there is nothing to worry about — meaning still remains the same, issues were related only to grammar. So, nor the existing, neither the future translations should be affected by this.

PR #489, opened 2016-07-03
Add 'debugging' section to Metaprogramming lesson

brain-geek:
This new section should allow simpler debugging and look into the result of macro-related actions. This may also give a small insight into Elixir, as things like 'if is a macro' are expanded.


doomspork:
This is great @brain-geek!  Could you also do the RU translations for this?

brain-geek:
Sure, @doomspork . But I guess that translation should be in separate PR after this one is merged.

lessons/advanced/metaprogramming.md
```diff
@@ -131,6 +131,53 @@ def test do
 end
 ```
 
+## Debugging
+
+Okay, right now we know how to `quote/2`, `unquote/1` and write macroses. But what if you have a huge chunk of quoted code and want to understand it? In this case, you can use `Macro.to_string/2`. Take a look at this example:
```
doomspork: "know how to use", "macros"

doomspork: Maybe we can say "want to troubleshoot it" instead of "want to understand it".

lessons/advanced/metaprogramming.md
```diff
@@ -131,6 +131,53 @@ def test do
...
+
+```elixir
+iex> Macro.to_string(quote(do: foo.bar(1, 2, 3)))
+"foo.bar(1, 2, 3)"
+```
+
+And when you want to look at the code generated by macros you can also combine this with `Macro.expand/2` and `Macro.expand_once/2`. These functions expand macros in the given quoted code. The first may expand it several times, while the former - only once. For example, if we take slightly modified 'unless' example from the previous section:
```
doomspork: Let's do "you can also combine this with" -> "you can combine them with".

What about: ". These functions" -> ", these functions"

doomspork: What about "For example, let's modify `unless`"?

lessons/advanced/metaprogramming.md
```diff
@@ -131,6 +131,53 @@ def test do
...
+end
+
+require OurMacro
+quoted = quote do
+  OurMacro.unless true, do: "Hi"
+end
+```
+
+```
+iex(7)> quoted |> Macro.expand_once(__ENV__) |> Macro.to_string |> IO.puts
```
doomspork: Can you remove the line number from `iex`?  `iex> quoted |>`

lessons/advanced/metaprogramming.md
```diff
@@ -131,6 +131,53 @@ def test do
...
+iex> quoted |> Macro.expand(__ENV__) |> Macro.to_string |> IO.puts
+case(!true) do
+  x when x in [false, nil] ->
+    nil
+  _ ->
+    "Hi"
+end
+```
+
+You may remember that we've already mentioned that `if` is a macro in Elixir. Here you can see that really it gets transformed to `case` statement.
```
doomspork: "Here you can see that it gets transformed to a `case` statement."

brain-geek:
Thanks for the feedback, @doomspork !

I've applied all changes you suggested. Can you look through this PR again? 

Thanks!

lessons/advanced/metaprogramming.md
```diff
@@ -130,6 +130,53 @@ def test do
...
+## Debugging
+
+Okay, right now we know how to use `quote/2`, `unquote/1` and write macros. But what if you have a huge chunk of quoted code and want to understand it? In this case, you can use `Macro.to_string/2`. Take a look at this example:
+
+```elixir
+iex> Macro.to_string(quote(do: foo.bar(1, 2, 3)))
+"foo.bar(1, 2, 3)"
+```
+
+And when you want to look at the code generated by macros you can combine them with `Macro.expand/2` and `Macro.expand_once/2`, these functions expand macros in the given quoted code. The first may expand it several times, while the former - only once. For example, let's modify `unless` example from the previous section:
```
doomspork: How about: "these functions expand macros into their given quoted code."?

lessons/advanced/metaprogramming.md
```diff
@@ -130,6 +130,53 @@ def test do
...
+iex> quoted |> Macro.expand(__ENV__) |> Macro.to_string |> IO.puts
+case(!true) do
+  x when x in [false, nil] ->
+    nil
+  _ ->
+    "Hi"
+end
+```
+
+You may remember that we've already mentioned that `if` is a macro in Elixir. Here you can see that it gets transformed to a `case` statement.
```
doomspork: How about:

You may recall that we've mentioned  `if` is a macro in Elixir, here we see it expanded into the underlying `case` statement.

doomspork:
@brain-geek I left two more pieces of feedback, neither is a must.  This is good to go for me either way 👍

brain-geek:
Thanks for new feedback, @doomspork . I've applied both and merging.

PR #498, opened 2016-07-07
Fix mis-spelling of 'oops' in Plug lesson

GregMefford:


doomspork:
@GregMefford could you see if this change needs to be made in other languages as well?  Thank you 👍

GregMefford:
I bet it does. I'll check and update the PR.

GregMefford:
I've updated the PR to include all the languages that have the Plug lesson translated.

As an aside, I noticed that there's a `vi` language, which is actually English but with very minor grammatical differences between it and the main English version. Is this intended to be a translation for the US Virgin Isles, or maybe it was a mistake made by someone using Vim as an editor?

devleoper:
`vi` stands for Vietnamese. I think these translators just copied all of English lessons before starting translation.

doomspork:
What @devleoper said is correct 😀

doomspork:
Thank you @GregMefford 👍

Issue #480, opened 2016-06-30
Updates to Basic's Atom section


doomspork:
It would be great to incorporate the additions from #470 into the other languages.

Translation progress:
- [x] en
- [x] ru /cc @brain-geek / @nscyclone
- [x] ko /cc @devleoper 
- [x] jp /cc @3100 
- [ ] no /cc @therobinsj 
- [x] pl /cc @Koziolek 
- [ ] sk /cc @michalvalasek 
- [ ] fr /cc @julienXX
- [x] cn /cc @cizixs 
- [ ] es /cc @aleandros 
- [ ] pt /cc @pragmaticivan 
- [ ] vi /cc @joneslee85 
- [ ] it /cc @lucidstack 
- [ ] my /cc @alkaha
- [ ] id /cc @rezaprima

doomspork:
If this format doesn't work for folks, let me know what might work better.  Thank you! 👍

Koziolek:
Done

christiansakai:
@doomspork I can help for id translation if @rezaprima doesn't mind.

doomspork:
@christiansakai all contributions are welcome! 😀

PR #502, opened 2016-07-11
CN: update atom section in basic lesson.

cizixs:
Apply updates mentioned in #480 in Chinese.


doomspork:
Thank you @cizixs!! 👍

Issue #481, opened 2016-06-30
Translate section on updating/accessing maps


doomspork:
Add section on updating and accessing maps with atoms #452 

Translation progress:
- [x] en
- [x] ru /cc @brain-geek / @nscyclone 
- [x] ko /cc @devleoper 
- [ ] jp /cc @3100 
- [ ] no /cc @therobinsj 
- [x] pl /cc @Koziolek 
- [ ] sk /cc @michalvalasek 
- [ ] fr /cc @julienXX
- [x] cn /cc @cizixs 
- [ ] es /cc @aleandros 
- [ ] pt /cc @pragmaticivan 
- [ ] vi /cc @joneslee85 
- [ ] it /cc @lucidstack 
- [ ] my /cc @alkaha
- [ ] id /cc @rezaprima

Koziolek:
done

anderkonzen:
Gonna take the PT translation

PR #503, opened 2016-07-11
CN: add section on updating and accessing maps with atoms.

cizixs:
apply changes mentioned in #481 in Chinese.


doomspork:
@cizixs you're awesome, thank you so much 👍

PR #505, opened 2016-07-12
Bump Elixir, erts & OTP versions.

nscyclone:


doomspork:
And it's so easy thanks to you @nscyclone! 👍

Issue #305, opened 2016-04-25
Possible improvement of an example in Basics > Pipe Operator


odiumediae:
One of the examples in the Pipe Operator section could be replaced, since:

```
"Elixir rocks" |> String.split |> Enum.map( &String.upcase/1 )
```

could simply be written:

```
iex> "Elixir rocks" |> String.upcase |> String.split
["ELIXIR", "ROCKS"]
```

which is not only shorter and more idiomatic, but also more readable. Under "Best Practices" the same example is used again, without the parentheses. A slightly different example might be helpful here. Maybe something like this:

```
# bad example
iex> "Elixir rocks" |> String.downcase |> String.split |> Enum.map &String.to_atom/1

#good example
iex> "Elixir rocks" |> String.downcase |> String.split |> Enum.map(&String.to_atom/1)

[:elixir, :rocks]
```

doomspork:
@odiumediae please open a PR with your suggested changes 👍

odiumediae:
[@doomspork](https://github.com/doomspork): I'd like to. Sorry, if this is a dumb question, but should I change the respective templates of all supported languages or just in the main (English) one?

Apropos languages, if I find the time, I'd like to provide the German translation. Any other German speakers here who'd like to help me?

doomspork:
@odiumediae if you wouldn't mind making the example changes across all languages everyone would greatly appreciate that.

We have an issue [here](https://github.com/doomspork/elixir-school/issues/200) for German Translations, a few people have expressed interest but no one has started.

doomspork:
This was addressed in #511  👍

PR #511, opened 2016-07-16
Improved "Pipe operator" lesson

brain-geek:
Improved "Pipe operator" lesson example according to suggestion in #305 task.


doomspork:
Awesome @brain-geek, thanks for your help 👍

PR #517, opened 2016-07-16
Changing header metadata to show correct information with sharing.

devleoper:
Changing `og:url` value to URL of each page will help showing information in Open Graph correctly. (Currently if we share any page to Facebook, we will see same previews because `og:url` points only the main page.)

And changed `description` and `og:description` to show translated description of Elixir School.

### Where can we go further?

We can set `description`s which make summary for each page to make sneak-peeking easy.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -5,13 +5,13 @@
 
   <link href="http://gmpg.org/xfn/11" rel="profile">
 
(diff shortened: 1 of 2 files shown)


doomspork:
This is great @devleoper!!  This is good to merge from my perspective, any additional changes you'd like to make before I do so?

devleoper:
It's ready to be merged. I think there isn't additional thing at this time for this. :smiley:

PR #518, opened 2016-07-16
[KO] Renamed and applied enhancement for Modules Lesson

devleoper:
There was many changes in PR #424, but I thought that I must apply this first to prevent 404 Errors. There are also readability improvements.

@malkoG @Dalgona @riseshia Do you think this changes will make out translations more readable?

PR #519, opened 2016-07-16
Renamed "Composition" Lessons in other languages to "Modules" Lesson

devleoper:
Applied the rename from #424 to other languages.
Renaming of Korean lesson is seperated to #518 with minor changes 

There are not many changes in the contents, I just renamed the lesson for preventing (possible) 404 while changing language.


doomspork:
Thank you @devleoper 👍

PR #522, opened 2016-07-17
Rename `lessons` to `en` in order to clarify lesson structure/location

ybur-yug:
The first time I cloned and jumped in after looking over some issues and wanting to poke around the project, I actually couldn't find the English lessons until I realized it was `lessons` and not the 2-character code for english (`en`) like the other languages.

This is not only slightly confusing but treats English as if it is a higher priority over other languages. I thought it made sense to just simply treat all languages equally and consider them equal first class citizens.

### Note
There may be jekyll configuration changes needed to handle the name change. I poked around but didn't see anything obvious after grepping/checking general config. Let me know if there would be more needed changes.


doomspork:
Thank you @ybur-yug, I concur this is a good idea.  I'll give this a whirl locally to check the configuration.

ybur-yug:
Thanks 👯 👯 👯

nscyclone:
We could probably move all the translated lessons into `lessons` along with the new `en` language folder. What do you think about this?

doomspork:
I like that @nscyclone, keeps the root folder clean.

One issue I mentioned to @ybur-yug is that this will generate new URLs and the previously crawled ones will now 404, so we need to handle that.

ybur-yug:
@doomspork @nscyclone I'm for doing all of them at once into a `lessons` dir. Gets it done and there is very little chance we'll have to deal with the change again.

doomspork:
Let's do that than @ybur-yug

ybur-yug:
@doomspork do you have a timeline on when you'd like to do this? Happy to help out with the grunt work involved for crawlers etc

doomspork:
@ybur-yug no timeline, just need to make sure this PR covers our bases which it does not at the moment.  Can't merge and auto-deploy without a solution in place to address the _massive_ number of 404 and broken links.

Feel free to investigate and try some solutions out.

doomspork:
I've been swamped lately with work so I haven't had a chance to further investigate how to handle all these redirects, @ybur-yug have you had looked into this any further?

ybur-yug:
@doomspork I've been busy as all hell lately too. I briefly looked into it but haven't gotten anywhere noteworthy.

ybur-yug:
I'm probably going to have a hard time contributing much of anything of significance for a few weeks with the travel/work schedule I've got going right now, honestly. So if someone else wants to investigate further that would be awesome, or I can pick it up around the 20th when things slow down for me.

nscyclone:
I suppose that the solution is just to set some simple 301 redirect rules.
If I am not mistaken, we use Cloudflare's nginx as the production server. I am not very familiar with it but this [article](https://support.cloudflare.com/hc/en-us/articles/200170536-How-do-I-redirect-all-visitors-to-HTTPS-SSL-) might be helpful.
```bash
# Translations
http://elixirschool.com/([\w]{2})/lessons/(.*) → http://elixirschool.com/lessons/$1/$2
# Original lessons
http://elixirschool.com/lessons/([\w]{3,})/([\w]+) → http://elixirschool.com/lessons/en/$1/$2
```

doomspork:
@ybur-yug / @nscyclone — as nice as this change would be, I think the logistics to pull it off are posing to be quite significant as-is.  Can we close this and evaluate it at a later date?

ybur-yug:
So many updates have happened I'm just going to close this. Would be easier to do from scratch when we have a better plan.

PR #530, opened 2016-07-18
Add note to `CONTRIBUTING.md` about handling addtions' translations

ybur-yug:
There is currently no documented preference on how best to handle making sure the translations all stay in sync with the main english version. This note gives the example of a simple PR I did [here](https://github.com/doomspork/elixir-school/issues/529) and shows a process I just mimicked from @doomspork. This hopefully will allow us to keep better track of the state of all guides.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,10 @@
 # Contribution Guidelines
+
+## General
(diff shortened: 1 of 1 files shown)


doomspork:
Thank you @ybur-yug 👍

PR #533, opened 2016-07-19
Fixed DETS example

jswny:
I've fixed a small error in the DETS example which could cause confusion.


nscyclone:
Thanks, @jswny! 
I suppose that this should also be applied to the translations. Could you take a look at them, please?

jswny:
@nscyclone Sure, I can do that.

doomspork:
Thank you @jswny 👍

PR #534, opened 2016-07-19
Improve Enum basics section

ybur-yug:
This includes:
- several additions to examine functions and enumerate over them
- a dive into built in documentation with some helpers
The reasoning here originally started when I wanted to double-check the
accuracy of the count of functions that was listed.
Running

```
Enum.__info___(:functions) |> Enum.count
```

yields `82`, which is less than 100, so I clarified the language.
But then I realized some of the other tidbits I could do now that I had
a list of functions and their arity via the `__info__` call, which led to
the changes seen.

### DIFF
--- a/lessons/basics/enum.md
+++ b/lessons/basics/enum.md
@@ -31,6 +31,7 @@ at/3
 ...
 ```
 
+Using this, its clear that we have a vas amount of functionality, and that is for a clear reason.
 Enumeration is at the core of functional programming and is an incredibly useful thing.
 By leveraging it combined with other perks of Elixir, such as documentation being a first class citizen as we just saw, it can be incredibly empowering to the developer as well.
 
--- a/lessons/basics/enum.md
+++ b/lessons/basics/enum.md
@@ -31,7 +31,7 @@ at/3
 ...
 ```
 
-Using this, its clear that we have a vas amount of functionality, and that is for a clear reason.
+Using this, its clear that we have a vast amount of functionality, and that is for a clear reason.
 Enumeration is at the core of functional programming and is an incredibly useful thing.
 By leveraging it combined with other perks of Elixir, such as documentation being a first class citizen as we just saw, it can be incredibly empowering to the developer as well.
 
--- a/lessons/basics/enum.md
+++ b/lessons/basics/enum.md
@@ -14,14 +14,14 @@ A set of algorithms for enumerating over collections.
 
 The `Enum` module includes nearly 100 functions for working with the collections we learned about in the last lesson.
 
-This lesson will only cover a subset of the available functions, however we can actually enamine them ourselves.
+This lesson will only cover a subset of the available functions, however we can actually examine them ourselves.
 Let's do a little experiment in IEx.
 
 ```elixir
-$ iex
-iex(1)> Enum.__info__(:functions) |> Enum.each(fn({function, arity}) ->
-...(1)>   IO.puts "#{function}/#{arity}"
-...(1)> end)
+iex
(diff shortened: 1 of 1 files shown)


lessons/basics/enum.md
```diff
@@ -12,9 +12,94 @@ A set of algorithms for enumerating over collections.
...
+at/3
+...
+```
+We can take this even further.
+By enumerating in this manner we can even get the documentation for every function of a given module by leveraging the `h` operator, which prints docs.
+
+```elixir
+$ iex
+iex(1)> Enum.__info__(:functions) |> Enum.each(fn({function, arity}) ->
+...(1)>   Code.eval_string("require IEx.Helpers; IEx.Helpers.h Enum.#{function}")
```
ybur-yug: We should probably add a note to NEVER USE THIS lol

lessons/basics/enum.md
```diff
@@ -12,9 +12,95 @@ A set of algorithms for enumerating over collections.
 
 ## Enum
 
-The `Enum` module includes over one hundred functions for working with the collections we learned about in the last lesson.
+The `Enum` module includes nearly 100 functions for working with the collections we learned about in the last lesson.
 
-This lesson will only cover a subset of the available functions, to see a complete set of functions visit the official [`Enum`](http://elixir-lang.org/docs/stable/elixir/Enum.html) docs; for lazy enumeration use the [`Stream`](http://elixir-lang.org/docs/stable/elixir/Stream.html) module.
+This lesson will only cover a subset of the available functions, however we can actually enamine them ourselves.
```
doomspork: "enamine" -> "examine"

lessons/basics/enum.md
```diff
@@ -12,9 +12,95 @@ A set of algorithms for enumerating over collections.
...
+Let's do a little experiment in IEx.
+
+```elixir
+$ iex
```
doomspork: Please drop the `$ iex` and the numbers from the `iex` prompt, we have a style guide that tries to capture some of these: https://github.com/doomspork/elixir-school/wiki/Lesson-Styleguide

ybur-yug: Didn't see that before. Thanks and will abide 👍

lessons/basics/enum.md
doomspork: I like the first example of enumerating the functions and their arity but I think this one is unnecessary.  For one, the output from this huge and only a very small portion of it even stays within my terminal buffer.  Lastly, this is a Basic lesson and this code is well beyond that level and may confuse more than it helps.

lessons/basics/enum.md
doomspork: No worries!  Also on my TODO list is to clean-up the README and call better attention to the contributing resources.

lessons/basics/enum.md
ybur-yug: Agreed upon further thought.

Fun idea: Write a function that generates the entirety of the elixir documentation and prints it to a file or generates a PDF using pandoc with this technique iterating through all std library modules.

ybur-yug:
@doomspork totally took me a year but finally updated this after forgetting for far too long.

ybur-yug:
I can also create an issue for translations if you think this looks good to merge.

lessons/basics/enum.md
```diff
@@ -12,9 +12,30 @@ A set of algorithms for enumerating over collections.
...
+all?/1
+all?/2
+any?/1
+any?/2
+at/2
+at/3
+...
+```
+
+Using this, its clear that we have a vas amount of functionality, and that is for a clear reason.
```
doomspork: "vas amount" -> "vast amount"

PR #541, opened 2016-07-20
update phoenix_slim deps lesson 9: Mix

yepesasecas:
phoenix_slim deps changed, Im just updating Mix Lesson to show new deps.


doomspork:
Thank you @yepesasecas 👍

Issue #529, opened 2016-07-18
Update `Enum.chunk_by` in translated guides


ybur-yug:
PR #528 updated the `Enum.chunk_by/2` section of the basics/enum lesson to be a bit more clear.

You can see the exact changes in English [here](https://github.com/doomspork/elixir-school/pull/528/commits/cdc2d2bec63ab9f7759ee8ba3c0a09ed38f41809).
## Update

**This has since been included to show the arity, see [here](https://github.com/doomspork/elixir-school/compare/master...improvement/add_arity_to_enum_chunk_by_basics#diff-da9b21c2c73f7335cedc97aa18d36b0cR53).**
## Translation

Translation progress:
- [x] en
- [x] ru /cc @brain-geek / @nscyclone
- [x] ko /cc @devleoper 
- [x] jp /cc @3100 
- [ ] no /cc @therobinsj 
- [x] pl /cc @Koziolek 
- [ ] sk /cc @michalvalasek 
- [ ] fr /cc @julienXX
- [x] cn /cc @cizixs 
- [x] es /cc @aleandros / @jonahoffline / @edwinallenz
- [ ] pt /cc @pragmaticivan 
- [ ] vi /cc @joneslee85 
- [ ] it /cc @lucidstack  / @sformisano
- [ ] my /cc @alkaha
- [ ] id /cc @rezaprima

nscyclone:
It seems like `it` is mentioned twice in the task list.

ybur-yug:
@nscyclone woops! Fixing

ybur-yug:
@nscyclone fixed

nscyclone:
I wonder whether we should mention the functions' arities in the [references](https://github.com/doomspork/elixir-school/pull/528/files#diff-da9b21c2c73f7335cedc97aa18d36b0cR53) and [headers](https://github.com/doomspork/elixir-school/pull/528/files#diff-da9b21c2c73f7335cedc97aa18d36b0cR51)? They are used in the [Strings lesson](https://github.com/doomspork/elixir-school/blob/master/lessons/basics/strings.md#length1) and [sometimes in this one](https://github.com/doomspork/elixir-school/blob/master/lessons/basics/enum.md#chunk).
What do you think, @ybur-yug, @doomspork?

ybur-yug:
@nscyclone since its core to the nature of the language itself I think this was an oversight by me. I will update.

ybur-yug:
[...]

PR #546, opened 2016-07-22
[sk] Update `Enum.chunk_by` in translated guides #529

michalvalasek:


sk/lessons/basics/enum.md
```diff
@@ -49,11 +49,11 @@ Táto funckia má niekoľko možností použitia, pozrite sa do oficiálnej doku
 
 ### chunk_by
 
-Ak potrebujete rozdeliť kolekciu do menších na základe niečoho iného, než ich veľkosť, použite funkciu `chunk_by`:
+Ak potrebujete rozdeliť kolekciu do menších na základe niečoho iného, než ich veľkosť, použite funkciu `chunk_by`. Ako argumenty funkcia berie kolekciu (Enum) a funkciu - ak sa zmení jej výstup, ukončí sa aktuálna podkolekcia a začne sa vytvárať ďalšia:
 
 ```elixir
-iex> Enum.chunk_by(["one", "two", "three", "four", "five"], fn(x) -> String.length(x) end)
```
nscyclone: As far as I know we didn't remove the [first example](https://github.com/doomspork/elixir-school/blob/master/lessons/basics/enum.md#chunk_by) from the original lesson after this improvement.

ybur-yug: We did not. I added both because the intention was to show that when calling `Enum.chunk_by/2` the grouping begins anew when the `function` argument changes its returned result. Though using almost-the-same list may not have been the perfect decision to change it again and back-propogate to all other translation branches would be a much bigger pain.

michalvalasek: Sorry, I did a mistake when reading the original diff. Should I add the first example back? TBH I think the second example is sufficient.

ybur-yug: I agree but it would be hard to go back and remove all the original first examples from the other translations and eventually they may end up mis-synced. But if the people responsible for that piece that have already done the changes to catch up dont mind changing it, then its no problem. I'm not sure how many have started on this small bit. From the chart it looks like only a few but I'm thinking in broad safe-rather-than-sorry terms.

doomspork: Let's keep the first example so each translation is as close as possible to what is on master.  If we decide to remove this example later we can.

doomspork:
Thank you @michalvalasek 👍

PR #547, opened 2016-07-23
Fix a typo in the `specifics/mnesia` lesson

nscyclone:


lessons/specifics/mnesia.md
```diff
@@ -115,7 +115,7 @@ We define the columns using the atoms `:id`, `:name`, and `:job`. When we execut
 
 ## The Dirty Way
 
-First of all we will look at the dirty way of reading and writing to an Mnesia table. This should generally be avoided as success is not guaranteed, but it should help us learn and become comfortable working with Mnesia. Let's add some entries to our **Person** table.
+First of all we will look at the dirty way of reading and writing to a Mnesia table. This should generally be avoided as success is not guaranteed, but it should help us learn and become comfortable working with Mnesia. Let's add some entries to our **Person** table.
```
ybur-yug: I don't actually know the proper English rule for this. But, I can see both sides of this. By simple rule, `an` is before a word that starts with a vowel. So in this sense, that is proper. But, by pronunciation aloud it is `am-nē-ZHə` which sounds much akin to `amnesia`. I think since `Mnesia` is not a proper English word, it may make more sense to utilize the phonetically correct rather than the grammatically correct version here, but I am not an expert. It just 'feels' right to me. Heh. Thoughts, anyone?

nscyclone: Thanks, @ybur-yug! 
Yes, 'an' is used before words which sound like they start with a vowel even if their first letter is a consonant. But I think that there is no starting `a` in this case. `Mnesia` and [`amnesia`](https://en.wikipedia.org/wiki/Amnesia) are different words with the opposite meanings. I decided to watch several videos about Mnesia to find out the most commonly used case but it didn't really help that much: it seems like there are many options occurring with equal frequency like `əmnɪʒə`, `mnɪʒə`,  `nɪʒə` and so on. If we pronounce `Mnesia` as `əmnɪʒə`, how would we pronounce `amnesia` then?
Maybe @doomspork can shed some light on it.

doomspork: Wow, this is tough.  I looked online and both Google and the official docs use "a Mnesia ...".

I'm going to merge this one 👍

doomspork:
Thank you @nscyclone, your help is really appreciated 😀

PR #554, opened 2016-07-26
Update elixir version to 1.3.2, OTP 20 and erts: 8.0.1

alexwbuschle:


doomspork:
Thanks @alexwbuschle 👍

PR #557, opened 2016-07-26
Changed 3.41 to 3.14 to reflect numerical representation of pi

mercul3s:
This looked like a typo, so I thought I would submit a pull request to fix it. I'm finding the lessons and structure very useful overall!


ybur-yug:
@mercul3s I believe this was just chosen as a random floating point number, and not specifically supposed to be pi or anything. If that was the case we would use `:math.pi/0`. @doomspork correct me if I'm wrong.

doomspork:
@mercul3s is correct, it fits with the different pie types mentioned in the examples. I can't believe I made that mistake, someone is going to revoke my :neckbeard: card.

We don't want to use a function @ybur-yug, it's an example number.  We don't need the _actual_ number pi.

Thanks @mercul3s 👍

ybur-yug:
I was thinking of it from the few lines I worked on in a change earlier on that piece. Woops. 👍

mercul3s:
Glad I could help!

PR #562, opened 2016-07-27
[sk] Add basics lesson Mix Tasks.

michalvalasek:
I skipped the first example with Phoenix because I found it quite confusing for the reader, since we haven't been talking about Phoenix so far. Also IMO it doesn't add anything substantial to the rest of the lesson.


doomspork:
Thank you @michalvalasek

Issue #338, opened 2016-05-01
Lesson about custom types and @spec


Koziolek:
Probably basic, but I'm not sure. 

Short summary:
- the `@spec` and how to define "interfaces"
- Creating custom types
  - Why we need it - tuples are too generic 
  - `@type` and how to use it
  - short exercise - create custom type of email that will be keyword of `[{:name NAME}, {:domain, DOMAIN}]` with set of functions like `toEmail` and `fromEmail`. 

@doomspork @pjhampton any suggestions?

doomspork:
I've wanted to do a typespec lesson for awhile!  I think this is a good idea.

pjhampton:
Yep. I also love this idea. Do you want me to write this lesson or do you want to take it @Koziolek?

Koziolek:
I'm writing this. Should be ready to review in next few days.

doomspork:
Awesome @Koziolek!  Looking forward to it.

Koziolek:
In https://github.com/doomspork/elixir-school/pull/564

@doomspork, @nscyclone  guys, could you take a look and put your comments.

doomspork:
Thank you @Koziolek, I will look today 👍

pjhampton:
I like this @Koziolek. Once we firm up the semantics, I will help you fine tune the English. Thanks for contributing as usual ❤️

nscyclone:
Closing this one because the original lesson is already merged (#564) and the translation task list is in another issue (#850).

PR #564, opened 2016-07-28
Issue #338 spectypes

Koziolek:
@doomspork, @nscyclone please take a look.

### DIFF
--- a/lessons/advanced/typespec.md
+++ b/lessons/advanced/typespec.md
@@ -33,80 +33,8 @@ def sum_product(a) do
 end
 ```
 
-Everything looks ok and when we call valid result will be return, but function `Enum.sum` returns `number` not `integer` as we expected in `@spec`. It could be source of bugs!
-
-### Dialyxir and Dialyzer
-
-The [Dialyzer](http://erlang.org/doc/man/dialyzer.html), a **DI**screpancy **A**nal**YZ**er for **ER**lang programs is a tool for static code analysis. In other words they _read_ but not _run_ code and analyse it e.g. looking for some bugs, dead, unnecessary or unreachable code.
-   
-The [Dialyxir](https://github.com/jeremyjh/dialyxir) is a mix task simplify usage of Dialyzer in Elixir.  
-
-Specification helps tools like Dialyzer to understand code better. Unlike documentation that is readable and understandable only for other humans (if only exists and is good written), `@spec` use more formal syntax and could be understand by machine.
-
-Let's add Dialixyr to our project. The simples way is to add dependency to `mix.exs` file:
+Everything looks ok and when we call valid result will be return, but function `Enum.sum` returns `number` not `integer` as we expected in `@spec`. It could be source of bugs! There are tools like Dialyzer to static analysis of code that helps us to find this type of bugs. We will talk about them in another lesson.
  
-```elixir
-defp deps do
-  [{:dialyxir, "~> 0.3.5", only: [:dev]}]
-end
-```
-
-Then we call:
-
-```shell
-$ mix deps.get
-...
-$ mix deps.compile
-```
-
-First command download and install Dialyxir. You could be asked for installing Hex. Second compile Dialyxir app. If you want to install Dialyxir as global app, please read [documentation](https://github.com/jeremyjh/dialyxir#installation).
-
-Last step is to run Dialyzer to rebuild PLT. You need to do this every time after installation of new version of Erlang or Elixir. Thanks to this Dialyzer will not be try to analyse standard library every time that you try to use it. It takes few minutes.
-
-```shell
-$ mix dialyzer.plt
-Starting PLT Core Build ... this will take awhile
-dialyzer --build_plt --output_plt /.dialyxir_core_18_1.3.2.plt --apps erts kernel stdlib crypto public_key -r /Elixir/lib/elixir/../eex/ebin /Elixir/lib/elixir/../elixir/ebin /Elixir/lib/elixir/../ex_unit/ebin /Elixir/lib/elixir/../iex/ebin /Elixir/lib/elixir/../logger/ebin /Elixir/lib/elixir/../mix/ebin
-  Creating PLT /.dialyxir_core_18_1.3.2.plt ...
-...
- done in 5m14.67s
-done (warnings were emitted)
-```
-
-### Static analysis of code
- 
-Now, when we have Dialyxir on board we could run it:
- 
-```shell
-$ mix dialyzer
-...
-examples.ex:3: Invalid type specification for function 'Elixir.Examples':sum_times/1. The success typing is (_) -> number()
-...
-```
-
-Communicate is clear. Return type of our function `sum_times/1` is different than declared. That because `Enum.sum` return `number` not `integer` and effective return type of `sum_times/1` is `number`. Because `number` is not `integer` so we have error. How to fix it? We could add `round` function that change `number` to `integer`:
-
-```elixir
-@spec sum_times(integer) :: integer
-def sum_times(a) do
-    [1, 2, 3]
-    |> Enum.map(fn el -> el * a end)
-    |> Enum.sum
-    |> round
-end
-```
-
-Finally:
-
-```shell
-$ mix dialyzer
-...
-  Proceeding with analysis... done in 0m0.95s
-done (passed successfully)
-```
-
-Using specifications with tools to static code analysis helps us to make code that is self tested and contains less bugs.  
-
 ## Custom types
 
 Writing specification is nice, but sometimes our functions works with more complex data structures than simple numbers or collections. In that cases definition in `@spec` could be hard to understand and very hard to change by other developers. Sometimes functions needs to take many parameters or return complex data. Long parameters list is one of bad smells in code. In object oriented languages like Ruby or Java we could easily define classes that helps us to solve this problems. Elixir hasn't classes but because is easy to extends that we could define our types.
--- a/lessons/advanced/typespec.md
+++ b/lessons/advanced/typespec.md
@@ -37,7 +37,7 @@ Everything looks ok and when we call valid result will be return, but function `
  
 ## Custom types
 
-Writing specification is nice, but sometimes our functions works with more complex data structures than simple numbers or collections. In that cases definition in `@spec` could be hard to understand and very hard to change by other developers. Sometimes functions needs to take many parameters or return complex data. Long parameters list is one of bad smells in code. In object oriented languages like Ruby or Java we could easily define classes that helps us to solve this problems. Elixir hasn't classes but because is easy to extends that we could define our types.
+Writing specifications is nice, but sometimes our functions works with more complex data structures than simple numbers or collections. In that definition's case in `@spec` it could be very hard to understand and/or change for other developers. Sometimes functions need to take in a large number of parameters or return complex data. A long parameters list is one of many potential bad smells in one's code. In object oriented-languages like Ruby or Java we could easily define classes that help us to solve this problem. Elixir hasn't classes but because is easy to extends that we could define our types.
   
 Out of box Elixir contains some basic types like `integer` or `pid`. You  can find full list of available types in [documentation](http://elixir-lang.org/docs/stable/elixir/typespecs.html#types-and-their-syntax).
  
@@ -57,12 +57,12 @@ def sum_times(a, params) do
 end
 ```
 
-We introduce struct in `Examples` module that contains two fields `first` and `last`. That is simpler version of struct from `Range` module. About structures we talk when discussing [modules](lessons/basics/modules/#structs). Lets imagine that we need to specification with `Examples` struct many times. It will be very annoying and could be source of bugs. Solution of this problem is `@type`.
+We introduced a struct in `Examples` module that contains two fields `first` and `last`. That is simpler version of struct from `Range` module. We will talk about `structs` when we get into discussing [modules](lessons/basics/modules/#structs). Lets imagine that we need to specification with `Examples` struct in many places. It would be very annoying to write long, complex specifications and could be a source of bugs. A solution to this problem is `@type`.
  
 Elixir has three directives for types:
 
   - `@type` – simple, public type. Internal structure of type is public. 
-  - `@typep` – type is private and could be used only in module where is defined. 
+  - `@typep` – type is private and could be used only in the module where is defined. 
   - `@opaque` – type is public, but internal structure is private. 
 
 Let define our type:
@@ -79,9 +79,9 @@ defmodule Examples do
 end
 ```
 
-We defined type `t(first, last)` that is representation of struct `%Examples{first: first, last: last}`. At this point we see types could takes parameters, but we defined type `t` as well and this time it is representation of struct `%Examples{first: integer, last: integer}`.   
+We defined the type `t(first, last)` already, which is a representation of the struct `%Examples{first: first, last: last}`. At this point we see types could takes parameters, but we defined type `t` as well and this time it is a representation of the struct `%Examples{first: integer, last: integer}`.   
 
-What is a difference? First one represents struct `Examples` of which the two keys could be any type. Second one represents struct which keys are `integers`. That means code like this:
+What is a difference? First one represents the struct `Examples` of which the two keys could be any type. Second one represents struct which keys are `integers`. That means code like this:
   
 ```elixir
 @spec sum_times(integer, Examples.t) :: integer
@@ -111,7 +111,7 @@ end
 
 ### Documentation of types
 
-Last element that we need to talk about is how to document our types. As we know from [documentation](/lessons/basic/documentation) lesson we have `@doc` and `@moduledoc` annotations to create documentation for functions and modules. For documenting our types we can use `@typedoc`:
+The last element that we need to talk about is how to document our types. As we know from [documentation](/lessons/basic/documentation) lesson we have `@doc` and `@moduledoc` annotations to create documentation for functions and modules. For documenting our types we can use `@typedoc`:
 
 ```elixir
 defmodule Examples do
--- a/lessons/advanced/typespec.md
+++ b/lessons/advanced/typespec.md
@@ -72,14 +72,16 @@ defmodule Examples do
 
     defstruct first: nil, last: nil
 
-    @type t :: %Examples{first: integer, last: integer}
-
     @type t(first, last) :: %Examples{first: first, last: last}
+
+    @type t :: %Examples{first: integer, last: integer}
 
 end
 ```
 
-We defined two types. First `t` is representation of `%Examples{first: integer, last: integer}`. Second is `t(first, last)` that is an alias to `%Examples{first: first, last: last}`. What is a difference? First one represents struct `Examples` of which the two keys are `integer`. Second one represents struct which keys could has any type. That means code like this:
+We defined type `t(first, last)` that is representation of struct `%Examples{first: first, last: last}`. At this point we see types could takes parameters, but we defined type `t` as well and this time it is representation of struct `%Examples{first: integer, last: integer}`.   
+
+What is a difference? First one represents struct `Examples` of which the two keys could be any type. Second one represents struct which keys are `integers`. That means code like this:
   
 ```elixir
 @spec sum_times(integer, Examples.t) :: integer
--- a/lessons/advanced/typespec.md
+++ b/lessons/advanced/typespec.md
@@ -109,7 +109,7 @@ end
 
 ### Documentation of types
 
-Last element that we need to talk about is how to document our types. As we know from [documentation](/lessons/basic/documentation) lesson we have `@doc` and `@moduledoc` annotations to create documentation for functions and modules. For documenting ou types we can use `@typedoc`:
+Last element that we need to talk about is how to document our types. As we know from [documentation](/lessons/basic/documentation) lesson we have `@doc` and `@moduledoc` annotations to create documentation for functions and modules. For documenting our types we can use `@typedoc`:
 
 ```elixir
 defmodule Examples do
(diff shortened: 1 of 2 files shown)


lessons/basics/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+
+It's not uncommon you would like to describe interface of your function. Of course You can use [@doc annotation](/lessons/basic/documentation), but it is only information for other developers that is not checked in compilation time. For this purpose Elixir has `@spec` annotation to describe specification of function that will be checked by compiler.
+
+However in some cases specification is going to be quite big and complicated. If you would like to reduce complexity, you want to introduce custom type definition. Elixir has `@type` annotation for that. In the other hand, Elixir is still dynamic language. That means all information about type will be ignored by compiler, but could be used by other tools.   
+
+## Specification
+
+If you have experience with Java or Ruby you could think about specification as an `interface`. Specification defines what should be type of function parameters and return value.
+
+To define input and output types we use `@spec` directive placed right before function definition and taking as a _params_ name of function, list of parameter types, and after `::` type of return value.
```
ybur-yug: I believe for this since _params_ represents an argument it should be backticked over italicized. ie `params`

lessons/basics/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+def sum_product(a) do
+    [1, 2, 3]
+    |> Enum.map(fn el -> el * a end)
+    |> Enum.sum
+end
+```
+
+Everything looks ok and when we call valid result will be return, but function `Enum.sum` returns `number` not `integer` as we expected in `@spec`. It could be source of bugs!
+
+### Dialyxir and Dialyzer
```
ybur-yug: I believe we have plans to break out Dialixir/Dialyzer to their own lesson module. This might fit better elsewhere. But is also very relevant here. Hmm. Thoughts @doomspork?

lessons/basics/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+
+The [Dialyzer](http://erlang.org/doc/man/dialyzer.html), a **DI**screpancy **A**nal**YZ**er for **ER**lang programs is a tool for static code analysis. In other words they _read_ but not _run_ code and analyse it e.g. looking for some bugs, dead, unnecessary or unreachable code.
```
devleoper: Github diff preview seems broken from this line..

Koziolek: That is only in github preview.

lessons/basics/typespec.md
doomspork: That was originally the plan.  @Koziolek would you be interested in pulling the Dialyzer info out into it's own lesson?

lessons/basics/typespec.md
```diff
@@ -0,0 +1,197 @@
+---
+layout: page
+title: Specifications and types
+category: basics
```
doomspork: I think we should classify this as an "advanced" lesson.  You don't _need_ to have specs and types to work with Elixir, it comes after you've grasped the basics.

ybur-yug: I agree. Even if coming from something like Java and seeking this out immediately, its far from anything required to function.

I think things like the `h` helper that access docs/specs/types are wonderful for the basics, but actually making them and leveraging what they offer etc is something I've never seen come as a core competency for anyone who is just starting to learn.

lessons/basics/typespec.md
Koziolek: OK. Will drop it just to one line info - That we have tools to static analysis

lessons/basics/typespec.md
Koziolek: moved

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+```elixir
+@spec sum_times(integer) :: integer
+def sum_times(a) do
+    [1, 2, 3]
+    |> Enum.map(fn el -> el * a end)
+    |> Enum.sum
+    |> round
+end
+```
+
```
pjhampton: Somewhere around here might be a good place to have a table like this one found in the erlang docs?

http://erlang.org/doc/reference_manual/typespec.html

I often find that there is one for many things so I don't have to roll my own spec :)

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+   
+The [Dialyxir](https://github.com/jeremyjh/dialyxir) is a mix task simplify usage of Dialyzer in Elixir.  
+
+Specification helps tools like Dialyzer to understand code better. Unlike documentation that is readable and understandable only for other humans (if only exists and is good written), `@spec` use more formal syntax and could be understand by machine.
+
+Let's add Dialixyr to our project. The simples way is to add dependency to `mix.exs` file:
+ 
+```elixir
+defp deps do
+  [{:dialyxir, "~> 0.3.5", only: [:dev]}]
```
pjhampton: `only: [:dev]` it concerns me that we haven't discussed this at all in Elixir School. Maybe you could leave a small explanation of what this is / does?

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,197 @@
+---
+layout: page
+title: Specifications and types
+category: advanced
+order: 9
+lang: en
+---
+
+If you want to define 'interface' like in Java or Ruby, Elixir contains syntax for that. More, if you want to define your custom type you can do that too. In this lesson we will learn about `@spec` and `@type` syntax.
```
pjhampton: Let's get ride of this `think of ruby or java` mentality. I think this is a problem in the Elixir community where we keep making assumptions about each other and other languages.

It would be nice to replace this with something like your sentence on line 44: `Unlike documentation that is readable and understandable only for other humans (if only exists and is good written), `@spec` use more formal syntax and could be understand by machine.`

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,197 @@
+---
+layout: page
+title: Specifications and types
+category: advanced
```
pjhampton: Hmm, sorry I was late to the party when you changed this. I think it might be a good idea to change this to specific as it makes use of an external lib and fits with erlang. Dialyzer fits much closer with Mnesia, ETS / DETS, etc than gen_server, error handling etc.

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+```elixir
+@spec sum_product(integer) :: integer
+def sum_product(a) do
+    [1, 2, 3]
+    |> Enum.map(fn el -> el * a end)
+    |> Enum.sum
+end
+```
+
+Everything looks ok and when we call valid result will be return, but function `Enum.sum` returns `number` not `integer` as we expected in `@spec`. It could be source of bugs!
```
pjhampton: Is this sentence too premature?

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+    defstruct first: nil, last: nil
+
+    @type t :: %Examples{first: integer, last: integer}
+
+    @type t(first, last) :: %Examples{first: first, last: last}
+
+end
+```
+
+We defined two types. First `t` is representation of `%Examples{first: integer, last: integer}`. Second is `t(first, last)` that is an alias to `%Examples{first: first, last: last}`. What is a difference? First one represents struct `Examples` of which the two keys are `integer`. Second one represents struct which keys could has any type. That means code like this:
```
pjhampton: I'm not sure I understand this to be honest.

lessons/advanced/typespec.md
Koziolek: Link to documentation line 44.

pjhampton: I think I mean the something similar to this:
```
Type :: any()                 %% The top type, the set of all Erlang terms
        | none()                %% The bottom type, contains no terms
        | pid()
        | port()
        | reference()
        | []                    %% nil
        | Atom
        | Bitstring
        | float()
        | Fun
        | Integer
        | List
        | Map
        | Tuple
        | Union
        | UserDefined           %% described in Type Declarations of User-Defined Types

  Atom :: atom()
        | Erlang_Atom           %% 'foo', 'bar', ...

  Bitstring :: <<>>
             | <<_:M>>          %% M is a positive integer
             | <<_:_*N>>        %% N is a positive integer
             | <<_:M, _:_*N>>

  Fun :: fun()                  %% any function
       | fun((...) -> Type)     %% any arity, returning Type
       | fun(() -> Type)
       | fun((TList) -> Type)

  Integer :: integer()
           | Erlang_Integer                    %% ..., -1, 0, 1, ... 42 ...
           | Erlang_Integer..Erlang_Integer    %% specifies an integer range

  List :: list(Type)                           %% Proper list ([]-terminated)
        | maybe_improper_list(Type1, Type2)    %% Type1=contents, Type2=termination
        | nonempty_improper_list(Type1, Type2) %% Type1 and Type2 as above
        | nonempty_list(Type)                  %% Proper non-empty list

  Map :: map()                                 %% stands for a map of any size
       | #{}                                   %% stands for the empty map
       | #{PairList}

  Tuple :: tuple()                             %% stands for a tuple of any size
         | {}
         | {TList}

  PairList :: Pair
            | Pair, PairList

  Pair :: Type := Type                         %% denotes a pair that must be present
        | Type => Type

  TList :: Type
         | Type, TList

  Union :: Type1 | Type2```

We need to prevent linking people off the website unless necessary

File: lessons/advanced/typespec.md (hunk 7 above)
Comment (doomspork) [inline lessons/advanced/typespec.md]: Actually this was brought up during OpenCamps.  We should _definitely_ try to move away from comparing Elixir to Ruby and Phoenix to Rails, it does the community a disservice.  I take full blame for that, my original lessons made lots of references.

File: lessons/advanced/typespec.md (hunk 8 above)
Comment (doomspork) [inline lessons/advanced/typespec.md]: @pjhampton that is a valid point.  @Koziolek what do you think?

File: lessons/advanced/typespec.md (hunk 9 above)
Comment (doomspork) [inline lessons/advanced/typespec.md]: Maybe we could re-word this:

"Everything looks good and our new `sum_product/1` function returns a valid result, but is that all?  If we look at `Enum.sum/1` we'd find that it returns a `number` and not a `integer` as our `@spec` implies, this could be source of bugs!"

File: lessons/advanced/typespec.md (hunk 6 above)
Comment (nscyclone) [inline lessons/advanced/typespec.md]: The lesson about `Mix` is ninth in the `basics` section. I suppose we could just explain dependencies there more widely.

File: lessons/advanced/typespec.md (hunk 8 above)
Comment (nscyclone) [inline lessons/advanced/typespec.md]: I think if we will move the `Dialyxir/Dialyzer` part to another lesson in the `specifics` section the remaining could still be a part of `basics`.

File: lessons/advanced/typespec.md [hunk 11]
```diff
@@ -0,0 +1,197 @@
...
+    
+    @typedoc """
+        Type that represents Examples struct with :first as integer and :last as integer.
+    """
+    @type t :: %Examples{first: integer, last: integer}
+
+end
+```
+
+Directive `@typedoc` is similar to `@doc` and `@moduldoc`.
```
nscyclone: `@moduldoc` → `@moduledoc`

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+    end
+       |> Enum.map(fn el -> el * a end)
+       |> Enum.sum
+       |> round
+end
+```
+
+### Documentation of types
+
+Last element that we need to talk about is how to document our types. As we know from [documentation](/lessons/basic/documentation) lesson we have `@doc` and `@moduledoc` annotations to create documentation for functions and modules. For documenting ou types we can use `@typedoc`:
```
nscyclone: `ou` → `our`

lessons/advanced/typespec.md
doomspork: I agree with @nscyclone.  Let's revisit the Mix lesson and see where we could work this in.

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,197 @@
...
+
+Then we call:
+
+```shell
+$ mix deps.get
+...
+$ mix deps.compile
+```
+
+First command download and install Dialyxir. You could be asked for installing Hex. Second compile Dialyxir app. If you want to install Dialyxir as global app, please read [documentation](https://github.com/jeremyjh/dialyxir#installation).
```
doomspork: How about something like this? 

"The first command will download and install Dialyxir; be sure to select yes if you're asked to install Hex. Our second command will compile our dependencies which include the Dialyxir app.  To install Dialyxir globally, see the official documentation on [Dialyxir](https://github.com/jeremyjh/dialyxir#installation)."

lessons/advanced/typespec.md
doomspork: This is a toughy.  I'd prefer to not copy-paste the official docs when we can refrain from doing so but @pjhampton raises a valid concern RE: taking people away from the site.

doomspork:
Thanks again for doing this @Koziolek, it's a _great_ lesson!

lessons/advanced/typespec.md
Koziolek: @doomspork, `typespec` in `advanced` and Dialyzer/Dialyxir in `specifics`, because typespec is an element of language, but D&D are tools.

lessons/advanced/typespec.md
Koziolek: all stuff about D/D gone.

lessons/advanced/typespec.md
Koziolek: Need to rewrite this. After few days I know what I wanted to write, but i wrote it in terrible way.

lessons/advanced/typespec.md
doomspork: Sounds good @Koziolek 👍

lessons/advanced/typespec.md
Koziolek: Rewritend

doomspork:
@pjhampton / @devleoper / @ybur-yug — Any addition feedback or is this good to merge?

Thank you again @Koziolek, this is really great content.

pjhampton:
LGTM 👍

ybur-yug:
@doomspork giving it another read this morning & will update in a bit

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Introduction 
+
+It's not uncommon you would like to describe interface of your function. Of course You can use [@doc annotation](/lessons/basic/documentation), but it is only information for other developers that is not checked in compilation time. For this purpose Elixir has `@spec` annotation to describe specification of function that will be checked by compiler.
+
+However in some cases specification is going to be quite big and complicated. If you would like to reduce complexity, you want to introduce custom type definition. Elixir has `@type` annotation for that. In the other hand, Elixir is still dynamic language. That means all information about type will be ignored by compiler, but could be used by other tools.   
+
+## Specification
+
+If you have experience with Java or Ruby you could think about specification as an `interface`. Specification defines what should be type of function parameters and return value.
```
ybur-yug: I know were straying away from the comparisons to Ruby, and that this also throws in Java's hat, but should we remove it and just explain the concept of an interface agnostically if its useful as an analogy?

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,127 @@
...
+    |> Enum.map(fn el -> el * a end)
+    |> Enum.sum
+end
+```
+
+Everything looks ok and when we call valid result will be return, but function `Enum.sum` returns `number` not `integer` as we expected in `@spec`. It could be source of bugs! There are tools like Dialyzer to static analysis of code that helps us to find this type of bugs. We will talk about them in another lesson.
+ 
+## Custom types
+
+Writing specification is nice, but sometimes our functions works with more complex data structures than simple numbers or collections. In that cases definition in `@spec` could be hard to understand and very hard to change by other developers. Sometimes functions needs to take many parameters or return complex data. Long parameters list is one of bad smells in code. In object oriented languages like Ruby or Java we could easily define classes that helps us to solve this problems. Elixir hasn't classes but because is easy to extends that we could define our types.
```
ybur-yug: Grammar/spelling suggestions:

`Writing specification`

-->

`Writing specifications`

================================

`In that cases definition in @spec could be hard to understand and very hard to change by other developers.`

-->

`In that definition's case in @spec it could be very hard to understand and/or change for other developers.`

================================

`Sometimes functions needs to take many parameters or return complex data.`

-->

`Sometimes functions need to take in a large number of parameters or return complex data.`

================================

`Long parameters list is one of bad smells in code.`

-->

`A long parameters list is one of many potential bad smells in one's code.`

================================

`In object oriented languages like Ruby or Java we could easily define classes that helps us to solve this problems.`

-->

`In object oriented-languages like Ruby or Java we could easily define classes that help us to solve this problem.`

================================

This one I just cant make sense of:

`Elixir has no concept of classes or object-orientation but because it is easy to extends that we could define our types.`

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,127 @@
...
+    for i <- params.first..params.last do
+        i
+    end
+       |> Enum.map(fn el -> el * a end)
+       |> Enum.sum
+       |> round
+end
+```
+
+We introduce struct in `Examples` module that contains two fields `first` and `last`. That is simpler version of struct from `Range` module. About structures we talk when discussing [modules](lessons/basics/modules/#structs). Lets imagine that we need to specification with `Examples` struct many times. It will be very annoying and could be source of bugs. Solution of this problem is `@type`.
```
ybur-yug: `We introduce struct in `Examples` module that contains two fields `first` and `last`.`

-->

`We introduced a struct in our `Examples` module that contained two fields: ``` `first`, and `last`. ```

===================

`About structures we talk when discussing [modules](lessons/basics/modules/#structs).`

-->

```We will talk about `structs` when we get into discussing [modules](lessons/basics/modules/#structs).```

====================

I cant reason through what this is saying:

```Lets imagine that we need to specification with `Examples` struct many times. It will be very annoying and could be source of bugs. Solution of this problem is `@type`.```

===========================

```It will be very annoying and could be source of bugs. Solution of this problem is `@type`.```

-->

```It would be very annoying and could be a source of bugs. A solution to this problem is `@type`.```

File: lessons/advanced/typespec.md [hunk 17]
```diff
@@ -0,0 +1,127 @@
...
+ 
+Elixir has three directives for types:
+
+  - `@type` – simple, public type. Internal structure of type is public. 
+  - `@typep` – type is private and could be used only in module where is defined.
```
ybur-yug: `only in module` --> `only in the module`

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,127 @@
...
+    defstruct first: nil, last: nil
+
+    @type t(first, last) :: %Examples{first: first, last: last}
+
+    @type t :: %Examples{first: integer, last: integer}
+
+end
+```
+
+We defined type `t(first, last)` that is representation of struct `%Examples{first: first, last: last}`. At this point we see types could takes parameters, but we defined type `t` as well and this time it is representation of struct `%Examples{first: integer, last: integer}`.
```
ybur-yug: ```We defined type `t(first, last)` that is representation ```

lessons/advanced/typespec.md
```diff
@@ -0,0 +1,127 @@
...
+    @type t(first, last) :: %Examples{first: first, last: last}
+
+    @type t :: %Examples{first: integer, last: integer}
+
+end
+```
+
+We defined type `t(first, last)` that is representation of struct `%Examples{first: first, last: last}`. At this point we see types could takes parameters, but we defined type `t` as well and this time it is representation of struct `%Examples{first: integer, last: integer}`.   
+
+What is a difference? First one represents struct `Examples` of which the two keys could be any type. Second one represents struct which keys are `integers`. That means code like this:
```
ybur-yug: ```represents struct `Examples` ```

-->

```represents the struct `Examples` ```

File: lessons/advanced/typespec.md [hunk 20]
```diff
@@ -0,0 +1,127 @@
...
+    end
+       |> Enum.map(fn el -> el * a end)
+       |> Enum.sum
+       |> round
+end
+```
+
+### Documentation of types
+
+Last element that we need to talk about is how to document our types. As we know from [documentation](/lessons/basic/documentation) lesson we have `@doc` and `@moduledoc` annotations to create documentation for functions and modules. For documenting our types we can use `@typedoc`:
```
ybur-yug: ```Last element that we need to talk about is how to document our types. ```

-->

```The last element that we need to talk about is how to document our types. ```

ybur-yug:
@doomspork @Koziolek Went through and added various corrections/suggestions.

ybur-yug:
PS @Koziolek amazing work; please dont be offended that I was being quite nitpicky going through gramatically. I just want to make sure every section makes sense and wont have elements lost in translation to the other langs :)

devleoper:
New lesson looks great to me!

lessons/advanced/typespec.md
doomspork: This isn't really a comparison though, is it?  It's more of an association.  I think associating something to a reader's prior experience can help them grasp the topic.

ybur-yug: Thats true. I dont personally think it was detracting in any way but I know its a topic that had come up before.

Koziolek:
Fix grammar issues. 

@ybur-yug, about references to OO languages, IMO many people came here from OO world and looking for some "mental hooks". Types are not classes but many OO devs could thing about them in that way.

ybur-yug:
@Koziolek I agree. See @doomspork's comment above. I just knew that there had been some struggle in the earlier days comparing _too much_ to Ruby to the point that it made things harder to understand. But that doesn't seem to be the case here. Awesome work 👍 💯 :)

doomspork:
Thank you @Koziolek!!

PR #571, opened 2016-08-01
[KO] Fix a misused word in ETS lesson

Dalgona:
@devleoper, @malkoG, @riseshia,
I think we need to use the word "프로세스(process)" instead of "프로세서(processor)", am I correct?


riseshia:
I think that's miss typo. thanks XD

devleoper:
Thanks guys!

PR #573, opened 2016-08-01
[RU] Make the language kind explicit

gmile:
It's a _programming language_ we're talking here! :)


brain-geek:
Nice catch, @gmile . Thanks!

PR #576, opened 2016-08-02
[sk] Fix best practices example in `basics\pipe-operator` (#574)

michalvalasek:


doomspork:
Thank you @michalvalasek!

PR #580, opened 2016-08-03
Get rid of duplicate example with hex integers

brain-geek:
Right now we have hex example in both code blocks for 'integer' part of 'basics' lesson.

IMO, this may confuse some people who don't know about hex notation. And anyway, we cover this in the next code snippet.


ybur-yug:
LGTM

PR #582, opened 2016-08-04
[SK] Fix snippets in basics lesson

brain-geek:
For some reason, those examples are a bit corrupted - iex output is not present there.


doomspork:
Thank you @brain-geek 👍

Issue #583, opened 2016-08-05
Suggestion: Explain how functions are named early.


scouten:
In http://elixirschool.com/lessons/basics/collections/, which is early in the process, you start referring to functions by their arity (i.e. `++/2` when talking about List Concatenation). As a newbie to the language, that syntax is completely unfamiliar (except that I have read a couple of other introductions to the language). IMHO it would help to work in a reference to that syntax early on (or at least mention that you'll get to it later).
## Translation

Translation progress:
- [x] en #599
- [x]  ru /cc @brain-geek / @nscyclone #737
- [x]  ko /cc @devleoper
- [x]  jp /cc @hinamiyagk @3100 #735
- [ ]  no /cc @therobinsj
- [x]  pl /cc @Koziolek #849
- [ ]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [x]  cn /cc @cizixs
- [ ]  es /cc @aleandros / @jonahoffline / @eze1981
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @lucidstack  / @sformisano
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima

doomspork:
Thank you @scouten, I think this is  _great_ suggestion!  Would you be interested in taking a shot at adding this?

scouten:
@doomspork: Yes, I'll take a stab at that. Might be a couple of days before I can get to it.

scouten:
Turns out I had some time today. See PR #599.

nscyclone:
Shouldn't we update this issue with a translations task list?
The English updates were merged back in August, the [Japanese translation is pending](https://github.com/doomspork/elixir-school/pull/735).

doomspork:
Yes @nscyclone, we should

devleoper:
In Korean translation, PRs by @marocchino have already merged.

PR #599, opened 2016-08-05
Describe how Elixir names functions. Addresses #583.

scouten:


doomspork:
Thank you @scouten, this is great!   I'll look it over and let you know if I have any feedback.

lessons/basics/collections.md
```diff
@@ -40,6 +40,9 @@ iex> [1, 2] ++ [3, 4, 1]
 [1, 2, 3, 4, 1]
 ```
 
+A side note about the name (`++/2`) format used above: In Elixir (and Erlang, upon which Elixir is built), a function or operator name has two components: the name you give it (here `++`) and the number of arguments it takes (two in this case), combined with a slash. (The number of arguments is often referred to as "arity.") We'll talk more about this later; this knowledge will help you understand the notation for now.
```
ybur-yug: I personally think this should emphasize the word `arity` in priority over explaining it in plain terms. The reasoning being that arity is core to the entirety of notation in Elixir/Erlang and it will not be going anywhere. I think this implies that moreso than the way it reads now. I think how I would word it is:

```
A side note about the name (`++/2`) format used above: In Elixir (and Erlang, upon which Elixir is built), a function or operator name has two components: the name you give it (here `++`) and its `arity`. Arity is a core part of speaking about Elixir (and Erlang) code. It is the number of arguments a given function takes (two in this case), combined with a slash. We'll talk more about this later; this knowledge will help you understand the notation for now.
```

lessons/basics/functions.md
```diff
@@ -93,6 +93,29 @@ iex> Length.of [1, 2, 3]
...
 
+### Function Naming and Arity
+
+We mentioned earlier that functions are named by the combination of given name and arity (number of arguments). This means you can do things like this:
+
+```elixir
+defmodule Greeter2 do
+  def hello(), do: "Hello, anonymous person!"   # hello/0
+  def hello(name), do: "Hello, " <> name        # hello/1
+  def hello(name1, name2), do: "Hello, #{name1} and #{name2}"
```
ybur-yug: Should `name1` and `name2` be snakecased? I dont know the convention off the top of my head and poking around a few lessons I couldnt find an example to compare, but just noticed.

lessons/basics/collections.md
scouten: Makes sense. Just added your wording to the PR.

lessons/basics/functions.md
scouten: I'm new to the language, so I don't really know what convention should be followed. Happy to reword if someone can clarify.

ybur-yug: @doomspork has the final say :)
I remember when I first came in, reading through the official docs they defined it quite early and explicitly and there was little ambiguity, so I just personally thought that style works well. But obviously thats just my opinion.

doomspork: In general I refrain from using numbers within my variables names.  I would probably use something like `name_one` and `name_two` for this.

ybur-yug: Line 101 would fall under that as well then if thats preferred (and update the naming throughout) @scouten @doomspork

nscyclone:
@ybur-yug, @doomspork, there are some changes applied since your last review, could you take another peek? I think it looks good.

doomspork:
Thank you @scouten and @nscyclone, I'll check later today and let you know!

ybur-yug:
LGTM

doomspork:
Thank you again @scouten 👍

PR #604, opened 2016-08-07
Change non-existent 'ErlangVM' term to its real name

brain-geek:
This PR removes any mentiones of non-existent 'ErlangVM' term and replaces it with its real name - BEAM. While this may be not in line with some language grammars, we stop introducing wrong facts.


brain-geek:
Special thanks to @Slotos who pointed out this fact. :)

doomspork:
Thank you @brain-geek and @Slotos 👍

nscyclone:
@brain-geek, I've just found that BEAM is still called `Erlang VM` [here](https://github.com/doomspork/elixir-school/blob/56304171480f86ccfeee89c15d61dfc7d8655bed/ru/lessons/advanced/concurrency.md#L17) in both the original lesson and many translations. Is this made by purpose or should be fixed?

brain-geek:
@nscyclone It's made on purpose - to limit amount of new terms for readers.

PR #605, opened 2016-08-07
Use proper term in 'collections' lesson

brain-geek:
`|` is not a pipe operator, but cons operator. We're discussing this in #600 .

If we merge this - we'll need to update all translations.

PR #609, opened 2016-08-08
[RU] Use a proper term in the `basics/collections` lesson

nscyclone:
Related to #605.
[`оператор cons`](https://books.google.ru/books?id=autYAwAAQBAJ&pg=PA47&lpg=PA47&source=bl&ots=1WmjzllbaG&sig=cG3VjRdWCODqOgoOPL24UH15xIU&hl=en&sa=X&ved=0ahUKEwjxnKL5hLHOAhWEhSwKHf08CFwQ6AEITDAH#v=onepage&q&f=false) is used in the `Erlang Programming` book translation.
@brain-geek, I suppose we shouldn't add this to our glossary. What do you think about this?


ru/lessons/basics/collections.md
```diff
@@ -60,7 +60,7 @@ iex> tl [3.14, :pie, "Apple"]
 [:pie, "Apple"]
 ```
 
-Того же результата можно добиться с использованием оператора `|`. Мы его часто встретим в последующих уроках.
+Того же результата можно добиться с использованием оператора cons: `|`. Мы его часто встретим в последующих уроках.
```
Slotos: > использованием оператора cons **—** `|`

brain-geek:
Yeah, let's not create additional terms - luckily it's used only once in our translation.

BTW, do we really need this change? Original sounds better and I don't think beginning developers need to know this.

Slotos:
"Cons" is almost a proper noun the way it is. It is a shortening of a "construct", that has been used in a very specific environment. It could probably be translated as "оператор построения", if not for implications that get lost. It constructs not just anything, but a very specific structure.

If anything, Russian LISP documentation should be consulted for consistency. But as far as my googling took me, they didn't bother translating a function name.

nscyclone:
Thanks, @Slotos, I replaced a colon with an em-dash.
I couldn't find any translation for this in the LISP docs. They just call it `функция cons`. I'd suggest to stick with `оператор cons` or introduce a new term `оператор построения списков`. What do you guys think?

Slotos:
To summarize, I think `оператор cons` is quite sufficient at this stage. Operator name is not at fault for `cons` confusion.

---------------------------------------

Thought process:

Cons operator constructs a cons cell. A minimal non-empty list in cons notation `[ 1 | [] ]` basically reads as "a cons cell with value "1", that links to (addresses) an empty cell". Given that in Erlang/Elixir cons cells seem to be used exclusively for lists, something like `оператор связывания` might work, but it loses the implication of building a very particular data structure yet again.

At this point I do think that expanding `collections.md` with cons cells information will provide better learning, than operator name localization. A few examples or exercises amount to better knowledge incorporation, than even the best translated name would.

We could use LISP specs and documentation as a starting point:
- [Dotted pair Notation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Dotted-Pair-Notation.html) describes exactly is what in Elixir looks like `[a | b]`.
- [Lists and Conses](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node28.html) \[[Russian translation](http://filonenko-mikhail.github.io/cltl2-doc/ru/clmse6.html)\] describes conses in context of lists. Exactly the way Erlang uses them, including dotted lists:
```elixir
iex(1)> [1 | [2 | 3]]
[1, 2 | 3]
```

nscyclone:
@brain-geek, what is your opinion?

brain-geek:
@nscyclone , I think that current option is good enough.

If there are no objections from @Slotos , I think we can merge this one.

Issue #608, opened 2016-08-07
Using proper term in 'collection' - cons operator


brain-geek:
Changes made to the original lesson could be found in #605 .
## Translation

Translation progress:
- [x] en
- [x]  ru /cc @nscyclone
- [x]  ko /cc @devleoper
- [x]  jp /cc @3100
- [x]  no /cc @therobinsj
- [x]  pl /cc @Koziolek
- [x]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [x]  cn /cc @cizixs
- [x]  es /cc @aleandros / @jonahoffline / @eze1981
- [x]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @lucidstack  / @sformisano
- [ ]  ms /cc @alkaha
- [ ]  id /cc @rezaprima

Koziolek:
Done :)

brain-geek:
Thanks, @Koziolek !

PR #611, opened 2016-08-08
[SK] Using proper term in 'collection' - cons operator #608

michalvalasek:
And translate term "linked lists".


doomspork:
Thank you @michalvalasek 👍

PR #614, opened 2016-08-08
[JP] Fix best practices example in pipe operator lesson

hinamiyagk:
#574


doomspork:
Thank you @hinamiyagk 👍

PR #623, opened 2016-08-17
[KO] Using proper term in Collection lesson

malkoG:
cc: @Dalgona @devleoper @riseshia 
ref: #608  

`cons` operation *itself* is so difficult to explain easily, I added some reference to comprehension.
Any feedback is welcome 😹 😹 😹


riseshia:
LGTM...?!

devleoper:
Nice! Also +1 for the translator comment!

By the way, is there any way to show editorial or translational comments in some neat way? It will be great to add a comment without adding any additional paragraph.

Dalgona:
maybe we can just wrap the comment in parentheses, or we can use blockquote block...

devleoper:
We don't have to be worry about touching the words.

and.. linking Korean documentations will highly improve this original text especially for further reading. is there any documentation about `cons` written in Korean?

malkoG:
Hmmm... How about this.....?????? @devleoper @riseshia @Dalgona

devleoper:
I think the examples below explains well what does it do. Just keeping it simple or making a simple note(that it came from LISP) would be simple and helpful.

and.. an Erlang book translated into Korean translated something like `cons` to "생성자"(constructor): used the full meaning.
I think that is enough and wonder whether this make sense to other Korean readers. What do you think about this?

malkoG:
Well? 
I think translating `cons` to "생성자"(constructor) can be confusing. 
Although `cons` is abbreviated word for constructor, you can see in this context, `cons` operates like `prepend`. And, in many articles explaing OOP, constructor is different from `cons`. I'm worried about OOP exprienced users would be confused.

riseshia:
I agree to @malkoG, it's better to add some [real example](https://github.com/doomspork/elixir-school/pull/623/commits/4255edb71c193e9eb5f5dbe65d2e3e0b5a9984f0#diff-983ce0e559ebecd5b2fee0a287858a0dR67) if we use the term `constructor`.

devleoper:
Right, the same term means (both languages) totally different paradigms. :joy:

.. and the example that @malkoG added isn't kind of redundant thing. Actually, adding some example with more `|` looks better and more specific. that means... it shouldn't be ended in merely an editorial or translational comment. this examples doesn't improve only Korean translated texts. I think we can make this be part of the document, How about makign a PR to English text?

malkoG:
Oh, that's coooooool idea!

malkoG:
😂 😂 😂 😂 😂 
```iex
iex(8)> [["asd"|[["a"]]]|[["a"]]]
[["asd", ["a"]], ["a"]]
iex(9)> [["asd"|[["a"]]]|["a"]]
[["asd", ["a"]], "a"]
iex(10)> ["asd"|[["a"]]|["a"]]
** (CompileError) iex:10: undefined function |/2
    (stdlib) lists.erl:1354: :lists.mapfoldl/3
    (stdlib) lists.erl:1355: :lists.mapfoldl/3
iex(10)> ["asd"|[["a"]]]|["a"]]
** (SyntaxError) iex:10: unexpected token: ]

iex(10)> [["asd"|[["a"]]]|["a"]]
[["asd", ["a"]], "a"]
iex(11)> [["asd"|["a"]]|["a"]]
[["asd", "a"], "a"]
iex(12)> ["asd"|["a"]]
["asd", "a"]
iex(13)> ["a"|[["asd"|["a"]]]]
["a", ["asd", "a"]]
iex(14)> ["a"|["asd"|["a"]]]
["a", "asd", "a"]
iex(15)> ["a"|["a"|["asd"|["a"]]]]
["a", "a", "asd", "a"]
iex(16)> [1|[2|[3|nil]]]
[1, 2, 3 | nil]
iex(17)> [1|[2|[3|[]]]]
[1, 2, 3]
iex(18)> [1|[2|[3|[4|[]]]]]
[1, 2, 3, 4]
iex(19)> [1|[2|[3|[4|[5|[]]]]]]
[1, 2, 3, 4, 5]
```

nscyclone:
Not sure if this might be helpful but we had a [discussion](https://github.com/doomspork/elixir-school/pull/609) about translating `cons operator` to Russian and decided to translate only the word `operator`.  Probably Korean translations for Erlang or LISP docs and books may shed some light on this.

malkoG:
@nscyclone Thank you for your opinion 😄 😄 😄

devleoper:
We'll think about this later.

PR #640, opened 2016-08-29
[JP] Use proper term in 'collections' lesson

hinamiyagk:
#608


doomspork:
Thank you again @hinamiyagk 👍

PR #650, opened 2016-09-01
[en] Fix Typo

marocchino:


nscyclone:
Thanks, @marocchino! I've just noticed it and was going to propose a fix.

PR #653, opened 2016-09-04
[en] Fix link

riseshia:
not basic, but `basics`


doomspork:
Good catch @riseshia 👍

PR #661, opened 2016-09-06
[NO] Updated with proper terminology and substraction of duplicate lists for basics/collections lesson

therobinsj:
Reference issues #608 and  #612


nscyclone:
Thanks, @therobinsj!

PR #665, opened 2016-09-09
Update message passing in Concurrency section [resolve 664]

ipmsteven:


lessons/advanced/concurrency.md
```diff
@@ -64,6 +66,8 @@ iex> send pid, :ok
 :ok
 ```
 
+You may notice that the `listen` function is recursive, if we remove the recursive call in the end, the process will exit right after it finish handling the first message, so we use a recursive call to make the process able to handle multiple messages.
```
doomspork: @ipmsteven how about something like this:

"You may notice that the `listen/0` function is recursive, this allows our process to handle multiple messages.  Without recursion our process would exit after handling the first message."

ipmsteven: @doomspork updated in	487d71f

doomspork:
Thank you @ipmsteven 👍

PR #666, opened 2016-09-09
Fix command in generating child project in Umbrella section

ipmsteven:
@doomspork 
Do all the projects in `apps` need to be a `sup` project? If then, we would also need to update the command in generating `datasets` app, if we do that on purpose, maybe we would add some explanation on why `datasets` app is an exception?

### DIFF
--- a/gr/lessons/advanced/umbrella-projects.md
+++ b/gr/lessons/advanced/umbrella-projects.md
@@ -70,7 +70,7 @@ You can use "mix" to compile it, test it, and more:
 Run "mix help" for more commands.
 
 
-$ mix new datasets
+$ mix new datasets --sup
 
 * creating README.md
 * creating .gitignore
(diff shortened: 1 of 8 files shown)


doomspork:
Thanks @ipmsteven!  I can take a peek at this later today.  In the mean time @pjhampton do you mind checking this out since you wrote the original lessons?

pjhampton:
Hey @ipmsteven 

Super thanks for the feedback and the PR! 😃  This time, I'm **not** going to except it, however I think you have a valid point regarding this that could inspire another PR. The reason behind _datasets_ not having a supervisor is simply because it holds data and has no functionality per se. Sorry - being a data scientist I made a bad assumption about the reads knowledge even though correct.

I think you should submit a PR removing the `--sup` flag from the two other sub-apps. As this isn't a lesson on mix command and would make more sense to the reader.

Cheers!
pj

doomspork:
Thank you @ipmsteven and @pjhampton 👍

PR #669, opened 2016-09-10
[es] Some fixes on spanish translation

fleveque:
I'm just starting to learn Elixir and found your site, so I thought it would be good to help a little if I can :)


doomspork:
Thank you @fleveque 👍

PR #671, opened 2016-09-12
[CN] Translate comprehensions into Chinese

cizixs:
As the title says, this commit translates `comprehensions` lesson in `basics` section  to Chinese.


doomspork:
Thank you @cizixs 💜

PR #678, opened 2016-09-17
Add assert_receive, capture_io and capture_log to testing

holsky:
I saw that there is nothing on assert_receive yet. I am not sure how much this is actually used when writing tests for Elixir applications, but if it is I think it is worth mentioning since it is different from other language's xUnit toolkit.


doomspork:
This is great @holsky!  There's two other functions we consider adding to this: `capture_log` and `capture_io`, what do you think?

holsky:
Great idea, should I do them all in this PR and rename it to "add assert_receive, capture_log and capture_io to testing"?

holsky:
I amended my previous commit and did a force push, so that the resulting history is cleaner (only 1 commit in total). Is that the way you work here?

doomspork:
Thank you @holsky!  This is great.

PR #676, opened 2016-09-14
#590 add information about @spec annotation in documentation lesson

Koziolek:

PR #681, opened 2016-09-22
[JP] Add information about @spec annotation in documentation lesson

hinamiyagk:
#676


doomspork:
Thank you @hinamiyagk

PR #692, opened 2016-09-27
Update all bulgarian lessons to latest english

AgIt0:
I noticed there were some paragraphs missing here and there in the already translated bulgarian lessons, so I added them. Also did some fixups on a few sentences/paragraphs to make them clearer


doomspork:
Thank you @AgIt0!

PR #705, opened 2016-10-06
Additional reduce string example with accumulator

luzeduardo:
- Added an extra example with strings and accumulator to explain correct order that reducer works


doomspork:
Thank you @luzeduardo 👍

PR #709, opened 2016-10-08
Minor improvements for basic lessons

brain-geek:
Recently I've discovered a `proselint` tool and wanted to try it out.

Here is the first batch of changes suggested by the tool. While minor, I think that these changes make text better.

If this PR goes OK - I'll create a few more with fixes for other lessons and a CI check for it.


doomspork:
Thanks @brain-geek!

Issue #701, opened 2016-10-04
Update for 1.4 Enum functionalities/deprecations


ybur-yug:
PR #696 updated the Enumerable section to clarify the new functionalities of `min`/`max` now having an arity 2 version, the deprecation of `uniq/1`, and the addition of `map_every/3`.
# NOTE

**As elixir 1.4 is not yet released, please submit translation PR's to point at `improvement/elixir-1.4`**
## Translation

Translation progress:
- [x] en
- [x]  ru /cc @brain-geek / @nscyclone #708 
- [x]  ko /cc @devleoper #942
- [ ]  jp /cc @3100
- [x]  no /cc @therobinsj #713 
- [x]  pl /cc @Koziolek #715 
- [ ]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [ ]  cn /cc @cizixs
- [ ]  es /cc @aleandros / @jonahoffline / @eze1981
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @lucidstack  / @sformisano
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima

doomspork:
Thank you @ybur-yug!

PR #713, opened 2016-10-10
[NO] Updated the Enum lesson with elixir 1.4 changes

therobinsj:
Referenced in #701


doomspork:
Thank you @therobinsj!

ybur-yug:
@doomspork not a big deal but just something I noticed --

These translation branches have PR'd to master rather than my original branch with the english version.

Because of this, in production the translated sections explain/use features that currently do not exist in the basics/enum section now because of this, as 1.4 is not yet released.

I don't think this is a _huge_ deal, but thought it was worth noting as I just thought of it/noticed now.

Issue #700, opened 2016-10-04
Clarify simple_one_for_one strategy for Supervisor in translations


ybur-yug:
PR #695 updated the section on Supervisors to include some additional notes, we should update the translations to include these changes.
## Translation

Translation progress:
- [x] en
- [x]  ru /cc @brain-geek / @nscyclone
- [x]  ko /cc @devleoper
- [ ]  jp /cc @3100
- [ ]  no /cc @therobinsj
- [x]  pl /cc @Koziolek #714 
- [ ]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [ ]  cn /cc @cizixs
- [ ]  es /cc @aleandros / @jonahoffline / @eze1981
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @lucidstack  / @sformisano
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima

brain-geek:
Thanks for creating this one, @ybur-yug !
I'll take russian translation.

PR #721, opened 2016-10-13
[PT] Update translation for supervisors

anderkonzen:
Related to #700


anderkonzen:
Thanks @ybur-yug! I'll try to keep helping here whenever I can!

PR #724, opened 2016-10-13
[PT] Update atom section in basic lesson

anderkonzen:
Related to #480


doomspork:
@pragmaticivan / @capaci — would either of you mind looking this over when you have a moment? 😀

PR #702, opened 2016-10-06
Added new lesson for poolboy

ayayalar:
New lesson for poolboy.

PR #431, opened 2016-06-01
Initial Guardian basic

hassox:
This is just an initial PR to get some feedback.

PR #729, opened 2016-10-15
Add the new section for libraries

nscyclone:
During discussion in #702 we came to a decision that we need a new separate section for libraries.
This PR adds the mentioned section and it also hides the sections without lessons from the sidebar.
cc @doomspork @brain-geek @ybur-yug @Koziolek and everyone else.

Once this will be merged, #702 and #431 will require corresponding updates (location, order and category). Then I'll also create an issue for translations.


brain-geek:
Let's merge it with at least one lesson in this category. e.g. after #702 gets merged.

Koziolek:
Looks nice, let't merge it with first lesson plus add some info about library vs specific, because sometimes is hard to say what is what.

ybur-yug:
Looks like we need to update the configuration for the new ruby version on Heroku to get this one synced properly with master now

nscyclone:
Thanks @ybur-yug!
Actually the conflict was caused by another set of changes. Anyway, I merged latest master into this branch and resolved it, so the latter one is mergeable again.

PR #731, opened 2016-10-16
Fix the markup in the latest changes of the `specifics/mnesia` lesson

nscyclone:
The nested unordered list in the last paragraph didn't look as expected.


nscyclone:
Thanks, @ybur-yug. Fixed the only existing Polish translation.

doomspork:
Thank you @nscyclone and @ybur-yug 👍

PR #743, opened 2016-10-20
Fix typo in first concurrency paragraph

zaid:
The first line in the Concurrency paragraph reads:
```
One the selling points of Elixir is its support for concurrency. Thanks to the Erlang 
```
Instead of
```
One of the selling points of Elixir is its support for concurrency. Thanks to the Erlang 
```


ybur-yug:
Good catch. Thanks @zaid!

Issue #642, opened 2016-08-29
Translate Improvement from PR #612 On Duplicates & List Subtraction


ybur-yug:
Changes made to the original lesson could be found in #612.
## Translation

Translation progress:
- [x] en
- [x]  jp /cc @hinamiyagk 
- [x]  es /cc @aleandros / @jonahoffline / @eze1981
- [x]  ru /cc @nscyclone
- [x]  ko /cc @devleoper
- [ ]  no /cc @therobinsj
- [x]  pl /cc @Koziolek
- [ ]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [x]  cn /cc @cizixs
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @lucidstack  / @sformisano
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima

PR #745, opened 2016-10-20
[Cn] Describe function name format

cizixs:
This commit solves #583 and #642 , plus update other collections content.


doomspork:
Thank you @cizixs!

PR #747, opened 2016-10-20
[CN] translate pipe operator

cizixs:
This newly translated article also solves #574


doomspork:
@cizixs, thank you for your help!

PR #754, opened 2016-10-21
Use relative links in all lessons

brain-geek:
When reviewing the #736 I've noted that this lesson uses an absolute link to other lessons.

I think that having those links relative will help translators a bit + we already use this approach in the majority of links across the site.


nscyclone:
Nice catch, @brain-geek!
I've also found ~30 occurrences of absolute paths. Shouldn't we fix them all in a single shot? If so, searching for `)[/` might be a good start.
This could also be mentioned in `CONTRIBUTING` when it will be updated.

brain-geek:
@nscyclone Thanks for the idea, I'll look into it and change the PR accordingly.

brain-geek:
Commited also all other names.

Looks like the majority of those links were in places where english version had absolute links.

Not sure whether we should add one more point to `CONTRIBUTING`.

doomspork:
This is awesome @brain-geek!

brain-geek:
@doomspork Does this look OK for you to merge? :)

doomspork:
@brain-geek done!

PR #755, opened 2016-10-21
Use relative URLs in index translation links

brain-geek:
Right now when using local site you may unintentionally get to https://elixirschool.com/ via those links.

That's a minor inconvenience, but this fix may be helpful to save some time on debugging. 

Also, looks like we have those links on only two translations - other translations just don't have this block. Any ideas whether it's intentional?


doomspork:
I imagine this is unintentionally, not too sure how we ended up with these differences.  If we can verify there's no impact on the generated links then I think these are good to merge.

brain-geek:
@doomspork I've clicked through few random ones. Everything worked.

I've also tried running linkchecker, but there are a lot of 404s due to abscent pages.

doomspork:
Thanks @brain-geek!

I'd love to find a solution for the 404s at some point I'm just not sure if it's feasible with Jekyll.

PR #751, opened 2016-10-21
[KO] Update translation

marocchino:
@Dalgona, @malkoG, @devleoper, @riseshia review please.

FYI. I use this command for diff:

~~~
git diff f8baebe..bcb0548 -- lessons/
~~~

Also, includes this [review](https://github.com/doomspork/elixir-school/pull/726#discussion_r83355310).

PR #760, opened 2016-10-22
[KO] Fix the syntax highlighting in the `basics/basics` lesson

nscyclone:


nscyclone:
My bad, already fixed in #751.

PR #762, opened 2016-10-24
[RU] Fix basics/collections.md comparison link

basavind:


ybur-yug:
This one looks really simple and was going to merge because it looked right to me, but CC @nscyclone / @brain-geek just to verify :)

PR #767, opened 2016-10-30
Genstage Lesson Proposal on top of @doomspork's original

ybur-yug:
## See prior thread [here](https://github.com/doomspork/elixir-school/pull/759)


doomspork:
Thanks @ybur-yug, I'm going to go through this now.

doomspork:
@ybur-yug I think this is great but still think there's a lot we can take off that doesn't fit with our current lesson format (the whole Postgres portion for instance).  I've gone ahead and pulled your changes to begin editing them.  What I'd like to do is merge this, apply my changes, and push it up for a team wide review.  Thoughts?

ybur-yug:
@doomspork that works for me. I'll keep the origin bit I built up as I have it now for my own stuff and might just make it a blog post and link to the lesson once its up or something. Its bigger in scope but I think it might have some use elsewhere. So have at it.

doomspork:
Thanks @ybur-yug!  I think the additional sections you added would be great for a blog post as they're quite useful and informative.

PR #777, opened 2016-11-05
update GenStage lesson introduction

revdan:
The new `GenStage` tutorial is great! But it promises to find the primes up to who knows how high of a number, and the sample code checks for even numbers.

In lieu of a good solution to reverse-engineer primes - unless I have misunderstood the description of course - I changed the introduction text to match the example code.

I can submit a more computationally expensive function than checking for even integers if necessary, but this seemed like a good stopgap so that the code isn't misleading.

Thanks!

### DIFF
--- a/lessons/advanced/gen-stage.md
+++ b/lessons/advanced/gen-stage.md
@@ -44,7 +44,7 @@ Now that we've covered the roles within GenStage let's start on our application.
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


doomspork:
Thank you @revdan!  This is what I had intended.  Originally I was going to do primes but realized it was probably best to stick with evens so the focus would be on GenStage more so than the implementation.

revdan:
Cool, thanks!

Issue #598, opened 2016-08-05
Fix description of UTF-8 encoding


nscyclone:
Changes made to the original lesson could be found [here](https://github.com/doomspork/elixir-school/pull/597/files).
## Translation

Translation progress:
- [x] en
- [x]  ru /cc @nscyclone
- [x]  ko /cc @devleoper
- [x]  jp /cc @3100
- [ ]  no /cc @therobinsj
- [x]  pl /cc @Koziolek
- [x]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [ ]  cn /cc @cizixs
- [ ]  es /cc @aleandros / @jonahoffline / @eze1981
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @lucidstack  / @sformisano
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima

julienXX:
FYI Basic/Strings lesson is not yet translated in French.

doomspork:
Thanks for the heads up @julienXX

PR #779, opened 2016-11-06
[CN] translate basics string lesson

cizixs:
This commit translate `basics/strings` to Chinese, thus solving #598


doomspork:
Thank you again @cizixs 👍

PR #782, opened 2016-11-07
Fix a wrong arity of `fun2ms/1` in ets.md

aquarhead:


doomspork:
Thank you @aquarhead 👍

PR #788, opened 2016-11-08
Remove compiler warning

blogscot:
Compiling 1 file (.ex)
warning: the variable "word" is unsafe as it has been set inside a case/cond/receive/if/&&/||. Please explicitly return the variable value instead. For example:

case int do
  1 -> atom = :one
  2 -> atom = :two
end
should be written as

atom =
  case int do
    1 -> :one
    2 -> :two
  end
Unsafe variable found at:
lib/build_basics.ex:19

### DIFF
--- a/cn/lessons/advanced/escripts.md
+++ b/cn/lessons/advanced/escripts.md
@@ -60,8 +60,7 @@ defmodule ExampleApp.CLI do
   end
 
   defp response({opts, word}) do
-    if opts[:upcase], do: word = String.upcase(word)
-    word
+    if opts[:upcase], do: String.upcase(word), else: word
   end
 end
 ```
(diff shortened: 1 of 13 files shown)


doomspork:
@blogscot could you make the Elixir code change to the example across all languages?

blogscot:
@doomspork Sure. 

What's the best way of doing this? Should I create a new pull request for the other languages, or close this one and put everything in a single pull request?

doomspork:
@blogscot you can make those changes as part of this PR 👍

blogscot:
@doomspork I've updated the other language files that contain this advanced escript page.

Incidentally, I had been thinking up creating a PR for the GenStage page, but after I realised it would be pulled into this PR I undid all the changes. It seems this PR still has a single line change to GenStage, which seems to be a missing CR at the end of the file. I hope that isn't a problem.

doomspork:
Awesome @blogscot, thank you for your help!  No worries about the modification to GenStage.

I look forward to your next PR 👍

PR #790, opened 2016-11-09
Add guardian

brain-geek:
I'm opening new PR for this, as #431 is pointing to old version of branch.

I've rebased it and applied some feedback. My changes are added in separate commit.

Let's have some discussion and merge it soon! :)

### DIFF
--- a/lessons/specifics/guardian.md
+++ b/lessons/specifics/guardian.md
@@ -2,7 +2,7 @@
 layout: page
 title: Guardian (Basics)
 category: specifics
-order: 6
+order: 7
 lang: en
 ---
 
@@ -12,21 +12,20 @@ lang: en
 
 ## JWTs
 
-JWT provides a rich token for authentication. Where many authentication systems provide access to only a subject identifier for the resource, JWTs provide this along with other information like:
+A JWT can provide a rich token for authentication. Where many authentication systems provide access to only a subject identifier for the resource, JWTs provide this along with other information like:
 
 * Who issued the token
 * Who is the token for
 * Which system should use the token
 * What time was it issued
-* What time should you consider it valid
 * What time does the token expire
 
-In addition to these fields Guardian provides some other fields to facilitate additional functionality.
+In addition to these fields Guardian provides some other fields to facilitate additional functionality:
 
 * What type is the token
 * What permissions does the bearer have
 
-These are just the basic fields in a JWT. You're free to add any additional information that your application requires. Just remember to keep it short. The JWT has to fit in the HTTP header.
+These are just the basic fields in a JWT. You're free to add any additional information that your application requires. Just remember to keep it short, as JWT has to fit in the HTTP header.
 
 This richness means that you can pass JWTs around in your system as a fully contained unit of credentials. 
 
@@ -42,8 +41,7 @@ JWT tokens can be used to authenticate any part of your application.
 * Inter-process
 * 3rd Party access (OAuth)
 * Remember me functionality
-* Other interfaces - raw TCP, UDP etc
-* Command line (if you _really_ want)
+* Other interfaces - raw TCP, UDP, CLI, etc
 
 JWT tokens can be used everywhere in your application where you need to provide verifiable authentication.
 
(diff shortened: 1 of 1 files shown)


brain-geek:
And after merging this we'll also have opportunity to close #729 .

lessons/specifics/guardian.md
```diff
@@ -0,0 +1,264 @@
...
+* Single page applications
+* Controllers (via browser session)
+* Controllers (via authoirzation headers - API)
+* Phoenix Channels
+* Service to Service requests
+* Inter-process
+* 3rd Party access (OAuth)
+* Remember me functionality
+* Other interfaces - raw TCP, UDP etc
+* Command line (if you _really_ want)
```
ybur-yug: This is in `specifics` (or maybe `libraries`, if we go that route) and I feel like this is going just a little too advanced for covering the dead basics. Its already made quite clear that its a portable interface, mentioning the CLI stuff is true but I think it might be unnecessary for the actual purposes of the guide.

lessons/specifics/guardian.md
```diff
@@ -0,0 +1,264 @@
...
+
+JWT tokens can be used everywhere in your application where you need to provide verifiable authentication.
+
+### Do I have to use a database
```
ybur-yug: I also think this might be misleading. In almost every IRL situation I can think of, you use a database with this style of auth flow.

lessons/specifics/guardian.md
doomspork: I agree with @ybur-yug here,  I think we can remove this line.  The previous one, "Other interfaces", covers CLI.

doomspork:
Thanks for taking the initiative on this @brain-geek!  I still need to read through and leave feedback.

The big thing for us to discuss is really just: do we want lessons on libraries and if so, how are we going to pick which to cover?

My $0.02 💰  — I think the value of adding certain libraries to Elixir School is tremendous.  Poolboy, Distillery, Guardian, Porcelain, and some others are either widely used or clear winners in their category, including them seems like an obvious value add.  That said, I _do_ see the downsides:

+ We will likely see a surge in lessons for libraries that won't have staying power or are similar to existing lessons, we don't need 10 lessons on using libraries to send emails.
+ As time goes on certain libraries may fall from favor, how do we handle this?  I think the lessons would still be valuable but a notice might be necessary.
+ We all keep up with Elixir changes but who's going to track and manage the library changes?  We'll need to think about mentioning a version.
+ It's a bunch more translation work

There may be others I'm not thinking of at the moment. 

I think it's worth hearing everyone's thoughts, so what are folks thinking?  cc @pjhampton @ybur-yug @pragmaticivan @nscyclone

lessons/specifics/guardian.md
brain-geek: And in some cases one party issues JWTs and second has key to validate them. But they are not sharing anything except for the secret.

brain-geek:
I've proofread the article once again. 

@ybur-yug @doomspork Can you look through?

@doomspork I guess that should be a topic for some `Issues` discussion.

> We will likely see a surge in lessons for libraries that won't have staying power or are similar to existing lessons, we don't need 10 lessons on using libraries to send emails.

While that's probable, actual lesson creation with passing all reviews is a pretty long way to go through. I'd consider this problem after we face it. 

> As time goes on certain libraries may fall from favor, how do we handle this? I think the lessons would still be valuable but a notice might be necessary.

If some library dies, we can discuss deprecating the lesson in `Issues`. e.g. it's broken in recent Elixir releases and nobody is going to fix it.

> We all keep up with Elixir changes but who's going to track and manage the library changes? We'll need to think about mentioning a version.

That's a nice point! Surely we should. Or get some crazy automated testing solution for this. 😄

doomspork:
All good points @brain-geek 👍   I'm on board with moving ahead with adding libraries.

@hassox would you care to give this final lessons a once over?

brain-geek:
Guys, I would love some final feedback to merge it.

doomspork:
@brain-geek let's merge and update with PRs for further feedback.  How do we want to do this?  Merge this PR, then #729, then update this lesson to be under libraries?

nscyclone:
@doomspork, I think it is easier to merge #729 first. The new section won't appear in the sidebar if it has no lessons for the selected language. Then we could update this lesson and trouble-free :ship: it in its place.

doomspork:
Sounds good @nscyclone 👍

doomspork:
@brain-geek will you update this to be lesson 1 in libraries?

brain-geek:
@doomspork Yeah, give me a minute.

doomspork:
Hah @brain-geek, that was _way_ faster than I expected.  You can have as many minutes as you need 😀

doomspork:
💥   Thank you @brain-geek and @hassox for putting this lesson together.

brain-geek:
@doomspork Just trying to do my best. 😄 Thanks for fast merging.

doomspork:
@brain-geek you're awesome and your help is always greatly appreciated!

PR #791, opened 2016-11-10
Onerous -> Onus

kitlangton:
The wrong word was used. Fix't!

I don't know what's up w/ the diff, I only changed one word.


ybur-yug:
Thanks @kitlangton!

PR #807, opened 2016-11-16
[PL] Replacement of ";" with "."

rafalbig:
My proposition is to replace ";" with "." and start new sentence.


doomspork:
Any thoughts on this @Koziolek?  Thanks again 👍

PR #808, opened 2016-11-18
Improve the explanation of pattern matching

### LINKED ISSUE #797: Improve explanation in Pattern Matching lesson
What really helped me understand how pattern matching in Elixir/Erlang work is to compare the '=' operator to the equals sign in algebra. I would phrase the introductory sentence in the section like so. What do you think?

Are you ready for a curveball? In Elixir, the = operator is actually a match operator, comparable to the equals sign in algebra. If you think about it, writing x = x + 1 is fundamentally wrong in algebra, so Elixir does not let you do it. Instead, writing = makes Elixir match the value of variables on the left-hand side with values on the right-hand side. Let's take a look:

(and so on, maybe smaller changes to the examples but I would put that in the Pull Request)

nscyclone:
Closes #797.

### DIFF
--- a/lessons/basics/pattern-matching.md
+++ b/lessons/basics/pattern-matching.md
@@ -12,7 +12,7 @@ Pattern matching is a powerful part of Elixir.  It allows us to match simple val
 
 ## Match Operator
 
-Are you ready for a curveball?  In Elixir, the `=` operator is actually a match operator.  Through the match operator we can assign and then match values, let's take a look:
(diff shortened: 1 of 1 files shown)


ybur-yug:
@nscyclone looks rad!

ybur-yug:
@nscyclone is there already an issue for translations?

nscyclone:
@ybur-yug, I'll create one in a few moments. Thanks for caring about this.

PR #817, opened 2016-11-22
fix modules link using relative link

techgaun:
The link to modules (struct) is incorrect. Just wanted to fix that


nscyclone:
Nice catch! Thanks, @techgaun!

PR #821, opened 2016-11-25
greek translation behaviours + gen-stage

gemantzu:


ybur-yug:
Thanks @gemantzu! Do you have anyone who would be up to review this for grammar etc, or do you think we should just merge it?

gemantzu:
I did the previous lessons, I don't know if anyone reviewed them.  Tbh, I wish I had a second opinion about some controversial things from someone who read the Greek translations, but I guess we can't have that kind of info.

ybur-yug:
Thats what I had thought but wasn't 100% sure. I will merge this, and tweet out seeing if I happen to know anyone who might be fluent that could do it.

ybur-yug:
Thanks again <3

gemantzu:
Ok, thnx!

gemantzu:
One more thing, could you do something about issues? I wish I was informed about corrections, new material to be translated.  I only got an issue update once and I'm pretty sure there have been more modifications since then.

ybur-yug:
Well we _sort of_ have a system. We aren't that great at using it. The normal flow is:

1. Something happens in English version
2. Issue is created to translate it with the list of CC'd people like you see [here](https://github.com/doomspork/elixir-school/issues/809)
3. We check boxes as people submit PR's and link to the #  

It appears greek isnt on that list. We need to start adding it. I will now. The project boards also exist but I am not sure how up to date they are.

This has become a very large project, and the real heroes are the people doing all of the translation at this point. We need to figure out a better way to give y'all the props you deserve, and we also could work on more effective ways to manage a project with so many simultaneous moving parts.

But, as is often the case with OSS, its a lot of time and a hard problem to solve and we have a bunch of great people trying and no one necessarily knows the best way.

gemantzu:
I understand, this is the first OSS I have contributed in this scale and it took a significant amount of effort to finish. I wish someone can spot some errors and give info on them, because I'm pretty sure there will be inconsistencies from the various changes that have happened after I did the translation.

gemantzu:
did something happen after you merged this pr?

doomspork:
@gemantzu I was away at the time but this is the commit message from @ybur-yug's change:

"These files were saved as latin1 (ISO-8859-7), but Jekyll expects UTF-8. The
previous version of this branch was incorrectly encoded from latin1
(ISO-8859-1) to UTF-8 through vim; this version used iconv to use the correct
version of ISO-8859 for Greek language texts."

gemantzu:
What could be the problem to this? Windows (unfortunately I have to use windows @ work), Emacs (I used atom in previous translations) or am I missing something else?

PR #822, opened 2016-11-25
Reencode Greek Files

halostatue:
These files were saved as latin1 (ISO-8859-7), but Jekyll expects UTF-8. The
previous version of this branch was incorrectly encoded from latin1
(ISO-8859-1) to UTF-8 through vim; this version used iconv to use the correct
version of ISO-8859 for Greek language texts.


ybur-yug:
This might fix the error in production we have. I dont have access, but since its already broken and auto-deploys maybe this will fix it. Can't make it any worse. Heh. We'll have to wait for @doomspork to know for sure but I'm going to merge this.

PR #825, opened 2016-11-27
New Lesson: Protocols

ybur-yug:
# Add Protocols Lesson
This is a first run at a lesson on Protocols for the advanced section from what I started with and finished with @doomspork's feedback.

### TODO
- [ ] Create issue for translation once reviewed/merged

### DIFF
--- a/lessons/advanced/protocols.md
+++ b/lessons/advanced/protocols.md
@@ -1,0 +1,115 @@
+---
+layout: page
+title: Protocols
+category: advanced
+order: 12
+lang: en
+---
+
+In this lesson we are going to look at Protocols, what they are, and how we use them in Elixir.
+
+{% include toc.html %}
+
+## What Are Protocols
+So what are they?
+Protocols are a means of achieving polymorphism in Elixir.
+One pain of Erlang is extending an existing API for newly defined types.
+To avoid this in Elixir the function is dispatched dynamically based on the value's type.
+Elixir comes with a number of protocols built in, for example the `String.Chars` protocol is responsible for the `to_string/1` function we've seen used previously.
+Let's take a closer look at `to_string/1` with a quick example:
+
+```elixir
+iex> to_string(5)
+"5"
+iex> to_string(12.4)
+"12.4"
+iex> to_string("foo")
+"foo"
+```
+
+As you can see we've called the function on multiple types and demonstrated that it work on them all.
+What if we call `to_string/1` on tuples (or any type that hasn't implemented `String.Chars`)?
+Let's see:
+
+```elixir
+to_string({:foo})
+** (Protocol.UndefinedError) protocol String.Chars not implemented for {:foo}
+    (elixir) lib/string/chars.ex:3: String.Chars.impl_for!/1
+    (elixir) lib/string/chars.ex:17: String.Chars.to_string/1
+```
+
+As you can see we get a protocol error as there is no implementation for tuples.
+In the next section we'll implement the `String.Chars` protocol for tuples.
+
+## Implementing a protocol
+
+We saw that `to_string/1` has not yet been implemented for tuples, so let's add it.
+To create an implementation we'll use `defimpl` with our protocol, and provide the `:for` option, and our type.
+Let's take a look at how it might look:
+
+```elixir
+defimpl String.Chars, for: Tuple do
+  def to_string(tuple) do
+    interior = 
+      tuple
+      |> Tuple.to_list
+      |> Enum.map(&Kernel.to_string/1)
+      |> Enum.join(", ")
+
+    "{#{interior}}"
+  end
+end
+```
+
+If we copy this into IEx we should be now be able to call `to_string/1` on a tuple without getting an error:
+
+```elixir
+iex> to_string({3.14, "apple", :pie})
+"{3.14, apple, pie}"
+```
+
+We know how to implement a protocol but how do we define a new one?
+For our example we'll implement `to_atom/1`.
+Let's see how to do that with `defprotocol`:
+
+```elixir
+defprotocol AsAtom do
+  def to_atom(data)
+end
+
+defimpl AsAtom, for: Atom do
+  def to_atom(atom), do: atom
+end
+
+defimpl AsAtom, for: BitString do
+  defdelegate to_atom(string), to: String
+end
+
+defimpl AsAtom, for: List do
+  defdelegate to_atom(list), to: List
+end
+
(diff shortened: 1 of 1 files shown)


lessons/advanced/protocols.md
```diff
@@ -0,0 +1,78 @@
...
+
+```elixir
+to_string({:foo})
+** (Protocol.UndefinedError) protocol String.Chars not implemented for {:foo}
+    (elixir) lib/string/chars.ex:3: String.Chars.impl_for!/1
+    (elixir) lib/string/chars.ex:17: String.Chars.to_string/1
+```
+
+As you can see, we get a protocol error.
+This is because there is no implementation for the Protocol in the tuples.
```
doomspork: As you can see we get a protocol error as there is no implementation for tuples.  In the next section we'll implement the `String.Chars` protocol for tuples.

## Implementing a protocol

We saw that `to_string/1` has not yet been implemented for tuples so let's add it.  To create an implementation we'll use `defimpl` with our protocol, and provide the `:for` option and our type.  Let's take a look at how it might look:

```elixir
defimpl String.Chars, for: Tuple do
  def to_string(tuple) do
    insides = 
      tuple
      |> Tuple.to_list
      |> Enum.map(&Kernel.to_string/1)
      |> Enum.join(", ")

    "{#{insides}}"
  end
end
```

If we copy this into IEx we should be now be able to call `to_string/1` on a tuple without getting an error:

```elixir
iex> to_string({3.14, "apple", :pie})
"{3.14, apple, pie}"
```

lessons/advanced/protocols.md
```diff
@@ -0,0 +1,78 @@
...
+To implement a protocol, `defimpl/3` is our tool of choice.
+
+## Working With Protocols
```
doomspork: How about we use an example that can be applied to a few different types?  Something like:

We know how to implement a protocol but how we do we define a new one?  For our example we'll implement `to_atom/1`, let's see how to do that with `defprotocol`:

```elixir
defprotocol AsAtom do
  def to_atom(data)
end

defimpl AsAtom, for: Atom do
  def to_atom(atom), do: atom
end

defimpl AsAtom, for: BitString do
  defdelegate to_atom(string), to: String
end

defimpl AsAtom, for: List do
  defdelegate to_atom(list), to: List
end
```

Here we've defined our protocol and it's expected method, `to_atom/1`, along with implementing it for a few types.  Now that we have our protocol, let's put it to use in IEx:

```elixir
iex> import AsAtom
AsAtom
iex> to_atom("string")
:string
iex> to_atom(:an_atom)
:an_atom
iex> to_atom([1, 2])
:"\x01\x02"
```

doomspork:
@ybur-yug thoughts on the feedback?

ybur-yug:
@doomspork let me read over this all after work today and give another pass :)

ybur-yug:
Finished a cleanup of a first proposed draft with the whole lesson after initial start + feedback. Feel to comment/review etc, all.

lessons/advanced/protocols.md
```diff
@@ -0,0 +1,107 @@
...
+AsAtom
+iex> to_atom("string")
+:string
+iex> to_atom(:an_atom)
+:an_atom
+iex> to_atom([1, 2])
+:"\x01\x02"
+```
+
+As we can see, protocols are a powerful way to achieve polymorphism, but we also want to be cautious with what features we spread between modules to avoid unintended results.
```
doomspork: This part I'm not sure I get: ", but we also want to be cautious with what features we spread between modules to avoid unintended results."

ybur-yug: If you implement stuff that just catches all and returns something (possible, but stupid) you are going to have a bad time and not get errors that should be being raised etc. I guess maybe its not worth mentioning. Makes sense without it still, too.

lessons/advanced/protocols.md
```diff
@@ -0,0 +1,107 @@
...
+iex> to_string(5)
+"5"
+iex> to_string(12.4)
+"12.4"
+iex> to_string("foo")
+"foo"
+```
+
+As you can see we've called the function on multiple types and demonstrated that it work on them all.
+What if we call `to_string/1` on tuples (or any type that hasn't implemented `String.Char`)?
```
nscyclone: `String.Char` → `String.Chars`

lessons/advanced/protocols.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+If we copy this into IEx we should be now be able to call `to_string/1` on a tuple without getting an error:
+
+```elixir
+iex> to_string({3.14, "apple", :pie})
+"{3.14, apple, pie}"
+```
+
+We know how to implement a protocol but how we do we define a new one?
```
nscyclone: `we do we` → `do we`

lessons/advanced/protocols.md
```diff
@@ -0,0 +1,107 @@
...
+end
+
+defimpl AsAtom, for: BitString do
+  defdelegate to_atom(string), to: String
+end
+
+defimpl AsAtom, for: List do
+  defdelegate to_atom(list), to: List
+end
+```
```
nscyclone: If I am not mistaken `Map`'s protocol implementation will not work for structs. 
Thus probably it's worth to extend this example by adding a protocol implementation for `Map` and shortly explain how protocols work with structs.

ybur-yug: I'll have to look into this one. I'm going to give the lesson a second pass today at the airport/on my plane.

lessons/advanced/protocols.md
doomspork: @ybur-yug that isn't what that sentence conveys though, not to me at least.

lessons/advanced/protocols.md
doomspork: Good idea @nscyclone 👍

lessons/advanced/protocols.md
```diff
@@ -0,0 +1,115 @@
...
+
+If we copy this into IEx we should be now be able to call `to_string/1` on a tuple without getting an error:
+
+```elixir
+iex> to_string({3.14, "apple", :pie})
+"{3.14, apple, pie}"
+```
+
+We know how to implement a protocol but how do we define a new one?
+For our example we'll implement `to_atom/1`
```
doomspork: I think we're missing punctuation here.

ybur-yug:
@doomspork incorporated + pushed

Issue #826, opened 2016-11-27
Ecto Changeset section mixed up error functions


jeramyRR:
The section below has the two functions, password_mismatch_error and password_incorrect_error in incorrect order.

```elixir
defp validate_password_confirmation(changeset) do
    case get_change(changeset, :password_confirmation) do
      nil ->
        password_mismatch_error(changeset)   <---- Should be in the confirmation case
      confirmation ->
        password = get_field(changeset, :password)
        if confirmation == password, do: changeset, else: password_incorrect_error(changeset)  <---- should be in the nil case
    end
  end
```

doomspork:
Thank you @jeramyRR 👍

PR #827, opened 2016-11-27
Fix password error function order in validate_password_confirmation (ref #826)

jeramyRR:
password_incorrect_error and password_mismatch_error are called in opposite order of where they should have been called in the get_change case.  This fix reorders them correctly.

### DIFF
--- a/lessons/specifics/ecto.md
+++ b/lessons/specifics/ecto.md
@@ -297,10 +297,10 @@ defmodule ExampleApp.User do
   defp validate_password_confirmation(changeset) do
     case get_change(changeset, :password_confirmation) do
       nil ->
-        password_mismatch_error(changeset)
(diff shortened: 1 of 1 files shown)


ybur-yug:
Thanks @jeramyRR! Corrections like this are always good and are so minor they can be quite frustrating for someone diving in fresh.

I normally would merge this one quick but since its currently not proper and I recently caused an outage merging something without a properly robust check, I'm being a bit cautious. I'll pull this sometime tomorrow and walk through/verify all of it and merge once I get it working.

Thanks again! <3

ybur-yug:
(obviously if someone else with perms to merge wants to verify first or override my cautious call feel free)

doomspork:
This should be good, thank you @jeramyRR 👍  

Thanks for the caution @ybur-yug 😀

PR #832, opened 2016-11-29
Relocate the Guardian lesson

### LINKED ISSUE #831: Routing/Naming Mismatch In Libraries Section
This may be intentional. If it is and I missed the conversation in the PR for adding the libraries section, apologies.

If we go to look at [the new Guardian Lesson](http://elixirschool.com/lessons/specifics/guardian/) the URL is `/lessons/specifics/guardian` but its listed under the `Libraries` entity which is listed with pieces that all have their own namespaced routes.

I would think that Libraries lessons would fall into `libraries/my_lesson`, just for the sake of good routing. But if its some limitation of jekyll that makes sense, too.

doomspork:
@ybur-yug this fixes #831, I'm going to merge this once I get a chance to test locally.


doomspork:
Works locally.

Issue #834, opened 2016-11-29
Translate the new section name


nscyclone:
The new [section](https://github.com/doomspork/elixir-school/blob/master/_config.yml#L107) for libraries was added in #729. 
## Translation

Translation progress:
- [x] en
- [x] ru /cc @nscyclone
- [x] uk /cc @brain-geek 
- [x]  gr /cc @gemantzu 
- [x]  ko /cc @marocchino 
- [x]  jp /cc @3100
- [x]  it /cc @lucidstack 
- [ ]  no /cc @therobinsj 
- [x]  pl /cc @Koziolek 
- [ ]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [x]  cn /cc @cizixs
- [ ]  es /cc @aleandros
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima

doomspork:
@ybur-yug thing we could use a tweet to get this done?

ybur-yug:
@doomspork worth trying, couldnt hurt. I will shoot one out now.

PR #835, opened 2016-11-29
Update _config.yml

gemantzu:
In response to #834


nscyclone:
Thanks a lot, @gemantzu!

PR #840, opened 2016-11-30
Apply #827 to the existing translations

nscyclone:


doomspork:
Thank you for your diligence @nscyclone, nothing gets by you 😀

PR #859, opened 2016-12-05
Update pipe-operator.md

csytan:
Clarified the part of Elixir that function arity doesn't matter to.


doomspork:
Thanks @csytan 👍

PR #865, opened 2016-12-08
basics.md better phrases for better understanding

gemantzu:


doomspork:
@gemantzu you're great, thanks for all your hard work 👍

PR #876, opened 2016-12-14
Change location of toc

gemantzu:
It is confusing to see the toc including the `Why use Poolboy?` when you have just read it.


doomspork:
@gemantzu good catch 👍

PR #879, opened 2016-12-17
Use relative protocol link for license image

brain-geek:
Right now we're linking http image from the https site ( https://elixirschool.com ).

This PR changes it to use the same protocol for image as ws used to access the page.


doomspork:
Thank you @brain-geek 👍

nscyclone:
After these changes the license shield seems to be broken on [GitHub](https://github.com/doomspork/elixir-school). Are we going to do something with this?

brain-geek:
Thanks for noting, @nscyclone !

I'm lookint into it.

PR #880, opened 2016-12-17
s/to_char_list/to_charlist/

jmitchell:
Now it's best to use `to_charlist/1` since `to_char_list/1` is being
deprecated (see https://github.com/elixir-lang/elixir/issues/4909).


doomspork:
Thank you @jmitchell, this is a great addition 👍

PR #883, opened 2016-12-18
Remove obsolete part of guardian doc

brain-geek:
When tranlating guardian lesson, I've found this part, which (IMO) doesn't belong there. It's not related to guardian and, when read incorrectly, may cause harm.

I think that it would be nice to remove it and simplify the article by doing that.

This PR removes this part from both original and translated (the only one) version.


doomspork:
Thank you @brain-geek 👍

PR #891, opened 2016-12-22
Fix the atoms in the `libraries/poolboy` lesson

nscyclone:


doomspork:
Thank you @nscyclone 👍

PR #895, opened 2016-12-22
Use correct name when mentioning PostgreSQL

brain-geek:
Right now in the Poolboy lesson we have `when using psql` part, which is actually not correct - that's the name of CLI utility, database is named PostgreSQL.


doomspork:
Good idea @brain-geek 👍

Issue #742, opened 2016-10-19
Guide about 'require' , 'use' , 'import'


zmcartor:
The three keywords 'require' , 'use' , 'import' are very close in meaning, but do subtlety different things within Elixir. Especially interested in when to use each, and how to write intelligent 'use' macros.

ybur-yug:
This is [RE this conversation](https://twitter.com/zmcartor/status/788872587363307520) if anyone wants to follow the original bit from twitter.

et:
I'd like to pick this up @ybur-yug and @zmcartor.
Should this be a new lesson or added to an existing section like the [modules](https://elixirschool.com/lessons/basics/modules/) section?

doomspork:
Howdy @et!  That would be great and as far as I know it no one is looking into this one currently.  If you'd like to take a whack at it please do!  

I think we can expand on the modules lesson, what do you think?

et:
Hey @doomspork, I think that seems like the right spot. I'm curious though what would need to be expanded for the [composition section](https://elixirschool.com/lessons/basics/modules/#composition). @zmcartor thoughts?

doomspork:
Not critical but while you're making changes it appears we're missing some backticks in that lesson:

![image](https://cloud.githubusercontent.com/assets/73386/20021241/9fc7fe9c-a29a-11e6-9bf1-85bcaa35a94e.png)

Should be:

"Using `UseImportRequire.UseMe` defines a `use_test/0` function through invocation of the `__using__` macro."

notactuallypagemcconnell:
I think we can close this as its handled in in [composition](https://elixirschool.com/en/lessons/basics/modules/) @doomspork

PR #899, opened 2016-12-29
Use/Require/Import Lesson

ybur-yug:
RE Issue #742

# Draft 1

I took a stab at this at a high level while taking a break from something else I'm about to go back to.
Its pretty simple, but is common enough that I think it is a good one (especially since we already had an issue open for it).

I know putting it elsewhere was discussed, but the format is simple enough I did it in its own lesson and am reading over the module lesson and see if I think it fits in. I dont really strongly about it so I'll default to @doomspork on location if we just want to shove it in modules.

I think maybe going into `alias` would be good for this, but am not sure. Thoughts?

### DIFF
--- a/lessons/basics/use-require-import.md
+++ b/lessons/basics/use-require-import.md
@@ -6,15 +6,15 @@ order: 16
 lang: en
 ---
 
-In Elixir, we commonly use several macros that give us some built-in conveniences.
-These are `import`, `require`, and `use`.
-These keywords all sound very similar, but do quite different things in action.
-In this lesson we will investigate how each of these work.
+In Elixir, we commonly use several macros that give us some built-in conveniences to pull in functionality from other parts of our code.
+The most common of these are `import`, `require`, and `use`.
+These keywords all sound very similar, but do quite different things in action when we utilize them.
+In this lesson we will investigate how each of these works.
 
 {% include toc.html %}
 
 ## Import
-Import brings in frunctions and macros from other modules.
+Import brings in functions and macros from other modules.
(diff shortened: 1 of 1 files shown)


lessons/basics/use-require-import.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+As you can see in line 2, we have `use ExUnit.Case`.
+This imports all the common pieces needed for a test module to behave the way mix needs it to in order to interact with it.
+We could define our own module to include that wraps this so we could call it in a different way to be included.
+
+```elixir
+# test/foo_test.exs
+defmodule TestWrapper do
+	defmacro __using__ do
```
ybur-yug: fix tabs

lessons/basics/use-require-import.md
```diff
@@ -0,0 +1,142 @@
...
+		quote do
+			use ExUnit.Case, async: true
+			@some_attributes "foobar"
+    end
+  end
+end
+
+defmodule FooTest do
+	use TestWrapper
```
ybur-yug: fix tabs

lessons/basics/use-require-import.md
```diff
@@ -0,0 +1,142 @@
...
+defmodule MyModule do
+  defmacro __using__(opts) do
+    quote do
+      import MyModule
+    end
+  end
+end
+```
+
+## Require
```
ybur-yug: Require is quite simple and the docs are short. I think referencing `IEx.pry` is enough but if anyone has something to add I'd be happy to incorporate it.

PR #905, opened 2017-01-03
[KO] a bugfix

namjae:


doomspork:
Thank you @namjae and @riseshia 👍

PR #923, opened 2017-01-16
Remove doubled colons in the `specifics/debugging` lesson

nscyclone:


doomspork:
Good catch @nscyclone 👍

PR #925, opened 2017-01-18
Improve Module Lesson

ybur-yug:
This includes:
- Unify style (previously was mixing `  ` and ` ` for spacing + not
  linebreaking per sentence as many other lessons have been in latest
  PRs, and use backticks on some words that were missed
- Mention `defoverridable` in `use` section
- Add example of overriding `use`'d function
- Mention `IEx.pry` as a common use case for `require`
- Mention lexical scope of `import`

## TODO
- [ ] Create translation issue once reviewed/merged

### DIFF
--- a/lessons/basics/modules.md
+++ b/lessons/basics/modules.md
@@ -151,7 +151,7 @@ end
 
 ### `import`
 
-If we want to import functions and macros rather than aliasing the module we can use `import/`:
+If we want to import functions and macros rather than aliasing the module we can use `import`:
 
 ```elixir
 iex> last([1, 2, 3])
@@ -194,9 +194,26 @@ import List, only: :functions
 import List, only: :macros
 ```
 
+It is also worth noting that `import` is lexical in scope.
+
+```elixir
+iex> defmodule MyModule do
+...>   def func_a do
+...>     import List, only: [duplicate: 2]
+...>     duplicate(:ok, 10)
+...>   end
+...>
+...>   def func_b do
+...>     duplicate(:ok, 10)
+...>   end
+...> end
+** (CompileError) iex: undefined function duplicate/2
+```
+Before we had access to that function from import, but in our second function it blew up.
+
 ### `require`
 
-Although used less frequently `require/2` is nonetheless important.  Requiring a module ensures that it is compiled and loaded.  This is most useful when we need to access a module's macros:
+Although used less frequently `require` is nonetheless important.  Requiring a module ensures that it is compiled and loaded.  This is most useful when we need to access a module's macros:
 
 ```elixir
 defmodule Example do
@@ -206,7 +223,7 @@ defmodule Example do
 end
 ```
 
-If we attempt to call a macro that is not yet loaded Elixir will raise an error.
+If we attempt to call a macro that is not yet loaded Elixir will raise an error.  The most commonly used bit that needs `require` due to use of macros is `require IEx; IEx.pry`.
 
 ### `use`
 
@@ -225,15 +242,20 @@ defmodule UseImportRequire.UseMe do
 end
 ```
 
-and we add this line to UseImportRequire:
+and we add this line to use `UseImportRequire`:
 
 ```elixir
 use UseImportRequire.UseMe
 ```
 
-Using UseImportRequire.UseMe defines a use_test/0 function through invocation of the `__using__/1` macro.
+Using `UseImportRequire.UseMe` defines a `use_test/0` function through invocation of the `__using__/1` macro.
+This is all that use does. However, it is common for the `__using__` macro to in turn call alias, require, or import. This in turn will create aliases or imports in the using module. This allows the module being used to define a policy for how its functions and macros should be referenced. This can be quite flexible in that `__using__/1` may set up references to other modules, especially submodules.  We can now call `use_test/0` simply:
 
-This is all that use does. However, it is common for the `__using__` macro to in turn call alias, require, or import. This in turn will create aliases or imports in the using module. This allows the module being used to define a policy for how its functions and macros should be referenced. This can be quite flexible in that `__using__/1` may set up references to other modules, especially submodules.
+```elixir
+iex> UseImportRequire.use_test
+use_test
+:ok
+```
 
 The Phoenix framework makes use of use and `__using__/1` to cut down on the need for repetitive alias and import calls in user defined modules.
 
@@ -251,6 +273,55 @@ end
 
 The `Ecto.Migration.__using__/1` macro includes an import call so that when you `use Ecto.Migration` you also `import Ecto.Migration`. It also sets up a module property which we will assume controls Ecto’s behavior.
 
-To recap: the use macro simply invokes the `__using__/1` macro of the specified module. To really understand what that does you need to read the `__using__/1` macro.
+We can also override a given function that is given to use with `__using__/1`.  This is done by using `defoverridable`.  Let's implement this:
 
-**Note**: `quote`, `alias`, `use`, `require` are a macro used when we work with [metaprogramming](../../advanced/metaprogramming).
+```elixir
+iex> defmodule Foo do
+...>   defmacro __using__(_) do
+...>     quote do
+...>       def bar do
+...>         2
+...>       end
+...>       defoverridable [bar: 0]
+...>     end
+...>   end
+...> end
+{:module, Foo,
+ <<. . .>>, {:__using__, 1}}
+iex> defmodule Baz do
+...>   use Foo
+...>   def bar do
+...>     super() + 2
+...>   end
+...> end
+{:module, Baz,
+ <<. . .>>, {:bar, 0}}
+iex> Baz.bar
+4
+```
+
+As you can see are able to override the default method `bar/0` from `use`, and even can call `super` to get the original functionality.  It is worth noting that `super` must be called with the same number of arguments as the parent method being overridden.  It also must be placed _after_ the functions being overridden.
+
+### Delegation
+It is possible to define a function that delegates to another module.  Using defdelegate we can define a public function and delegate it to another module's function, let's see it in action:.
+
+```elixir
+iex> defmodule MyModule do
+...>   def hello, do: "hello from my module"
+...> end
+{:module, MyModule,
+ <<70, ... >>, {:hello, 0}}
+iex> defmodule SomeOtherModule do
+...>   defdelegate world, to: MyModule, as: :hello
+...> end
+{:module, SomeOtherModule,
+ <<70, ... >>, [world: 0]}
+iex> IO.puts SomeOtherModule.world
+hello from my module
+```
+
+As you can see our calls to `SomeOtherModule.world/0` are delegated to `MyModule.hello/0`.
+
+To recap: the `use` macro simply invokes the `__using__/1` macro of the specified module.  To really understand what that does you need to read the `__using__/1` macro.
+
+**Note**: `quote`, `alias`, `use`, `require` are macros used when we work with [metaprogramming](../../advanced/metaprogramming).


lessons/basics/modules.md
```diff
@@ -194,9 +202,28 @@ import List, only: :functions
 import List, only: :macros
 ```
 
+It is also worth noting that import is lexical in scope.
```
doomspork: "that `import/2`"

lessons/basics/modules.md
```diff
@@ -194,9 +202,28 @@ import List, only: :functions
...
+
+```elixir
+iex> defmodule MyModule do
+...>   def func_a do
+...>     import List, only: [duplicate: 2]
```
doomspork: I have not seen an `import/2` used within a function in any Elixir code, is this something we want to encourage?

lessons/basics/modules.md
```diff
@@ -225,15 +254,58 @@ defmodule UseImportRequire.UseMe do
 end
 ```
 
-and we add this line to UseImportRequire:
+and we add this line to use UseImportRequire:
```
doomspork: "use `UseImportRequire`"

lessons/basics/modules.md
```diff
@@ -225,15 +254,58 @@ defmodule UseImportRequire.UseMe do
...
-Using UseImportRequire.UseMe defines a use_test/0 function through invocation of the `__using__/1` macro.
+Using `UseImportRequire.UseMe` defines a `use_test/0` function through invocation of the `__using__/1` macro.
+
+```elixir
+iex> UseImportRequire.use_test
+use_test
+:ok
+```
+
+We can also override a given function that is given to use with `__using__/1`.
```
doomspork: I'm not sure about this, for one there are too many "givens" but this seems kind of out-of-place, there's no lead-in or reason.  Maybe we can reword it or maybe it should come at the end of the section on `use`?

How about something like:

"Sometimes it may be necessary to allow certain functions within our macro to be overridden, to this we'll need to use `defoverridable`.

Let's look at how we can leverage this:"

lessons/basics/modules.md
```diff
@@ -225,15 +254,58 @@ defmodule UseImportRequire.UseMe do
...
+
+```elixir
+# lib/use_import_require/use_me.ex
+defmodule UseImportRequire.UseMe do
+  defmacro __using__(_) do
+    quote do
+      def use_test do
+        IO.puts "use_test"
+      end
+      defoverridable [use_test: 0]
```
doomspork: Does this need to come after the definition?  If so, we should mention that.

lessons/basics/modules.md
```diff
@@ -225,15 +254,58 @@ defmodule UseImportRequire.UseMe do
...
+    end
+  end
+end
+```
+
+and now we can override it in IEx if we put it in another module:
```
doomspork: "Now let's override it in another module using IEx:"

lessons/basics/modules.md
ybur-yug: Its straight out of the documentation:

```
iex(1)> h import

                         defmacro import(module, opts)

Imports functions and macros from other modules.

import/2 allows one to easily access functions or macros from others modules
without using the qualified name.
```

That said, if you dont like it I can see why.

lessons/basics/modules.md
ybur-yug: I dont know but it has in the examples I saw from the docs. Will investigate.

lessons/basics/modules.md
ybur-yug: Will work on something better than this now.

lessons/basics/modules.md
ybur-yug: Misunderstood original comments concerns. I'm for backticks around them, I think, but dont believe we need arity (for things like `use` `require` `import` `__using__` etc.)

doomspork: Works for me.

lessons/basics/modules.md
```diff
@@ -249,8 +288,44 @@ defmacro __using__(_) do
 end
 ```
 
-The `Ecto.Migration.__using__/1` macro includes an import call so that when you `use Ecto.Migration` you also `import Ecto.Migration`. It also sets up a module property which we will assume controls Ecto’s behavior.
+The `Ecto.Migration.__using__/1` macro includes an import call so that when you `use Ecto.Migration` you also `import Ecto.Migration`.
+It also sets up a module property which we will assume controls Ecto’s behavior.
+
+We can also override a given function that is given to use with `__using__/1`.
+This is done by using `defoverridable`.
+Lets implement this:
```
nscyclone: Shouldn't `Lets` be `Let's` in this case?

lessons/basics/modules.md
```diff
@@ -249,8 +288,44 @@ defmacro __using__(_) do
...
+...> end
+{:module, Baz,
+ <<. . .>>, {:bar, 0}}
+iex> Baz.bar
+4
+```
+
+As you can see, we are able to override the default method `bar` from `use`, and even can call `super` to get the original functionality.
+It is worth noting that `super` must be called with the same number of arguments as the parent method being overridden.
+It also must be placed _after_ the functions being overridden.
```
nscyclone: I suppose we could also mention that `defoverridable` should not be used for private functions because this is deprecated since [1.4](https://github.com/elixir-lang/elixir/releases/tag/v1.4.0).

lessons/basics/modules.md
```diff
@@ -249,8 +288,44 @@ defmacro __using__(_) do
...
 
-To recap: the use macro simply invokes the `__using__/1` macro of the specified module. To really understand what that does you need to read the `__using__/1` macro.
+### Finishing Up
+To recap: the `use` macro simply invokes the `__using__/1` macro of the specified module.
+To really understand what that does you need to read the `__using__/1` macro.
 
 **Note**: `quote`, `alias`, `use`, `require` are a macro used when we work with [metaprogramming](../../advanced/metaprogramming).
```
nscyclone: `are a macro` → `are macros`

lessons/basics/modules.md
nscyclone: Probably we could also leave here a small section about `defdelegate` since we're adding `defoverridable`.

ybur-yug: I think `defdelegate` doesn't really fit. It prescribes behaviour to another module, but in this case we are trying to take behaviour already happening and extend/nullify/change it. I agree with the note on private functions. Thoughts @doomspork?

ybur-yug:
Will make the grammar changes later today @nscyclone and we can all talk about the `defdelegate` bit too and decide what works best.

lessons/basics/modules.md
doomspork: This is a good question.  I'm not sure `defdelegate` makes sense in _this_ portion of the lesson but if we're expanding the Module content I don't see why we shouldn't find a way to mention `defdelegate`.

lessons/basics/modules.md
```diff
@@ -249,8 +288,44 @@ defmacro __using__(_) do
...
+...>       defoverridable [bar: 0]
+...>     end
+...>   end
+...> end
+{:module, Foo,
+ <<. . .>>, {:__using__, 1}}
+iex> defmodule Baz do
+...>   use Foo
+...>   def bar do
+...>     super + 2
```
nscyclone: I think `super` should be `super()` here.

ybur-yug: Youre right, in 1.4 it will warn w/o it.

ybur-yug: I'm gonna go through and edit RE suggestions and such later tonight or tomorrow at the airport and push everything up.

ybur-yug:
@doomspork @nscyclone I believe I addressed all the points you both brought up, please take a look and see what you think now that things are updated when you've a chance.

lessons/basics/modules.md
ybur-yug: I decided to leave this because it is specific to the behaviour of `import` itself, and worth noting. But now that I think more, its also true that `require` and `import` are also lexical. Maybe I should edit to just make one blanket statement about all of them, and not encourage it but just make it known.

Thoughts?

ybur-yug: CC @nscyclone @brain-geek or anyone else who has an opinion

doomspork: Let's remove this.  I don't like it and I don't want to encourage this nor confuse folks.

lessons/basics/modules.md
```diff
@@ -249,8 +288,66 @@ defmacro __using__(_) do
...
+...>     super() + 2
+...>   end
+...> end
+{:module, Baz,
+ <<. . .>>, {:bar, 0}}
+iex> Baz.bar
+4
+```
+
+As you can see, we are able to override the default method `bar` from `use`, and even can call `super` to get the original functionality.
```
doomspork: "As you can see are able to override the default method `bar/0`"

lessons/basics/modules.md
```diff
@@ -249,8 +288,66 @@ defmacro __using__(_) do
...
+It is worth noting that `super` must be called with the same number of arguments as the parent method being overridden.
+It also must be placed _after_ the functions being overridden.
+
+### Delegation
+It is possible to define a function that delegates to another module as well.
```
doomspork: Let's drop the "as well".

lessons/basics/modules.md
```diff
@@ -249,8 +288,66 @@ defmacro __using__(_) do
...
+If you use `defdelegate` instead of `def`, a given function will be public and invoked from outside the module it is defined in.
```
doomspork: The "invoked from outside the module" portion of this is a little confusing to me.  How about something like:

"Using `defdelegate` we can define a public function and delegate it to another module's function, let's see it in action:"

lessons/basics/modules.md
```diff
@@ -249,8 +288,66 @@ defmacro __using__(_) do
...
+iex> defmodule SomeOtherModule do
+...>   defdelegate world, to: MyModule, as: :hello
+...> end
+{:module, SomeOtherModule,
+ <<70, ... >>, [world: 0]}
+iex> IO.puts SomeOtherModule.world
+hello from my module
+```
+
+As you can see, the code is called from `MyModule` when called through `SomeOtherModule`.
```
doomspork: "As you can see our calls to `SomeOtherModule.world/0` are delegated to `MyModule.hello/0`."

lessons/basics/modules.md
brain-geek: I've seen it used in some cases when otherwise you'll have namespace conflicts.

lessons/basics/modules.md
```diff
@@ -6,13 +6,15 @@ order: 8
 lang: en
 ---
 
-We know from experience it's unruly to have all of our functions in the same file and scope.  In this lesson we're going to cover how to group functions and define a specialized map known as a struct in order to organize our code more efficiently.
+We know from experience it's unruly to have all of our functions in the same file and scope.
```
doomspork: Why are we putting every sentence on a new line?  This creates a huge amount of unnecessary changes on each PR.  If there isn't a good reason for this let's stop.

lessons/basics/modules.md
```diff
@@ -249,8 +288,65 @@ defmacro __using__(_) do
...
-The `Ecto.Migration.__using__/1` macro includes an import call so that when you `use Ecto.Migration` you also `import Ecto.Migration`. It also sets up a module property which we will assume controls Ecto’s behavior.
+The `Ecto.Migration.__using__/1` macro includes an import call so that when you `use Ecto.Migration` you also `import Ecto.Migration`.
+It also sets up a module property which we will assume controls Ecto’s behavior.
+
+We can also override a given function that is given to use with `__using__/1`.
+This is done by using `defoverridable`.
+Let's implement this:
+
+```elixir
+iex> defmodule Foo do
```
doomspork: Traditionally we've not put module code within `iex` because it becomes more difficult to read.

Let's try to keep the lesson styles the same

lessons/basics/modules.md
```diff
@@ -251,6 +273,56 @@ end
...
+...> end
+{:module, SomeOtherModule,
+ <<70, ... >>, [world: 0]}
+iex> IO.puts SomeOtherModule.world
+hello from my module
+```
+
+As you can see our calls to `SomeOtherModule.world/0` are delegated to `MyModule.hello/0`.
+
+### Finishing Up
```
doomspork: Let's remove this section header and get this merged 👍

ybur-yug:
@doomspork last feedback incorporated + pushed

doomspork:
I'm going to close this for now since it needs to be rebased and have feedback incorporated.  @ybur-yug let's sync up in Slack about these updates and some other changes for 1.5 😀

PR #928, opened 2017-01-18
Use singular for types in comparison description

nscyclone:
All the types besides `functions` and `maps` were in singular form. The same approach is used in the [official docs](http://elixir-lang.org/getting-started/basic-operators.html).


doomspork:
@nscyclone you're a beast 😀

Issue #930, opened 2017-01-19
IEX Helpers in wrong folder


Koziolek:
Header of file is:

```
---
layout: page
title: IEx Helpers
category: basics
order: 16
lang: en
---
```

but file is in specifics dir.

ybur-yug:
Thanks @Koziolek thats totally my bad.

PR #931, opened 2017-01-19
Move iex helpers lesson to basic folder

Koziolek:
Solution of #930.


doomspork:
Thank you @Koziolek 👍

PR #935, opened 2017-01-20
[KO] Improved comprehension section

Suckzoo:
cc: @Dalgona @devleoper @riseshia @malkoG 
Terms like `list comprehensions` are awkward when it is translated into Korean (i.e. translated term contains too much `Hanja` and readability critically falls) So, I replaced every terms that damages readability from translated one to original one.
Also, I paraphrased every sentences which looked awkward to me. (i.e. translating `As a rule of thumb` to `일반적인 경험 법칙`)

Feedbacks are always welcomed!


riseshia:
Totally agree. and `컨프리헨션` instead of `conprehension` may be better to read in my case.
how about all?

nscyclone:
Thanks, @Suckzoo!
Could you please check whether the terms that you are going to replace are present in any other files?
For example, `리스트+해석` [is used](https://github.com/doomspork/elixir-school/search?utf8=%E2%9C%93&q=%EB%A6%AC%EC%8A%A4%ED%8A%B8+%ED%95%B4%EC%84%9D) in `basics/control-structures`.

devleoper:
Hello @Suckzoo, Thanks for joining contributing Elixir School in Korean!

Choosing the word during translation is hard. I think the Hanja word "리스트 해석" is (a little) better than leaving as-is or merely writing pronounce of the word, like "리스트 컴프리헨션". At least the reader can guess what does that means. (leaving _list comprehension_ might not improve readability, either).

I thought "리스트 해석" looks common in Erlang or functional languages communities or resources such as [books](http://www.kyobobook.co.kr/product/detailViewKor.laf?barcode=9788991268401) or other web resources. and it can be translated in many way.

Could you please describe more which aspect of the word intimidates the readability?

/cc @Dalgona

Suckzoo:
Thanks for the feedbacks!
First of all, I want to mention that the term `readability damage` I used was mainly focused on giving right meaning as intended!

@nscyclone I'll check it out again when this discussion ends!

@devleoper 
> At least the reader can guess what does that means.

I agree with this point of view. However, translated text should not give wrong concept to readers. 

Translation should reflect its own meaning well. And the translated text should be intuitive to readers. Terms like `type inference` could be a good case of translation. Because it's intuitive, and natural! Everyone who had no experience of type inference may know what the concept is. _추론(infer)_ the _타입(type)_, isn't it intuitive?

But in this case, I don't think the translation reflects the concept at all.
1. I've searched _해석_ on K->E dictionary (I've used naver's one.). Result shows words like `interpret`, `read`, `translation` and so forth. No _comprehension_. In this case, reader cannot guess about the concept at all.
2. I've searched _리스트 해석_ and _리스트 컴프리헨션_(which is a pronunciation of _list comprehension_) on google, using secret tab feature. First one shows a page related with the concept as 5th result, and the other shows a page as 1st result. Also, you may find there are just few pages related with the concept when you search with the first keyword, while the other's not.

In addition to 2, I think this shows that most of programmers are already using the term _리스트 컴프리헨션_ more than _리스트 해석_, though it has a published case.

@riseshia Personally, -1 on that opinion. Because, 

0. I prefer as-is one. 
1. Yesterday, I made [a survey](https://twitter.com/suckzo_o/status/822365846357954564) about it and the majority said leaving the term as-is is much better than writing a pronunciation. (This does not reflect everyone's opinion, of course.)
2. Though there is a [rule](https://www.korean.go.kr/front/page/pageView.do?page_id=P000104&mn_id=97) of converting foreign word to korean, most of Koreans are still confused about it. Words having prefix `sh` may be a good case of confusion. (i.e. Most of people still write the word `shell` as `쉘`, while the right way is `셸`.) I think this CHAOS may lead incoherent translation.

But about this issue, I want to hear everyone which way is better. 

Also, to everyone who concerns this discussion, please give me feedback about these:
1. I also replaced the term `편의 문법` to `Syntactic sugar`. Though `편의 문법` reflects what it means, `syntactic sugar` is generally used terms. Please consider why everyone says `sugar` instead of `convenience`
2. You may see [line 86](https://github.com/doomspork/elixir-school/pull/935/commits/8feae6c742182261507b7760fc6063370e4a4b93#diff-0b976df2a3c617717ba3914894e0cd3bR86) changed a lot. Original one mainly focused on _filtering_ `odds` and `non-divisible by 3` _out_. I guessed this happened because `걸러내다` strongly means `filter` and `out`. So I replaced to new one, mainly focusing on `leaving element that satisfies given predicate`.
3. [Line 98](https://github.com/doomspork/elixir-school/pull/935/commits/8feae6c742182261507b7760fc6063370e4a4b93#diff-0b976df2a3c617717ba3914894e0cd3bR98) was also awkward. Is there any better translation of `a rule of thumb` than the changed one?

Feedbacks are always welcomed!

devleoper:
Thanks for telling your thoughts.

Personally, I'd love to avoid leaving words (except names like trademarks) for reasons described above. The word "comprehend" means "grasp mentally; understand", that made I think 해석(analyze; interpret) is worth enough to be a translated term for it. As the functional programming community in Korean language is still (relatively) small, I think sure you can find and suggest another translated word that makes sense more.

Speaking of the word "syntactic sugar", I think leaving it as-is can make sense, but it can be also explained in Korean (like 편의 문법) if Korean does't have proper translated word for it. it already explained and didn't used "설탕".

I agree and highly welcome the second(걸러내다) and third suggestion. it really boosts the readability :rocket:

Suckzoo:
@devleoper 
Posts like [this](http://pyengine.blogspot.kr/2011/11/list-comprehension.html), or [this](http://juehan.github.io/DiveIntoPython3_Korean_Translation/comprehensions.html) uses raw term `List comprehension`. I guess they used the raw one because there is no proper Korean word that replace `comprehend`. Moreover, the first link mentions "The word `comprehend` is awkward when it is translated"

I understand why you don't prefer leaving the word as-is. But, I think these terms should be treated as proper nouns. Terms like this have been used to describe related concept for a long time. Translating terms like these may lead to communication failure, as translation does not reflect the meaning as-is.

But I partly agree that leaving English word can be a stress to Korean reader. So I suggest writing the term by mentioning pronunciation and the original term, like `리스트 컴프리헨션(List comprehension)`. What do you think about it?

devleoper:
Good to go! One minor things, Could you check something of [this comment](https://github.com/doomspork/elixir-school/pull/935#issuecomment-274049012)?

After that I'll merge this. :smile:

Suckzoo:
@devleoper done.
Found no word to replace except the one on `basics/control-structures` Could you review my latest commit? Thanks in advance.

devleoper:
:rocket::tada:

Note: As this repository only allows squashing commits or rebasing, don't worry about adding more commits. :smile: And I made some little tweaks to the commit message to use present tence (like other PRs) and to be more descriptive of your contributions.

Thanks, and Happy new year!

PR #912, opened 2017-01-10
Fixes name error and typo

BryanJBryce:
#911 Changes Javascript to JSON
Fixes typo.

PR #951, opened 2017-02-06
Fix JWT name error in VI translation

brain-geek:
Applying fix from #912 to the last translation not yet affected by this change.


doomspork:
Thank you @brain-geek 👍

PR #958, opened 2017-02-10
greek translation advanced\protocols

gemantzu:
@kosmas take a peek plz!


nscyclone:
Thanks, @gemantzu and @Kosmas!

PR #959, opened 2017-02-10
Update protocols.md

gemantzu:
typo

### DIFF
--- a/lessons/advanced/protocols.md
+++ b/lessons/advanced/protocols.md
@@ -110,6 +110,6 @@ iex> to_atom(%{foo: "bar"})
 :foo
 ```
 
-It is worth noting that although underneath structs are Maps, but they do not share protocol implementations with Maps. They are not enumerable, they cannot be accessed.
+It is worth it to note, that although underneath structs are Maps, they do not share protocol implementations with Maps. They are not enumerable, they cannot be accessed.
 
 As we can see, protocols are a powerful way to achieve polymorphism.
--- a/lessons/advanced/protocols.md
+++ b/lessons/advanced/protocols.md
@@ -110,6 +110,6 @@ iex> to_atom(%{foo: "bar"})
 :foo
 ```
 
-It is worth nothing that although underneath structs are Maps, but they do not share protocol implementations with Maps. They are not enumerable, they cannot be accessed.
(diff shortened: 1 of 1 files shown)


nscyclone:
I don't think that this is a typo. `It is worth noting` means that there is something the reader should pay attention to.
Maybe [this](http://www.linguee.gr/%CE%B1%CE%B3%CE%B3%CE%BB%CE%B9%CE%BA%CE%AC-%CE%B5%CE%BB%CE%BB%CE%B7%CE%BD%CE%B9%CE%BA%CE%AC/%CE%BC%CE%B5%CF%84%CE%AC%CF%86%CF%81%CE%B1%CF%83%CE%B7/it+is+worth+noting.html) could be helpful.

gemantzu:
Maybe I got it wrong then? Is it intended to be `It is worth nothing`?

nscyclone:
I suppose that `It is worth noting` is the right option in this case.
For me, actually, `It is worth to note` sounds more usual and as a bonus it will not lead to a misread.

gemantzu:
Yes, that's what I got while reading this (actually, I didn't even notice that it said `nothing` at first) and that's why I proposed the change (and that's how I translated it in Greek)

lessons/advanced/protocols.md
```diff
@@ -110,6 +110,6 @@ iex> to_atom(%{foo: "bar"})
 :foo
 ```
 
-It is worth nothing that although underneath structs are Maps, but they do not share protocol implementations with Maps. They are not enumerable, they cannot be accessed.
+It is worth noting that although underneath structs are Maps, but they do not share protocol implementations with Maps. They are not enumerable, they cannot be accessed.
```
doomspork: @gemantzu / @nscyclone how about something like:  "It is worth it to note, that although..."

nscyclone: @doomspork I didn't have any issues with understanding that sentence but as I mentioned above this option could prevent potential misreading that led to creation of this issue.

gemantzu: I didn't have issues as well, I just thought that it was a typo, hence the issue. Both of your suggestions seem better to me than the original.

doomspork: @gemantzu if you want to make this change we can get this merged in 👍

doomspork:
Thank you @gemantzu and @nscyclone!

Issue #944, opened 2017-02-02
Versioning


gemantzu:
I think it is time to add some versioning in these translations, so we can see right away which files need editing. We can add an integer on the `header` of the file (right below language for example) that gets bumped up every time a change is submitted.  This has to work only on the english version.  The translations have to declare which version of the original file they cover.
This way, I think we can also show in the website that a translation is `deprecated` and needs some help to stay updated. 
What do u guys think?

doomspork:
@gemantzu I think this is a great idea.  Any suggestions on how best to tackle this?

CC @nscyclone @brain-geek @devleoper @Koziolek and friends?

Koziolek:
Absolutely yes. Versioning is great, but that deprecated is brilliant idea :)

gemantzu:
We could always add versioning like we do in our software (major_version.minor_version.bug_fix) or a simple integer like 100, it becomes 110 if there are some big changes and 111 in minor tweaks. 200 if someone does a major overhaul.  The status becomes `deprecated` when the major version is different, `needs updates` when the second number is different and nothing when the last number is different. This number gets added below the lang in the header.

doomspork:
So as a plan of attack @gemantzu, what all do we need to do?  I'm guessing something along the lines:

1) Add a `version: x.x.x` to the Jekyll metadata.  
2) Tag the current English lessons as `v1.0.0` and then update all translations to their appropriate version.  Those not current could be something like `v0.1.0`.
3) Add some script to compare the current version with the English version and label it accordingly

gemantzu:
1. Sure
2. Tag the current version as 1.0.0 and deprecate all translations, so that maintainers can bump up the versions depending on if the translation file is updated or not.
3. I don't have the slightest idea about Jeckyl, but can we have a script to do that each time there is a change?

doomspork:
@gemantzu I believe since Jekyll has to be recompiled and generated each time that we can hook into that for the checks.  I'll spend some time this weekend looking into it.

devleoper:
If we version documents like vx.x.x, we can display message depending on the differences of versions. Such as:

- There are small changes since this document was translated.
- There are changes since this document was translated. Some parts might not work.
- There was big changes since this document was translated. Most of the parts might be outdated and not working.

gemantzu:
@devleoper I like your messages, we should go with those.  
Could we also do a report for the translators? That will categorize the lessons by the magnitude of differences? 
example:
## Major Changes
[...]

Issue #956, opened 2017-02-09
Protocol Lesson Translation


ybur-yug:
Original PR Here #825.

This adds a lesson on Protocols. It is not yet merged but final feedback has been incorporated.

## Translation
- [x] en #825
- [ ] bg
- [ ] bn
- [ ] cn /cc @cizixs
- [ ] de
- [ ] es /cc @aleandros / @jonahoffline / @eze1981
- [ ] fr /cc @julienXX
- [x] gr /cc @gemantzu #958 
- [ ] id /cc @rezaprima
- [ ] it /cc @sformisano
- [ ] jp /cc @3100
- [ ] ko /cc @devleoper 
- [ ] my /cc @alkaha
- [ ] no  /cc @Devalo
- [ ] pl /cc @Koziolek
- [x] pt /cc @pragmaticivan #977 
- [x] ru /cc @nscyclone @brain-geek #978
- [ ] sk /cc @michalvalasek
- [ ] tr /cc @hasanozgan 
- [ ] uk /cc @brain-geek 
- [ ] vi /cc @joneslee85

brain-geek:
I'll take Russian translation. @nscyclone , do you agree?

nscyclone:
@brain-geek, I wanted to take it too but I don't mind if you'll take it.

gemantzu:
#958 for greek

pragmaticivan:
https://github.com/pragmaticivan/elixir-school/commit/a8543d622f6824b02097dc8147d95f50863ac170

Issue #922, opened 2017-01-16
Translation of new IEx Helpers lesson


doomspork:
Another new lesson, this one on IEx helpers courtesy of @ybur-yug!  Original PR here: https://github.com/doomspork/elixir-school/pull/824

## Translation

Translation progress:
- [x] en #824
- [ ]  cn /cc @cizixs
- [ ]  es /cc @aleandros / @jonahoffline / @eze1981
- [ ]  fr /cc @julienXX
- [x]  gr /cc @gemantzu #945
- [ ]  id /cc @rezaprima
- [ ]  it /cc @sformisano
- [x]  ja /cc @3100 #1248
- [x]  ko /cc @devleoper #1070
- [ ]  my /cc @alkaha
- [x]  no /cc @Devalo
- [x]  pl /cc @Koziolek #932 
- [x]  pt /cc @pragmaticivan
- [ ]  ru - /cc @nscyclone
- [x]  sk /cc @michalvalasek
- [ ]  vi /cc @joneslee85

3tarazona:
Hi, I would like to help with Spanish translations! Let me know if you need my help in any document.

nscyclone:
Hi, @3tarazona!
AFAIK the Spanish translations are not started yet, so you probably would be interesting in beginning it from scratch.
We have a [small section](https://github.com/doomspork/elixir-school/blob/master/README.md#translating) about translations in our README that might be helpful. There are also some [common rules](https://github.com/doomspork/elixir-school/blob/master/CONTRIBUTING.md) for contributions.
If you will have any questions after reading these, feel free to ask here. We are glad to help.
Thanks in advance!

brain-geek:
FYI @nscyclone I would like to take this one for translation also.

gemantzu:
#945 greek

stfnsr:
I'd like to get started with the german translation of this (as this seems to be the only missing section from the basics)

iguchi1124:
Hi! I'd like to try japanese translation of this doc:smile:

Issue #856, opened 2016-12-05
Translation of GenStage Lesson


Koziolek:
The new [GenStage](https://github.com/doomspork/elixir-school/commit/73f465be84608947588bb024868fed0677304b38) lesson is ready.

PR #770 was the origin.

## Translation

Translation progress:
- [x] en - #710  
- [x]  ru - /cc @brain-geek / @nscyclone #845 
- [ ]  ko /cc @devleoper
- [ ]  jp /cc @3100
- [ ]  no /cc @therobinsj 
- [x]  pl /cc @Koziolek #902 
- [ ]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [ ]  cn /cc @cizixs
- [ ]  es /cc @aleandros / @jonahoffline / @eze1981
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @sformisano
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima
- [x]  gr /cc @gemantzu #821

gemantzu:
#821 for greek, the system had an issue so it might be a later commit than this.

PR #961, opened 2017-02-13
Support versioning

doomspork:
Looking for feedback on the approach here.  I'll need a hand with the HTML styling once we've agreed on the rest.

This is part of #944.

RFC @gemantzu @brain-geek @Koziolek @pragmaticivan @devleoper — feel free to @ others.

image


_plugins/version.rb
```diff
@@ -0,0 +1,52 @@
+# Aggregate the En lesson versions together
+Jekyll::Hooks.register :site, :post_read do |site|
```
doomspork: If we use a `:site` `:pre_render` hook we can create a page/report for outstanding translation work.

_config.yml
```diff
@@ -192,3 +192,9 @@ share:
   pt: Compartilhe essa página
   bn: পৃষ্ঠা শেয়ার করুন
   vi: Chia sẻ trang này
+
+warnings:
+  en:
+    major: "Major differences"
```
doomspork: Thoughts on what these should actually say?

_plugins/version.rb
```diff
@@ -0,0 +1,52 @@
...
+
+# Compare a page's version to the master version and populate a warning if necessary
+Jekyll::Hooks.register :pages, :pre_render do |page, payload|
+  data           = page.data
+  category       = data["category"]
+  default_lang   = page.site.config["default_lang"]
+  lang           = data["lang"]
+  order          = data["order"].to_i
+  master_version = page.site.config["versions"][category][order]
+  page_version   = data["version"] || "0.0.0"
```
doomspork: Once @gemantzu's page updates are on this branch, the default versions here and on line 46 can probably go away

_plugins/version.rb
gemantzu: Ok that's super awesome! I believe we could find ways to gameify this (that's a word) so more people can contribute.

doomspork: I was thinking something similar!  I'd like to put a page together with a call to action to encourage folks to get involved.

_plugins/version.rb
doomspork: His PR is open @ #962

_config.yml
gemantzu: Name them what you want so that everyone can see them from the start. A sensible default is usually a good thing (Rails has corrupted me). If there is a better way to express it, someone will think about it.

_plugins/version.rb
nscyclone: I have a suggestion: we could use a color as the main indicator of the translation relevance. The idea itself is pretty simple:
* Display a colored circle next to the lesson name (red, orange, yellow or green). It will indicate how actual the translation is.
* At the first time, likely, the user will not know what does the color mean. So the whole header should be clickable. After a click an information block should appear:
  * It will briefly describe the difference between the original lesson and the translation.
  * It may also contain a call to potential contributors.

This is quite a spontaneous idea so there are some unanswered questions:
* Displaying the number of changes will require a lot of translation work and a more complex mechanism to compose the result string. We can probably use just a static text for every case like `There are _some_ changes...`.
* Maybe the information block should always be displayed instead of appearing after some user action.
* I am still not sure where should the links head to.
* It could likely be done way better in terms of design.

Click [here](http://imgur.com/a/GvZdB) to see some screenshots of a very raw sketch.

Any thoughts?
/cc @doomspork @brain-geek @gemantzu @Koziolek @pragmaticivan @devleoper and everyone else.

gemantzu: I like the idea of always displaying the message (and your raw sketch was pretty good). People would feel more confident reading the translated file, knowing that it is up to date. I don't know about displaying number of changes, how could we do that?

What I think would be best is that it would always display the message, and if it is not green, it would make the box clickable linking to the repo showing the lates commits on the english translation for the file. The link would be like this: 
https://github.com/doomspork/elixir-school/commits/master/lessons/#{category}/#{lesson}
so that anyone who wants to contribute, can check on the latest changes of the file.

doomspork: @nscyclone I want to incorporate your indicator idea and better messaging and get this merged but I think the change count may be a future enhancement, how's that sound?

nscyclone: @doomspork Sounds good. Then what should we do next? Are we going to merge the versioning updates in the current state (this PR and #962)?

nscyclone: @doomspork Recently I've pushed some updates. Looking forward to your feedback.

doomspork: @nscyclone you're awesome, thank you for the help!  I'll pull this down today and take a peek! 👍

doomspork:
@nscyclone with your changes I think this is good to roll out, thanks again!

@gemantzu / @nscyclone — thoughts on how we should roll this out?  Are there translated lessons we _know_ are up-to-date that we should bump to `1.0.0` ahead of time?

nscyclone:
@doomsporkCool, I am always glad to help! 
The [BN] translations (#970) seem to be up-to-date.
Also the newest lessons (#850, #855, #856, #922, #956) weren't changed yet so their translations likely are in the desired state.
In my opinion, the translator's guide should be updated.

gemantzu:
I believe that all translators should decide and bump all updated files, it will be a good chance to review if and where changes are needed.

_plugins/version.rb
gemantzu: I just found some time and tested the whole functionality, it seems pretty good.

nscyclone: @doomspork Is there anything else I could help you with to ship this? :rocket:

doomspork:
@gemantzu I tend to agree but there are some translations we know are up-to-date and I'm not sure we want to create unnecessary work for translators.  Thoughts?

nscyclone:
@doomspork I've just rebased the branch on top of the latest master and then bumped the versions to `1.0.0` as we've discussed [above](https://github.com/doomspork/elixir-school/pull/961#issuecomment-284303363). Also several lessons didn't have any version and were neither included in that comment so I gave them a default version of `0.9.0`.

gemantzu:
I guess if we already know they are up to date we could bump the versions. I will do a check on Greek translations to see if there is something missing and will bump them as well.

doomspork:
@nscyclone you're awesome man, I really appreciate your help!  Once @gemantzu has gone over the Greek lessons I think we'll be good to merge.

doomspork:
Alrighty, time to merge!  Thanks again @nscyclone & @gemantzu!!

nscyclone:
Thanks a lot, @doomspork and @gemantzu! Seeing this live made my morning happier!
There were not that many [tweets](https://twitter.com/elixirschool) recently. Maybe we could post about this to share the joy and ask for assistance with translations, wdyt?

gemantzu:
Tweet away @nscyclone !!! I'm afraid I do not have that many friends but I will retweet this to gain some movement!

doomspork:
@nscyclone you're right and I'm sorry about that, Twitter has gotten behind as I've become more and more busy.  I'm working on a bot that'll label PRs, tag translators, and tweet contributions.  I'll try to make some more progress this weekend and push it up for your consideration!

In the mean time I'll shoot out a tweet about this!

doomspork:
@nscyclone / @gemantzu — https://twitter.com/elixirschool/status/844583554415804417

PR #962, opened 2017-02-13
added version to all files

gemantzu:


doomspork:
This is related to #944 and #961.  This should be safe to merge before #961, Jekyll should just ignore the version.

doomspork:
Thank you again @gemantzu.  I am going to merge this and then starting updating the versions in #961

PR #967, opened 2017-02-17
fix Plug Router.start/2 is undefined or private

xinz:
The tracking thread as below:
https://github.com/doomspork/elixir-school/issues/965


ybur-yug:
@xinz working on a few other things, but I'd like to pull this and go through the lesson and verify its all copy-pasteable now that you've exposed this initial bit just to confirm. Thank you a ton for the PR :heart:

doomspork:
Thank you @xinz!

PR #973, opened 2017-02-22
[NO]: Correct to use default code snippets

robinsjdotcom:
Previous snippets were also translated. This is not necessary as the semantics and intuition of the code snippets is importanter.


doomspork:
Thank you again @robinsjdotcom!  This is very helpful, I much appreciate it.

@Devalo could you take a peek at these as well?

PR #980, opened 2017-02-26
Updated [GR]/lessons/Basics/Enum.md file

dodosiz:
I added some more information about the functions Enum.min/2 and Enum.max/2


doomspork:
Thank you @dodosiz!  @gemantzu care to take a peek at this?

gemantzu:
Thnx @dodosiz ! TBH I have mixed feelings over this, but it's not @dodosiz 's fault. I checked on the original and I find it very confsing as a non-native english speaker. I believe we should change the original to be more clear, and then let @dodosiz reiterate over this in the Greek version. @Kosmas could you comment on this as well?

Kosmas:
@dodosiz @gemantzu I think it would be better to replace the 
αυθαίρετη
with
προεπιλεγμένη
everything else looks fine.

gemantzu:
@kosmas sure about the replacement, but what I found weird (it may be just me though) was the `μέσω μιας ανώνυμης συνάρτησης`, which I think is not accurate to what actually happens in the function. And I believe that the culprit is the original English version. What I suggest is that we change the original in something like this: 

`min/2` does the same, but allows us to specify the original / starting / initial value to the anonymous function that we pass:

and translate accordingly. This way we make the lesson more clear to someone that is just starting out.
Sorry if I've dragged this more than I should, I just stated something that bugged me out.

Kosmas:
@gemantzu I think I see your point.
Maybe the description in max/2 is a little bit better and can be used in min/2 (in the English text)

gemantzu:
Tbh I don't like it as well. They way they are both written indicates that Enum is an entity itself and the min / max functions do something to that entity, while in FP functions are the entity, modules are just an enclosure, a package, nothing more. At least that is my point of view from things in these last 8 months or so of doing FP.

gemantzu:
I proposed a PR ( #984  ) that will clear this up. With a small change, @dodosiz 's translation will be perfect and in place (let's see if the PR is accepted first).

doomspork:
@gemantzu your PR is merged, thanks for the suggestion!  After @dodosiz incorporates your changes we need to make sure this branch is rebased, looks like there's some other changes in this PR now.

gemantzu:
Thnx @doomspork . @dodosiz go ahead and retranslate this (and don't forget to rebase as @doomspork proposed).

gemantzu:
I will do a recheck on the whole lesson to see what has changed since then and do a new PR when I find some time, unless ofc @dodosiz comes back and rebases his PR.

doomspork:
Thank you @gemantzu!  Hope all is well 😁

gemantzu:
so @dodosiz doesn't seem responsive, so I went ahead and updated the lesson (Enum.md). I don't know about the other changes though, should they be included?

doomspork:
@gemantzu your call.  We can close this PR if you want and when they're available they can re-open it.

gemantzu:
I guess you should close it then and if it is needed, it will come up again.

PR #981, opened 2017-02-27
Fix small vocab mistake

WDoan:
I believe that the author meant to use 'parentheses' instead of 'brackets' in the pipe operator lesson. Just a small change to avoid confusion.


doomspork:
Thank you @WDoan! 😀

PR #982, opened 2017-02-27
fix typo: exmaple -> example

aseigo:
Tiny typo fix I came across while reading through this section.


nscyclone:
Good catch! Thanks a lot, @aseigo!

PR #984, opened 2017-03-01
Update enum.md

gemantzu:
What do u think about these changes? I feel that this makes it more clear and accurate either for reading, or for translations.

### DIFF
--- a/lessons/basics/enum.md
+++ b/lessons/basics/enum.md
@@ -120,7 +120,7 @@ iex> Enum.min([5, 3, 0, -1])
 -1
 ```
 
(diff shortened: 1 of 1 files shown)


ybur-yug:
I think this makes it more clear and phrased better, but it would cause a lot of work for translators if there its stated similarly in the other langs and needs updated just for a couple sentences. You are more familiar with that side than myself, though. Thoughts?

doomspork:
Thank you @gemantzu 👍

gemantzu:
As a non native English speaker, it got me a bit confused. That's why I proposed these changes. I hope the other translators will benefit, not be bothered from this.

PR #988, opened 2017-03-03
Translate the libraries section label to pt

pedrosnk:


doomspork:
Thank you @pedrosnk 👍

PR #989, opened 2017-03-04
Character list sigil output: missing backslash

fdbeirao:
I noticed while executing the example commands that the output from my iex (1.4.2) has a backslash that is not present in the documentation.
[Screenshot](http://i.imgur.com/YLGv7GG.png)
Cheers.

### DIFF
--- a/lessons/basics/sigils.md
+++ b/lessons/basics/sigils.md
@@ -46,7 +46,7 @@ iex> ~c/2 + 7 = #{2 + 7}/
 '2 + 7 = 9'
 
 iex> ~C/2 + 7 = #{2 + 7}/
-'2 + 7 = #{2 + 7}'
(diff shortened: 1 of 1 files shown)


doomspork:
@fdbeirao you're awesome, thank you so much for all of these PRs!  Like the others, could you apply this change to all translations?

fdbeirao:
@doomspork thank you. I just made these PRs as I was reading along the docs. I will change it in all translations. Do you want separate PRs per issue, or should I fix the three issues in all translations in one single PR? 👍

doomspork:
@fdbeirao you can leave them as a separate PRs or handle them all together, whichever you prefer.

fdbeirao:
@doomspork I created #995 that has all these three PRs for all the languages.
Kudos 👍

PR #990, opened 2017-03-04
Incorrect output on regex sigil

fdbeirao:
Using elixir 1.4.2, Erlang/OTP 19 [erts-8.2] on Windows 10


doomspork:
Thank you @fdbeirao!  Would you mind applying this change to the translations as well?

PR #991, opened 2017-03-04
Use self() instead of self

fdbeirao:
Running the code in this example I got the following warning:
```
warning: variable "self" does not exist and is being expanded to "self()", please use parentheses to remove the ambiguity or change the variable name
  test/example_test.exs:20
```


doomspork:
Thank you @fdbeirao!  Would you mind applying this change to all the translations?

PR #999, opened 2017-03-06
[BN] Translated the lessons on String

code-shoily:
@masnun and @prio101 did the String chapter. I think we'll need to get back to it later and enrich it more, feels incomplete when re-read.

### DIFF
--- a/bn/lessons/basics/strings.md
+++ b/bn/lessons/basics/strings.md
@@ -1,31 +1,29 @@
 ---
 layout: page
-title: Strings (TODO)
+title: স্ট্রিং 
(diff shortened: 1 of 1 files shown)


prio101:
Looks good to me

code-shoily:
Wow, in less than a minute? Dude you read fast :O @prio101

prio101:
@code-shoily via  yes!! ... side effect of reading EULA...

ybur-yug:
Thanks @code-shoily @prio101! You folks are cranking stuff out :) <3

PR #1014, opened 2017-03-14
Use a consistent file naming pattern

nscyclone:
Filenames consisting of multiple words use `-` as a delimiter (e.g. `control-structures`, `umbrella-projects`). This lesson was the only exception.


doomspork:
Thank you @nscyclone 👍

PR #1027, opened 2017-03-25
German translation

Lechindianer:


doomspork:
Thank you @Lechindianer!

PR #1031, opened 2017-03-27
Update 'mnesia/create_schema' example to remove warning

hanmd82:
This simple update to the example code will remove the following warning:

Instead of
```
iex> Mnesia.create_schema([node])
warning: variable "node" does not exist and is being expanded to "node()", please use parentheses to remove the ambiguity or change the variable name
:ok
```

this works better:
```
iex> Mnesia.create_schema([node()])
:ok
```


doomspork:
Thank you @hanmd82!  Could you see whether this change needs to be applied to any of the translated lessons?

hanmd82:
Thanks @doomspork. I have applied the patch to all the translated lessons which would benefit from this

doomspork:
Thank you again @hanmd82 👍

PR #1037, opened 2017-03-29
Fix with syntax in control structures lesson

toothfairy:
The first match clause must be on the same line with `with` keyword. [Docs](https://hexdocs.pm/elixir/master/Kernel.SpecialForms.html#with/1)
Otherwise compiler throws `SyntaxError: unexpectedly reached end of line. The current expression is invalid or incomplete`


doomspork:
This is awesome, thank you for taking care of all the translations @toothfairy!!

< insert joke about waiting to meet the toothfairy since I was kid >

toothfairy:
😮 🔨

PR #1046, opened 2017-04-08
Fix typo on selection language menu

meox:
Italiano is more correct than: L'Italiano


doomspork:
Thank you and welcome @meox!

PR #1048, opened 2017-04-09
[JP] Translate GenStage and Protocols

rinosamakanata:
Hi, 
I translated GenStage and Protocols lessons into Japanese.
Please check it. Thank you :)


doomspork:
Hi @rinosamakanata thank you very much for your contribution and welcome!  @3100 do you have time to look this over?

3100:
Hi, @rinosamakanata. Nice work! 
I suppose 'sink' can be translated to '受け手', but what do you think?

rinosamakanata:
Hi, @3100. Thank you for review and comment!
Surely I agree. The word 'シンク' could be difficult to understand in this context.
So I translated 'sink' into '受け手' as you suggested, and correspondingly I changed 'ソース' to '送り元'. How does that look?

3100:
@rinosamakanata Thank you very much!

Issue #295, opened 2016-04-19
[cn]Translation error about tuple


Daqu:
It is in elixir-school/cn/lessons/advanced/metaprogramming.md **line 22**

I found the word `tuple` was translated in `列表` instead of  `元组`.In chinese, tuple often was translate into `元组` and list was translate into `列表`

the orginal

> ## Quote
> 
> 22  学习元编程的第一步是理解表达式是怎么表示的。Elixir 的 AST（抽象语法树）是列表构成的。这些列表包含了三个部分：函数名称，metadata，还有函数的参数。

my version

> ## Quote
> 
> 22  学习元编程的第一步是理解表达式是怎么表示的。Elixir 的 AST（抽象语法树）是元组构成的。这些元组包含了三个部分：函数名称，metadata，还有函数的参数。

doomspork:
Hi @Daqu, thank you for reaching out!  I will have to defer to @cizixs and others on this.  If we don't hear back, feel free to open a PR with the change 👍

AquarHEAD:
I think it's a correct fix

2016年4月21日木曜日、Sean Callan<notifications@github.com>さんは書きました:

> Hi @Daqu <https://github.com/Daqu>, thank you for reaching out! I will
> have to defer to @cizixs <https://github.com/cizixs> and @AquarHEAD
> <https://github.com/AquarHEAD> on this. If we don't hear back, feel free
> to open a PR with the change 👍
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/doomspork/elixir-school/issues/295#issuecomment-212984312>
>


-- 
楼洵 (AquarHEAD)
----------
Blossoms which were the joys that fell,
And leaves, the hopes that yet remain.
----------
http://about.aquarhead.me

cizixs:
@Daqu You're right, please open a PR if you want to correct this.

doomspork:
@Daqu did you still want to open a PR for this? 😀

cizixs:
Fixed by PR #1049 , I will close this issue.

PR #1049, opened 2017-04-11
[CN] This commit fixes issue #295

cizixs:
Fix the Chinese translation error referred in issue #295


doomspork:
Thank you @cizixs!! 😁

PR #1052, opened 2017-04-16
[KO] Fix a dead anchor link in the `specifics/ecto` lesson translation

nscyclone:


nscyclone:
Thanks for the review, @devleoper!

PR #1058, opened 2017-04-21
Remove no need valiable.

tarunama:
Fix sample code.

### DIFF
--- a/lessons/basics/functions.md
+++ b/lessons/basics/functions.md
@@ -43,12 +43,11 @@ Elixir uses pattern matching to identify the first set of parameters which match
 
 ```elixir
 iex> handle_result = fn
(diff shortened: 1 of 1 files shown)


lessons/basics/functions.md
```diff
@@ -43,12 +43,11 @@ Elixir uses pattern matching to identify the first set of parameters which match
 
 ```elixir
 iex> handle_result = fn
-...>   {:ok, result} -> IO.puts "Handling result..."
```
doomspork: I think it would be better if we made use of the result instead of removing it.  What do you think?

tarunama: I mistake that some_result valiable is need. So, close this Pull Request.

PR #1063, opened 2017-04-25
Update Ecto Lesson for Elixir 1.4

BryanJBryce:
Changes default Application file from `<project name.ex>` to `application.ex`
Removes adding dependencies to application list
Adds requirement for `ecto.gen.repo` to have `-r` flag
Adds requirement for `config.exs` to have `ecto_repos` configuration
Adds instruction to run `mix ecto.create`
Indicates location of migration files
Adds `()` on `timestamps` function
Replaces Model with Schema

### DIFF
--- a/lessons/specifics/ecto.md
+++ b/lessons/specifics/ecto.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.0
+version: 1.1.0
 layout: page
 title: Ecto
 category: specifics
--- a/lessons/specifics/ecto.md
+++ b/lessons/specifics/ecto.md
@@ -7,7 +7,7 @@ order: 2
 lang: en
 ---
 
-Ecto is an official Elixir project providing a database wrapper and integrated query language.  With Ecto we're able to create migrations, define models, insert and update records, and query them.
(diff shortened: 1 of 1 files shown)


BryanJBryce:
So I just went over the hexdocs getting started tutorial for Ecto (https://hexdocs.pm/ecto/getting-started.html) and it is really what this lesson should be. Maybe we just direct people to that and do some deeper dives on other parts of Ecto that aren't covered in Hexdocs?

doomspork:
@BryanJBryce we can definitely add a link but I believe having different resources available is better.  If there's things missing we can add them but Elixir School has never (and likely will never) offered comprehensive guides or step-by-step tutorial for anything, the intention is to serve as an introductory to the topics.  That said, we can certainly look at how to expand on the Ecto lesson in meaningful ways.

doomspork:
Howdy @BryanJBryce!  Any updates here or should we close this?

BryanJBryce:
Sorry, updated that metadata, anything else?

doomspork:
My apologizes @BryanJBryce, I must have inadvertently marked this notification as read and lost track of it 😞  If you could rebase I'd be happy to merge this in, sorry for the delay.

brain-geek:
@BryanJBryce Also, there are some conflicts already. Rebasing would be a very good thing to do here.

PR #1071, opened 2017-05-07
Fix typo on IEx Helpers

riseshia:
Could you check this @doomspork ?

and if there is no problem, we need to update version of documents.
how do you think? this is really small typo problem, `1.0.1` will be enough I think...


doomspork:
@riseshia I think `1.0.1` would be suitable 👍

riseshia:
Done!

PR #1076, opened 2017-05-14
Remove use of single vars in pattern matching

eksperimental:
Since it's a guide, it's better to be explicit on what is each variable


doomspork:
Thank you @eksperimental, this is a great addition 👍

PR #1078, opened 2017-05-14
Small improvements in Basics section

eksperimental:


doomspork:
@eksperimental would you care to apply this change to the translated lessons or bump the version here?

eksperimental:
I have bumped the minor version in the English version only.
And updated the versions in some languages
Additionally I have included a missing sentence in the Spanish version.
My only question is how should we proceed with the versions in the other languages

doomspork:
@eksperimental thanks for all of your contributions.  We're still ironing out the whole versioning scheme for these lessons, I will go back and update lessons where it's necessary 😁

eksperimental:
sweet as! thank you

PR #1080, opened 2017-05-14
Correct Elixir documentation links

eksperimental:


doomspork:
Thank you @eksperimental, this is good idea 👍

PR #1087, opened 2017-05-20
Fix markdown syntax

riseshia:
Do we need bump up? @doomspork

### DIFF
--- a/lessons/basics/control-structures.md
+++ b/lessons/basics/control-structures.md
@@ -129,7 +129,7 @@ iex> cond do
 
 The special form `with/1` is useful when you might use a nested `case/2` statement or situations that cannot cleanly be piped together. The `with/1` expression is composed of the keywords, the generators, and finally an expression.
 
-We'll discuss generators more in the (list comprehensions lesson)[../comprehensions/], but for now we only need to know they use [pattern matching](../pattern-matching/) to compare the right side of the `<-` to the left.
+We'll discuss generators more in the [list comprehensions lesson](../comprehensions/), but for now we only need to know they use [pattern matching](../pattern-matching/) to compare the right side of the `<-` to the left.
 
 We'll start with a simple example of `with/1` and then look at something more:
 
--- a/pt/lessons/basics/basics.md
+++ b/pt/lessons/basics/basics.md
@@ -150,7 +150,7 @@ iex> "foo\nbar"
 "foo\nbar"
 ```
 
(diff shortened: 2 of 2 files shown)


doomspork:
@riseshia does this need to be fixed in other lessons too?

riseshia:
I can't find this in the translations, It may be not translated in all ;_;

eksperimental:
This was my fault, It has recently been introduced. I have run a regex and I found only another file with the same issue.

```
$ ag "\([^\)\]]+\)\[[^\]]+\]"
pt/lessons/basics/basics.md
153:Elixir também inclui tipos de dados mais complexos. Nós vamos aprender mais sobre estes quando aprendermos sobre [coleções](../collections/) e (funções)[../functions/].

lessons/basics/control-structures.md
132:We'll discuss generators more in the (list comprehensions lesson)[../comprehensions/], but for now we only need to know they use [pattern matching](../pattern-matching/) to compare the right side of the `<-` to the left.
```

riseshia:
OTL Thanks @eksperimental , I don't think about finding same errors, but just check this one only `[../comprehensions/]`. I just applied that!

eksperimental:
I'm thinking about implementing a tool that will lint and check the validity of the markdown files, So we avoid future problems. We can make it work with Travis, or other CI tool

riseshia:
like [this one](https://github.com/mivok/markdownlint)?

eksperimental:
you are reading my mind!


```
$ mdl  lessons/basics/control-structures.md
lessons/basics/control-structures.md:13: MD002 First header should be a top level header
lessons/basics/control-structures.md:75: MD009 Trailing spaces
lessons/basics/control-structures.md:159: MD009 Trailing spaces
lessons/basics/control-structures.md:160: MD009 Trailing spaces
lessons/basics/control-structures.md:132: MD011 Reversed link syntax
lessons/basics/control-structures.md:179: MD012 Multiple consecutive blank lines
lessons/basics/control-structures.md:86: MD036 Emphasis used instead of a header
lessons/basics/control-structures.md:104: MD036 Emphasis used instead of a header
lessons/basics/control-structures.md:1: MD041 First line in file should be a top level header
```

look at the offending line:
> lessons/basics/control-structures.md:132: MD011 Reversed link syntax

riseshia:
LGTM XD how do you think @doomspork ?

doomspork:
Thank you @riseshia and @eksperimental! 👍

PR #1098, opened 2017-05-27
[KO] Fix typo

riseshia:
@devleoper @doomspork

Please review and merge this. This is just typo problem. I think we don't need to dedicated review with this ;_;


riseshia:
Many thanks @devleoper T-T

PR #1103, opened 2017-06-03
[KO] Remove verbose '만약' & fix '배열' to '리스트'

riseshia:
## Why?

- we have any problem without `만약` in sentences.
- There is no `배열` in Elixir as we know. so replace this to `리스트`, and actually origin wording is `list`, so there is no problem.

@Dalgona, @malkoG, @devleoper, @marocchino 

Please review this XD


marocchino:
Lgtm

PR #1111, opened 2017-06-13
Use the Elixir version placeholder where it was missing

nscyclone:
There were some places where the Elixir version in `iex` examples was hardcoded.
Also [1.4.4](https://github.com/elixir-lang/elixir/releases) is the latest version now.


doomspork:
Thank you @nscyclone!  You're awesome, hope all is well 😁

PR #1116, opened 2017-06-15
Replace elixir_school with elixir-school

eksperimental:


doomspork:
Thank you @eksperimental!

PR #1135, opened 2017-07-18
Fix the broken redirects

### LINKED ISSUE #1134: 404 not found from google links
This has happened a few times now and it's been getting annoying. Google is returning links like:

https://elixirschool.com/lessons/specifics/plug/

Which renders a text string saying "Not found" and that's it.

It appears that the correct link is (depending on your locale):

https://elixirschool.com/en/lessons/specifics/plug/

I think that the previous link should return do some auto-detection, ask you to choose the locale, or default to a locale.

nscyclone:
Closes #1134.


doomspork:
@nscyclone I will try deploying this first to see if it fixes so we can roll back if not.  I had tried updating Jekyll but I didn't change redirect paths 🤞

doomspork:
@nscyclone you're my friggin hero ❤️

PR #1142, opened 2017-07-29
Use "language" instead of "country"

caio:
The example snippets reference the optional parameter "country" but
what it means seems to be `language` (as [many countries are
multilingual][1]).

[1]: https://en.wikipedia.org/wiki/List_of_multilingual_countries_and_regions

---

A bit of a pedantic change - I know -, but having dealt with l10n some years ago this mix-up was my most frequent headache. In real code I would probably use `language_code` instead, but I don't think we need to be _this_ precise (I'm happy to change it nonetheless, it's a `$ find -exec sed` away).

### DIFF
--- a/bg/lessons/basics/functions.md
+++ b/bg/lessons/basics/functions.md
@@ -161,8 +161,8 @@ iex> Greeter.hello ["Sean", "Steve"]
 
 ```elixir
 defmodule Greeter do
-  def hello(name, country \\ "en") do
-    phrase(country) <> name
+  def hello(name, language_code \\ "en") do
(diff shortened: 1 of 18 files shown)


doomspork:
Howdy @caio and welcome!  I think it's a reasonable change, thanks for suggesting it 😀 

How about we make it `language_code` to be 100% accurate?  If you'd like to make that change we can get this merged!

caio:
Done :)

Thanks a lot for all your hard work on this, really helpful!

doomspork:
You're welcome @caio, thanks for joining the fun 😁

PR #1149, opened 2017-08-11
Update mistaken arity from other_function/1 to other_function/0 to match code sample

bbuck:
I noticed while reading through the pipe lesson that this code example is provided:

```elixir
foo(bar(baz(new_function(other_function()))))
```

And `other_function` is referenced as `other_function/1` but in the example it takes no arguments, so it should be `other_function/0`, I've corrected the issue in all language files that I was able to see it in.


doomspork:
Thank you and welcome @bbuck, this is a great change 👍

bbuck:
Any time @doomspork! I really appreciate this resource you've created and maintain here! It's a lot easier to digest these lessons than the docs on the official Elixir site, it's been great diving back in!

PR #1153, opened 2017-08-26
Fixing the incorrect result of 0x1F is 31

romenigld:
The correct result for 0x1F on iex is 31


doomspork:
Thank you and welcome @romenigld!  Would you mind checking to see if this issue applies to the other translations?  If it does, please update it there too 😁

romenigld:
Hello @doomspork I was doing another things and when I return on this tutorial and see something I will do another pull request.

PR #1155, opened 2017-08-27
Separate site files from project files

burden:
After noticing that the docroot was polluted with project files, I determined that the exclude setting needed to be updated.

ex:
https://elixirschool.com/app.json
https://elixirschool.com/GLOSSARY.md
https://elixirschool.com/LICENSE
https://elixirschool.com/CONTRIBUTING.md
https://elixirschool.com/README.md

I figured it’d be a better idea to simply move the site files into their own directory, so that in the future when new files are added to the project, it is no longer a requirement to update the exclude setting in `_config.yml`


doomspork:
@burden could you rebase your PR please?

doomspork:
We'll evaluate this in another PR for the redesign.

PR #1166, opened 2017-09-14
Benchee guide 1.0.0

gemantzu:


doomspork:
Wow that was super fast @gemantzu ❤️

PR #1167, opened 2017-09-19
Update reference to Enum.chunk to modern Enum.chunk_every

hlship:
As stated, the live version is for `Enum.chunk` which doesn't exist in Elixir 1.5.1.

Alas, I'm not in a position to handle all the non-English translations.


doomspork:
Thank you @hlship!  I'll take care of the other languages 👍

PR #1173, opened 2017-09-24
Remove unused SCSS

### LINKED ISSUE #996: Are assets from the `_sass` folder actually being used anywhere?
I couldn't find any usage of the files from the mentioned folder but I might be missing something.
I removed it in a separate branch and everything seems to be working as usual both on Heroku and locally.
So the question is: can these assets be safely removed? If not, I'd like to know the secret where and what are they used for.
The branch can be found [here](https://github.com/nscyclone/elixir-school/tree/fix/remove-unused-scss). Ready to propose a PR if needed.
/cc @doomspork

nscyclone:
After the [latest assets rearrangement](https://github.com/elixirschool/elixirschool/pull/1162) we can be sure that these files aren't needed anymore. Closes #996.


doomspork:
Awesome work as always @nscyclone 👍

PR #1175, opened 2017-09-26
Remove permanent scrollbar from the sidebar

nscyclone:
The scrollbar in the sidebar is present even if the content doesn't overflow its box. [Example.](https://elixirschool.com/uk/)


doomspork:
Thank you @nscyclone 👍

PR #1177, opened 2017-09-29
Fix chapter nesting in the `basics/testing` lesson

nscyclone:
`capture_io` and `capture_log` should be parts of `ExUnit` but currently the TOC looks like [this](https://elixirschool.com/en/lessons/basics/testing/):
exunit


doomspork:
You're the man @nscyclone, thank you for all of your help!

PR #1178, opened 2017-09-29
[CN]new translation 

Daqu:
someone who use windows may start elixir console fail because they type `$ iex` only


Daqu:
add a translation abou Sigils because none do that~~~~~

cn/lessons/basics/basics.md
```diff
@@ -26,6 +26,8 @@ Elixir 自带了 `iex` 这样一个交互 shell, 可以让我们随时计算 Eli
 
 运行 `iex` 命令，让我们开始教程：
 
+> 如果在Windows环境下，建议使用`iex.bat --werl`
```
nscyclone: Thanks, @Daqu!
I suppose that adding a section about `--werl` to the English translation could help us a lot. Would you like to take a stab at it?
I am also not sure whether we should bump both the minor and patch versions at the same time? I'd suggest to make it `1.1.0`. What do you think about that?

Daqu: 😁@nscyclone
1.I will take a stab at it in en version
2.I agree.I will follow your version(en) when I proof cn version

Daqu:
resolve conficts

doomspork:
Awesome, thank you @Daqu!

PR #1186, opened 2017-10-03
Show contributors in each file by commit

### LINKED ISSUE #927: Show who contributed to the page based on commits in one file
ex: https://webpack.js.org/configuration/entry-context/

screen shot 2017-01-17 at 8 46 11 pm

iboss-ptk:
Resolve #927

Using github api to query username from commit by file path.

There are 2 edge cases that haven't covered yet.
1. page can have only less than or equal 100 contributors due to github api per_page max is 100 or else need to fetch until next page is empty.
2. rate_limit for the api is 60 per hour so if we refresh page or navigate to new page more than that it  won't show contributors within 1 hour.

### DIFF
--- a/_assets/javascripts/contributors.js
+++ b/_assets/javascripts/contributors.js
@@ -8,7 +8,9 @@
   };
 
   $.fn.contributors = function() {
-    var filePath = window.location.pathname.replace(/^\/(.*)\/$/, "$1.md");
+    var strippedPath = window.location.pathname.replace(/^\/(.*)\/$/, "$1");
+    var isIndex = strippedPath.indexOf('/') === -1
+    var filePath = isIndex ? strippedPath + '/index.md' : strippedPath + '.md'
     var commitsApiEndpoint = "https://api.github.com/repos/elixirschool/elixirschool/commits?path=" + filePath + "&per_page=100";
     var self = this;
     var statusElement = $(self).find(".status");
--- a/_assets/javascripts/contributors.js
+++ b/_assets/javascripts/contributors.js
@@ -22,15 +22,16 @@
         return commit.author.login;
       });
 
-      unique(usernames).forEach(function(username) {
-        var content = "" +
-          "<a class=\"tooltip\" href=\"https://github.com/" + username + "\">" +
-            "<img src=\"https://github.com/" + username + ".png?size=50\" >" +
-            "<span class=\"tooltiptext\">" + username + "</span>" +
-          "</a>";
+      var content = unique(usernames)
+        .reduce(function(accHtml, username) {
+          return accHtml +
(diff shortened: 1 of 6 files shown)


doomspork:
@iboss-ptk this is awesome, I hadn't considered this approach.  Originally I was thinking this was something we'd try to incorporate into a Jekyll plug-in but this might actually prove to be a more flexible solution.

I'll take a peek at this today and let you know if I have any feedback 👍

ybur-yug:
This looks awesome @iboss-ptk, I'll pull it down tonight to get a look at things. You rock!

iboss-ptk:
The reason why I choose this approach is I've tried the jekyll plugin approach and it's doesn't really workout quite well because of the github api rate limit, so if we want to populate contributors in every file, it will fire a lot of requests while generating pages, and that takes long time to complete, and by doing that, github will not allow you to request for more at some point because we have more than 60 files which is the rate limit, which I think will result in bad contributor's experience.

While this approach allows us to delegate the api request responsibilty to the client, which will not downgrade the contributor's experience but the reader's experience is a little bit unpredictable. 😅

_assets/javascripts/contributors.js
```diff
@@ -0,0 +1,43 @@
...
+      });
+
+      unique(usernames).forEach(function(username) {
+        var content = "" +
+          "<a class=\"tooltip\" href=\"https://github.com/" + username + "\">" +
+            "<img src=\"https://github.com/" + username + ".png?size=50\" >" +
+            "<span class=\"tooltiptext\">" + username + "</span>" +
+          "</a>";
+
+        $(self).append(content);
```
nscyclone: I think that it's better to call `$.append()` only once outside of this `forEach`.

_assets/javascripts/contributors.js
```diff
@@ -22,15 +22,16 @@
...
-            "<span class=\"tooltiptext\">" + username + "</span>" +
-          "</a>";
+      var content = unique(usernames)
+        .reduce(function(accHtml, username) {
+          return accHtml +
+            "<a class=\"tooltip\" href=\"https://github.com/" + username + "\">" +
+              "<img src=\"https://github.com/" + username + ".png?size=50\" >" +
+              "<span class=\"tooltiptext\">" + username + "</span>" +
+            "</a>";
+        }, "")
```
iboss-ptk: How about this @nscyclone

nscyclone: Looks great, @iboss-ptk!

doomspork:
@iboss-ptk I will be testing and merging this shortly, what is your twitter handle?  😀

iboss-ptk:
@doomspork it's @ibossptk 😄

doomspork:
Thank you, thank you, thank you @iboss-ptk!  This is an _awesome_ feature to have 👍

PR #1187, opened 2017-10-04
Fix Erlang timer.tc function returns time in microseconds not miliseconds

crabonature:
8 ms for just three multiplications? This must be slow as 🐌


doomspork:
Thank you @crabonature! 😀

PR #1193, opened 2017-10-06
 [EN] Fix a misprint in the `advanced/metaprogramming` lesson

nscyclone:
In the previous sentence `Macro.expand_once/2` is the latter one, not the former.
By the way, shouldn't we create an `@elixirschool/english` team for this kind of issues?


doomspork:
@nscyclone that's a good idea.  I'll create the English team today 👍

doomspork:
@nscyclone we now have @elixirschool/english 👍

PR #1195, opened 2017-10-07
New Elixir and Erlang versions

crabonature:
Version update.


doomspork:
Thank you @crabonature!

PR #1197, opened 2017-10-08
Update of error messages in basics/functions examples

crabonature:
https://github.com/elixir-lang/elixir/blob/144df88caedfae891e12fcd3595c61b64783fd12/lib/elixir/src/elixir_def.erl#L342

https://github.com/elixir-lang/elixir/blob/23517c5413f3830daaf69d859fe7418085b21268/lib/elixir/lib/exception.ex#L836


doomspork:
💥  Thank you for being on top of this @crabonature, this is fantastic!

PR #1204, opened 2017-10-09
[EN] Update mix generated test file

crabonature:
Changed by:
https://github.com/elixir-lang/elixir/blame/dd23bef088cd5014da54e2fc3459d782612640fc/lib/mix/lib/mix/tasks/new.ex#L404

https://github.com/elixir-lang/elixir/blame/master/lib/mix/lib/mix/tasks/new.ex#L451

I've added some paragraph too about that doctest was run from the generated file by mix.

### DIFF
--- a/en/lessons/basics/testing.md
+++ b/en/lessons/basics/testing.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.0
+version: 1.1.0
 title: Testing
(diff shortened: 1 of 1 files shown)


doomspork:
Thank you for getting involved @crabonature, really appreciate your contributions 👍

PR #1210, opened 2017-10-12
[PT] Translations: Update basics lesson

anderkonzen:
These changes bring the Portuguese basics lesson up-to-date with the English one. The content itself has not changed, so I made just a few grammar corrections.

I also added back the elixir and iex command line examples in the English version, I believe they were removed incorrectly (let me know if it's not the case).


doomspork:
Thank you @anderkonzen!  We'll have someone take a peek at this and let you know if there's any feedback 😁

PR #1222, opened 2017-10-16
[PT] Fix specification definition

clairton:


doomspork:
Thank you and welcome @clairton 👍

PR #1225, opened 2017-10-17
Fix with/1 refactoring example inconsistency between case/2 and with/1 versions

### LINKED ISSUE #1224: Code versions of the with/1 refactoring example are inconsistent
The code example that shows a benefit of using `when/1` over `case/2` on some cases is inconsistent and a bit misleading (check [when/1 section of Control Structures](https://elixirschool.com/en/lessons/basics/control-structures/#with))

The variables used on the `Guardian.encode_and_sign` call are different in both examples and this could lead the reader to misunderstand the example or take more time to actually get it. Check below:

`case/2` version: 

```elixir
case Repo.insert(changeset) do
  {:ok, user} ->
    case Guardian.encode_and_sign(resource, :token, claims) do
      {:ok, jwt, full_claims} ->
        important_stuff(jwt, full_claims)
      error -> error
    end
  error -> error
end
```
`with/1` version:
```elixir
with {:ok, user} <- Repo.insert(changeset),
     {:ok, jwt, full_claims} <- Guardian.encode_and_sign(user, :token),
     do: important_stuff(jwt, full_claims)
```

Therefore it would be good to update the example to prevent the reader from thinking that the selection of the parameters could be related to the usage of `with/1` or `case/2`.

araujoigor:
Fixes #1224.

I updated all the existing examples (14 occurrences). Hopefully, the PR is good :)

Languages affected:

* `bg`
* `bn`
* `de`
* `en`
* `gr`
* `id`
* `jp`
* `ko`
* `no`
* `pl`
* `ru`
* `sk`
* `tr`
* `vi`

### DIFF
--- a/bg/lessons/basics/control-structures.md
+++ b/bg/lessons/basics/control-structures.md
@@ -154,7 +154,7 @@ iex> with {:ok, first} <- Map.fetch(user, :first),
 ```elixir
 case Repo.insert(changeset) do
   {:ok, user} ->
-    case Guardian.encode_and_sign(resource, :token, claims) do
+    case Guardian.encode_and_sign(user, :token, claims) do
       {:ok, jwt, full_claims} ->
         important_stuff(jwt, full_claims)
       error -> error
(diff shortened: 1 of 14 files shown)


doomspork:
Thank you and welcome @araujoigor!  This is a wonderful change, thank you for noticing it and contributing the fix 👍

PR #1229, opened 2017-10-21
[DE Add IEx Helpers lesson in german

stfnsr:
German translation for IEx Helpers, completing the Basics in german
#922


doomspork:
Thank you @stfnsr!!

PR #1233, opened 2017-10-26
Fixed a typo in collections lesson

kalarani:


doomspork:
@kalarani there appears to be a conflict here, could you look into that?  Thank you!

kalarani:
@doomspork - Have resolved the conflict. Please let me know if there is any other issue. Thanks!

doomspork:
@kalarani I'm not sure about this change, is the addition of the whitespace necessary?

kalarani:
@doomspork - The change is actually in the version / title table in the markdown. When i initially created the file, it didnt come up as a table, instead it was considered as a heading. Couldn't figure out the reason, but adding this whitespace helped to resolve the issue.

doomspork:
Hmm 🤔  We haven't had to do this before, maybe one of the @elixirschool/developers could chime in.  I think it's perhaps a missing line break rather than an extra space?

ta/lessons/basics/collections.md
```diff
@@ -1,6 +1,6 @@
-﻿---
+---
```
nscyclone: Thanks for pointing this out, @kalarani! The problem was caused by BOM, so I've removed it from the lessons containing it in #1241.

nscyclone:
Thanks, @kalarani! The issue is fixed in #1241.

kalarani:
Thanks for the fix @nscyclone ! Looks like it was caused in both the indian language translations :)

PR #1236, opened 2017-10-26
Favicon closes #1203

### LINKED ISSUE #1203: Favicon
I would love for there to be a favicon so it would be easier to find the elixirschool tabs I have open. Has this been discussed anywhere?

squaresurf:
Why:

* It's very useful for those that use tabs as they can quickly identify
  which tab is ElixirSchool.

This change addresses the need by:

* Generate a favicon set from a squared og_image.jpg on
  https://realfavicongenerator.net/.
* Add generated favicon markup to the html head.
* Add generated files to the assets.


squaresurf:
I am by no means a designer, but I did my best to rework the og_image.jpg into a square image that could be used for a favicon. I would not be offended in the least if someone with a better design background wants to create a better image.

doomspork:
@squaresurf should we just drop the ES for now since it's hard to see in the icon?  I'm going to work with a designer on a look and icon for the Elixir School org, we can aim for a better favicon then.

squaresurf:
@doomspork that sounds good. My only concern was that it might be too close to the official elixir logo. Will there be any copyright issues with that?

doomspork:
@squaresurf as far as I am aware we're in the clear, we've been using this logo for awhile now.

squaresurf:
@doomspork I've updated the favicon to not include the ES. Let me know what you think.

doomspork:
Wonderful, thank you @squaresurf!

PR #1237, opened 2017-10-26
Add a .tool-versions file.

squaresurf:
Why:

* To set the ruby version via [asdf](https://github.com/asdf-vm/asdf)

This change addresses the need by:

* `asdf local ruby 2.3.1`


squaresurf:
No worries if this is not something desired and I can close it out. I found it useful when I was working on the site.

doomspork:
@squaresurf I think we'll hold off on this for now, thanks for the suggestion though 👍

PR #1241, opened 2017-11-01
Remove BOM from `.md` files

nscyclone:
Jekyll docs [state](https://jekyllrb.com/docs/frontmatter/) that BOM should be avoided.
This PR fixes a problem uncovered in #1233 and the similar one with another translation.


doomspork:
💥 Awesome catch @nscyclone

PR #1243, opened 2017-11-05
Fix warning in using `else` alongside `with` example code block.

ramytamer:
The example code block for using `else` with `with` throws a warning for importing `Integer` and not using it; fixed by removing `Integer.is_even` and just calling `is_even` directly.

### DIFF
--- a/bn/lessons/basics/control-structures.md
+++ b/bn/lessons/basics/control-structures.md
@@ -183,7 +183,7 @@ import Integer
 m = %{a: 1, c: 3}
 
 a = with {:ok, res} <- Map.fetch(m, :a),
-  true <- Integer.is_even(res) do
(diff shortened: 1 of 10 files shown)


doomspork:
Howdy @ramytamer, thanks for pointing this out!  Should we correct this warning or remove the import?  I'm leaning towards the latter.

ramytamer:
@doomspork we should keep the import and call the function without the module as in some other next lesson ([Comprehensions](https://elixirschool.com/en/lessons/basics/comprehensions/#filters)) In the filter example we `import Integer` and call the `is_even` directly without the module. So I think we should do it that way for consistency.

ramytamer:
@doomspork Done.

doomspork:
Thank you @ramytamer!

Issue #1043, opened 2017-04-04
Translate Plug lesson 1.1.0


moxley:
Major revisions have been made to the Plug lesson, in #861 and #1033. Translations need to be created.

## Translation

Translation progress:

- [x] en
- [ ] ru /cc @brain-geek / @nscyclone
- [ ] ko /cc @devleoper
- [ ] jp /cc @3100
- [ ] no /cc @therobinsj
- [ ] pl /cc @Koziolek
- [ ] sk /cc @michalvalasek
- [ ] fr /cc @julienXX
- [ ] cn /cc @cizixs
- [x] es /cc @aleandros / @jonahoffline / @edwinallenz
- [ ] pt /cc @pragmaticivan
- [ ] vi /cc @joneslee85
- [ ] it /cc @lucidstack / @sformisano
- [ ] my /cc @alkaha
- [ ] id /cc @rezaprima

edwinallenz:
While I'm translating this lesson to spanish, I'm also doing the Plug exercise. As I'm doing the section https://github.com/doomspork/elixir-school/blob/master/lessons/specifics/plug.md#adding-another-plug 
I'm getting a 500 response when a POST is performed to the /upload endpoint, and I'm not able to see any error message in the terminal.

Other endpoints as `/` or other are responding well with 200 and 404
This is my repo if you're able to  give it a shot
https://github.com/edwinallenz/plug_lesson

thanks

gemantzu:
Greek #1128

PR #1244, opened 2017-11-06
[JP] translate plug 1.1.1 to jp

iguchi1124:
issue https://github.com/elixirschool/elixirschool/issues/1043

I'm trying to translate lesson Plug 1.1.0 into japanese.

### DIFF
--- a/jp/lessons/specifics/plug.md
+++ b/jp/lessons/specifics/plug.md
@@ -1,15 +1,35 @@
 ---
-version: 0.9.0
+version: 1.1.1
 title: Plug
+redirect_from:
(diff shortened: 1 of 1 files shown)


doomspork:
Thank you and welcome @iguchi1124!  We'll see if the @elixirschool/japanese translators have any thoughts 😁

iguchi1124:
Now I completely translated this docs into japanese! please reviews! @elixirschool/japanese

doomspork:
Thank you @iguchi1124!  If @elixirschool/japanese doesn't get back to you in the next few days, we can merge this and let folks submit PRs to correct anything 👍

PR #1245, opened 2017-11-06
Fix user input shell script to unifying doller sign prefixed

iguchi1124:


doomspork:
Wonderful correction, thank you @iguchi1124!!

Issue #1025, opened 2017-03-22
Translation of the version messages


nscyclone:
The original changes can be found [here](https://github.com/doomspork/elixir-school/blob/d37d8aea1ccd518e21a431d5d1fd98b84703df81/_config.yml#L197-L203).

## Translation
- [x] en #961
- [ ] bg /cc @merkata
- [ ] cn /cc @cizixs
- [x] de /cc @Lechindianer #1027
- [x] es /cc @aleandros / @jonahoffline / @edwinallenz #1036
- [ ] fr /cc @julienXX
- [x] gr /cc @gemantzu #1023
- [ ] id /cc @rezaprima
- [ ] it /cc @lucidstack  / @sformisano
- [ ] jp /cc @3100
- [x] ko /cc @devleoper #1089
- [ ] my /cc @alkaha
- [ ] no  /cc @Devalo
- [ ] pl /cc @Koziolek 
- [ ] pt /cc @pragmaticivan
- [x] ru /cc @nscyclone @brain-geek #1024
- [ ] sk /cc @michalvalasek
- [ ] tr /cc @hasanozgan 
- [ ] uk /cc @brain-geek 
- [ ] vi /cc @kiennt

PR #1249, opened 2017-11-07
[JP] Add version_messages

iguchi1124:
I added japanese translation of version messages :smile:
issue https://github.com/elixirschool/elixirschool/issues/1025


doomspork:
Thank you thank you! ❤️

Issue #947, opened 2017-02-05
Translation of GenEvent Removal


ybur-yug:
Original PR Here #946.

This removes any reference to GenEvent, as it is being deprecated.
This issue also has the latest list of all languages and contributors to use as an issues template. I added some that are typically missing.

## Translation
- [x] en #946
- [x] bg
- [x] bn
- [ ] cn /cc @cizixs
- [ ] de
- [x] es /cc @aleandros / @jonahoffline / @eze1981
- [x] fr /cc @julienXX
- [x] gr /cc @gemantzu
- [ ] id /cc @rezaprima
- [x] it /cc @sformisano
- [x] jp /cc @3100
- [x] ko /cc @devleoper 
- [x] my /cc @alkaha
- [x] no  /cc @Devalo
- [x] pl /cc @Koziolek #949
- [x] pt /cc @pragmaticivan
- [x] ru /cc @nscyclone @brain-geek #948
- [x] sk /cc @michalvalasek
- [x] tr /cc @hasanozgan 
- [x] uk /cc @brain-geek 
- [ ] vi /cc @joneslee85

eksperimental:
Only 5 languages left to close this issue
```
$ ag genevent -l | sort
cn/lessons/advanced/otp-concurrency.md
de/lessons/advanced/otp-concurrency.md
id/lessons/advanced/otp-concurrency.md
pt/lessons/advanced/otp-concurrency.md
vi/lessons/advanced/otp-concurrency.md
```

/cc @elixirschool/chinese @elixirschool/indonesian @elixirschool/portuguese @elixirschool/vietnamese

eksperimental:
Could you guys translate this into your language:

> In this lesson we'll focus on **the biggest piece: GenServers**.

the rest is already being taken care of in [here](#1412)

here are the old lines, for reference:
- Chinese: 这节课，我们主要讲两个东西：Genservers 和 GenEvents。
- German: In dieser Lektion werden wir uns hauptsächlich mit zwei wichtigen Teilen beschäftigen: GenServers und GenEvents.
- Indonesian: Dalam pelajaran ini kita akan fokus pada dua bagian penting: Genserver dan GenEvent.
- Portuguese: Nessa lição vamos nos focar em duas peças importantes: GenServers e GenEvents.
 - Vietnamese: Trong bài này, chúng ta sẽ tập trung vào hai phần chính: GenServers và GenEvents.

anderkonzen:
[...]

PR #1250, opened 2017-11-07
[JP] Remove references to GenEvent

iguchi1124:
issue https://github.com/elixirschool/elixirschool/issues/947
I removed `GenEvent` references.


doomspork:
@iguchi1124 does the version need to be updated in the meta data?  😁

riseshia:
This origin pr doesn't have version though, do we need it? 
https://github.com/elixirschool/elixirschool/pull/946/files
Or how about go version up to latest at once? @iguchi1124

doomspork:
Good catch @iguchi1124 👍

PR #1262, opened 2017-11-17
[KO] Update ecto to 1.1.1

marocchino:


doomspork:
Thank you very much @marocchino 😁

PR #1268, opened 2017-11-27
Fix an error in the version definition of pipe-operator lesson

ruslanshakirov:
English version of pipe-operator lesson wasn't formatted as others.
This commit fixes next error: There is an error in the version definition of this chapter.
[1, 0, 0] vs "1.0.1"

### DIFF
--- a/en/lessons/basics/pipe-operator.md
+++ b/en/lessons/basics/pipe-operator.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.0
+version: 1.0.1
 title: Pipe Operator
 redirect_from:
   - /lessons/basics/pipe-operator/
@@ -32,14 +32,14 @@ For this set of examples, we will use Elixir's String module.
 - Tokenize String (loosely)
 
 ```elixir
-iex> "Elixir rocks" |> String.split
+iex> "Elixir rocks" |> String.split()
 ["Elixir", "rocks"]
 ```
 
 - Uppercase all the tokens
 
 ```elixir
-iex> "Elixir rocks" |> String.upcase |> String.split
+iex> "Elixir rocks" |> String.upcase() |> String.split()
 ["ELIXIR", "ROCKS"]
 ```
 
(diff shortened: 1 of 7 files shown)


en/lessons/basics/pipe-operator.md
```diff
@@ -1,5 +1,5 @@
 ---
-version: 1.0.0
+version: 1.0.1
```
nscyclone: Currently in case when the translation's version is higher than the original lesson's version is considered an error. I am not sure whether this kind of behaviour is correct.
/cc @elixirschool/developers

en/lessons/basics/pipe-operator.md
```diff
@@ -32,14 +32,14 @@ For this set of examples, we will use Elixir's String module.
 - Tokenize String (loosely)
 
 ```elixir
-iex> "Elixir rocks" |> String.split
+iex> "Elixir rocks" |> String.split()
```
nscyclone: A quick lookup for `String.split$` has detected a couple more places with stale formatting, would you like to take a stab at those too, @ruslanshakirov?

nscyclone:
Nice catch, thanks @ruslanshakirov!

en/lessons/basics/pipe-operator.md
ruslanshakirov: Yes, I'll fix these places.

ruslanshakirov:
@nscyclone please review new commit.

ruslanshakirov:
Thank you for review, @nscyclone!

en/lessons/basics/pipe-operator.md
doomspork: @nscyclone that's an interesting point.  I would imagine only the patch version would increment out of sync with the english lesson?

doomspork:
Thank you @ruslanshakirov 💜

en/lessons/basics/pipe-operator.md
nscyclone: @doomspork I think so. When there is an update to an actual translation we just bump the patch version. Should we remove the warning for patch versions mismatch?

doomspork: I think that is a reasonable solution @nscyclone 👍

PR #1269, opened 2017-11-29
Changes link to point correctly to en subfolder

gemantzu:
Is there a better way than hardcoding the default language?


doomspork:
@gemantzu that's a good question.  As we work on the new site, we should try to tackle some of these.

PR #1270, opened 2017-11-29
Suppress the patch version mismatch warning

nscyclone:
Discussed in #1268. This PR disables the warning when the translation's patch version is higher than the original lesson's one.


doomspork:
Thank you @nscyclone ❤️

PR #1271, opened 2017-11-30
Escripts: Better practice in mixfile

eamontaaffe:
Similar to how the deps are initialized in a fresh mix project, we should declare the escript argument using brackets (to signify that it is a function) and the function itself should be private (because it is never accessed outside of the module).

## Translations

- [x] cn
- [x] de
- [x] en
- [x] es
- [x] gr
- [x] id
- [x] jp
- [x] ko
- [x] pl
- [x] pt
- [x] ru
- [x] sk
- [x] vi

## Not applicable translations

The following translations do not contain the relevant section.

- ar
- bg
- bn
- fr
- it
- my
- no
- ta
- th
- tr
- tw
- uk

### DIFF
--- a/cn/lessons/advanced/escripts.md
+++ b/cn/lessons/advanced/escripts.md
@@ -25,10 +25,10 @@ end
 ```elixir
 defmodule ExampleApp.Mixfile do
   def project do
-    [app: :example_app, version: "0.0.1", escript: escript]
+    [app: :example_app, version: "0.0.1", escript: escript()]
   end
 
-  def escript do
+  defp escript do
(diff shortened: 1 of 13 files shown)


doomspork:
Great enhancement @eamontaaffe!  Would you mind copying this change to the other translations?  If not, one of us can do it 😁

eamontaaffe:
Yep I'll do it.

eamontaaffe:
@doomspork: that's all been updated now

doomspork:
Thank you @eamontaaffe ❤️

PR #1277, opened 2017-12-09
Add enhancement in explanation for Process Linking in Concurrency lesson

### LINKED ISSUE #1276: Suggestion for Process Linking lesson
Reading through [Process Linking](https://elixirschool.com/en/lessons/advanced/concurrency/#process-linking), I found the example code about trapping the exits confusing at first. I didn't know if `:trap_exit` was being defined or was a standard Elixir thing or something else. I did some reading on Elixir's documentation, and maybe this is what the lessons are trying to encourage (there's nothing wrong with that), but it sounds like it could be more clear changing the text for something like this:

>Sometimes we don’t want our linked process to crash the current one. For that we need to trap the exits. To trap the exits, we use the `Process` module's `flag/2` function. This module uses [erlang's `process_flag` function](http://erlang.org/doc/man/erlang.html#process_flag-2) for the `trap_exit` flag. When trapping exits (`trap_exit` is set to `true`), exit signals will be received as a tuple message: `{:EXIT, from_pid, reason}`.

picandocodigo:
Fixes #1276

### DIFF
--- a/en/lessons/advanced/concurrency.md
+++ b/en/lessons/advanced/concurrency.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.1
+version: 1.1.0
 title: Concurrency
 redirect_from:
   - /lessons/advanced/concurrency/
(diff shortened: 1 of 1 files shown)


en/lessons/advanced/concurrency.md
```diff
@@ -1,5 +1,5 @@
 ---
-version: 1.0.1
```
doomspork: Let's call this `1.1.0` so the translations get the appropriate warning.

en/lessons/advanced/concurrency.md
```diff
@@ -83,7 +83,7 @@ iex> spawn_link(Example, :explode, [])
 ** (EXIT from #PID<0.57.0>) evaluator process exited with reason: :kaboom
 ```
 
-Sometimes we don't want our linked process to crash the current one.  For that we need to trap the exits.  When trapping exits they will be received as a tuple message: `{:EXIT, from_pid, reason}`.
+Sometimes we don't want our linked process to crash the current one. For that we need to trap the exits. To trap the exits, we use the `Process` module's `flag/2` function. This module uses [erlang's process_flag function](http://erlang.org/doc/man/erlang.html#process_flag-2) for the `trap_exit` flag. When trapping exits (`trap_exit` is set to `true`), exit signals will be received as a tuple message: `{:EXIT, from_pid, reason}`.
```
doomspork: Could we change this to: "we use `Process.flag/2`" and "erlang's [process_flag](...) function"?

picandocodigo:
@doomspork Updated with the suggested changes 💪

doomspork:
💥  Thank you @picandocodigo!

PR #1287, opened 2017-12-25
Improve the section that clarifies the exits trapping

nscyclone:
I decided not to bump the version numbers since there are no translations of the latest version yet.


doomspork:
Wonderful, thank you @nscyclone 😀

PR #1184, opened 2017-10-02
Thai translation for basics page

iboss-ptk:
* Thai basics page translation
* patch `en/lessons/basic.md` to adhere shell prompt style guide
* refine some translation and fix typo in `th/index.md`

PR #1143, opened 2017-08-01
minor syntax changes

RichMorin:

PR #1296, opened 2018-01-07
[KO] Update basics/basics 1.1.0 -> 1.1.2

riseshia:
## Ref

- 1.1.0 -> 1.1.1 #1143
- 1.1.1 -> 1.1.2 #1184

cc @devleoper @doomspork

### DIFF
--- a/ko/lessons/basics/basics.md
+++ b/ko/lessons/basics/basics.md
@@ -15,7 +15,7 @@ elixir-lang.org 홈페이지의 [Installing Elixir](http://elixir-lang.org/insta
 
 Elixir를 설치하고 나서 어떤 버전이 설치되었는지 손쉽게 확인할 수 있습니다.
 
-    % elixir -v
+    $ elixir -v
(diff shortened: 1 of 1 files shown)


riseshia:
Question to @doomspork,
Currently, I can not merge my own pull request, but only from others. Is it intended permission? ;)

riseshia:
> You need at least one 👍 review from other folks to merge it and it looks like now you can merge it :shipit:

actually, that's the point... korean is better case, but  japanese is not the case. ;_; anyway Thanks to approve!

doomspork:
@riseshia it's unfortunately a limitation of all of the different permissions of GitHub.  To keep master locked and require a review, GitHub won't allow a person to merge their own PRs.  It's not ideal but we've previously had a handful instances where malformed markdown in a PR took the site offline so this "roadblock" forces a second pair of eyes on the PR.

If you @ me in the future I'll get things merged for you ❤️ 

Sorry for the trouble but thanks for all of your hard work 😁

riseshia:
🙆‍♂️

PR #1307, opened 2018-01-13
[KO] Bump up version of basics/basics

riseshia:
Forgive me that I forgot to bump up version in #1296 T-T

@doomspork @devleoper


doomspork:
Thank you @riseshia!!

Issue #1278, opened 2017-12-13
Fix the command in Repository section of Ecto


flavono123:
Following [the Ecto lesson](http://elixirschool.com/en/lessons/specifics/ecto/#repository), I've got this error:
```
$ mix ecto.gen.repo
...
warning: could not find Ecto repos in any of the apps: [:example_app].

You can avoid this warning by passing the -r flag or by setting the
repositories managed by those applications in your config/config.exs:

    config :example_app, ecto_repos: [...]

** (Mix) ecto.gen.repo expects the repository to be given as -r MyApp.Repo
```

When I run with `-r ExampleApp.Repo`, it works:
```
$ mix ecto.gen.repo -r ExampleApp.Repo
* creating lib/example_app
* creating lib/example_app/repo.ex
* updating config/config.exs
Don't forget to add your new repo to your supervision tree
(typically in lib/example_app/application.ex):

    supervisor(ExampleApp.Repo, [])

And to add it to the list of ecto repositories in your
configuration files (so Ecto tasks work as expected):

    config :example_app,
      ecto_repos: [ExampleApp.Repo]


```
`mix ecto.gen.repo` ➡️ `mix ecto.gen.repo -r ExampleApp.Repo`(or `mix ecto.gen.repo -r <ProjectName>.Repo`)
May I fix it? 🤔

doomspork:
@flavono123 hello and welcome!  Please do submit a fix for this 👍  If you'd like to go ahead and apply it to the translations that would be spectacular!

PR #1309, opened 2018-01-14
Fix command in Repository section of Ecto (#1278)

mbirman:
As proposed, changed `mix ecto.gen.repo` to `mix ecto.gen.repo -r ExampleApp.Repo` (all languages). See #1278.


doomspork:
Thank you @mbirman, I think this is a good change.  👍

PR #1312, opened 2018-01-15
[JP] Bump up basics/collections to 1.2.1

riseshia:
Could you check this? @doomspork


doomspork:
Thank you @riseshia for all your help 😁

PR #1317, opened 2018-01-20
[JP] Replace `method` to `function`

riseshia:
Please check this @doomspork @marocchino :pray:


doomspork:
@marocchino would you like to join the @elixirschool/japanese team so you can @riseshia can merge for one another?

PR #1323, opened 2018-01-23
[JA] Change Japanese language to "ja"

### LINKED ISSUE #1322: Change Japanese language code to "ja"
Actually jp is country code.

It is not conflict with other langs but, still wrong.

- set redirect_from 
- rename folders 
- modify some settings.

Note that #1255 has effected by this changes.

related #1109 

ref: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

marocchino:
close #1322 

@elixirschool/japanese  review please. :)


_data/interlang.yml
```diff
@@ -1,27 +1,27 @@
...
+#  "no":
+#  pl:
+#  pt:
+#  ru:
+#  sk:
+#  ta:
+#  th:
+#  tr:
+#  uk:
+#  vi:
```
eksperimental: the trailing space from all these lines should not be removed.

eksperimental: other than this, this PR looks good to me

marocchino: 🙆‍♂️ f4371af

doomspork:
Thank you @marocchino, @riseshia, and @eksperimental ❤️

PR #1325, opened 2018-01-23
Fixes in basic Pattern Matching pin operator

killondark:


doomspork:
Howdy @killondark!  Thanks for the PR but I'm going to close this for now.  The term `Mornin'` is a colloquialism from where I grew up so I'd like to leave this for now 😁 

I hope all is well!

eksperimental:
This is a good example of why code should bee translated as well. Morning' for somebody who doesnt speak English might not mean a thing. Should we make a decision on this and promote it if agreed on?

PR #1331, opened 2018-02-05
Add a section on race conditions with ETS tables

nathanl:
I'm not sure if this is the best place in the document to add this, but as soon as I read that `:public` access was possible, I wanted to know how race conditions could be prevented. After reading and experimenting, this is what I came up with.

### DIFF
--- a/en/lessons/specifics/ets.md
+++ b/en/lessons/specifics/ets.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.2
+version: 1.0.3
 title: Erlang Term Storage (ETS)
 redirect_from:
   - /lessons/specifics/ets/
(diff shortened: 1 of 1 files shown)


doomspork:
@nathanl I think this is a good addition, could you bunch the lesson's version?

  I'll let some of the others take a peek before we merge 👍

en/lessons/specifics/ets.md
```diff
@@ -1,5 +1,5 @@
 ---
-version: 1.0.2
+version: 1.0.3
 title: Erlang Term Storage (ETS)
 redirect_from:
   - /lessons/specifics/ets/
```
eksperimental: I would remove this redirect from new lessons. All it is doing is causing trouble when people translate lessons and it adds no benefit at all.

doomspork: Good catch @eksperimental 👍

eksperimental: My bad, this is an old lesson. For some reason I thought this was a brand new lesson. So the directive is needed.

eksperimental: @doomspork But it will be valid for the new Bypass lesson, https://github.com/elixirschool/elixirschool/commits/master/en/lessons/libraries/bypass.md

PR #1332, opened 2018-02-11
Remove redirect from newer lessons

doomspork:
Good catch @eksperimental, this should remove the redirect from the new lessons 👍


doomspork:
@eksperimental fixed 👍

Issue #1342, opened 2018-02-28
Add Enum.filter to Basic Enum Page


wilbertliu:
Hello folks 👋🏻,

Elixir School has been my go-to resource when I need to find something that I might be forgotten. Recently I wonder if there's a `filter` function on Enum and want to find its usage. After I went to the Basic Enum Page, I'm surprized because there's no `Enum.filter` section in it. That made me try to Google it because I'm not sure Elixir doesn't have such function, and found it afterwards.

I believe it might not only me that need such section to learn / recall.
Do we mind to add the `Enum.filter` section? I would be more than happy to contribute. 🙂

ybur-yug2:
@wilbertliu good eye! [We have a contributing guide here](https://github.com/elixirschool/elixirschool/blob/master/CONTRIBUTING.md) and if you can get up and running PRs with improvements are welcome! I personally think its absolutely fitting and would be a good addition.

PS I'm actually @ybur-yug  and am currently just locked out of that account because of an email snafu :P

doomspork:
@wilbertliu please do open a PR 👍 

We're always happy to have community help in enhancing the resource 😁

wilbertliu:
@ybur-yug2 Thanks for pointing that out. Will review the contributing guide then! 😀
@doomspork Sure thing Sean! Will seek a good time to do this. Thanks a lot!

PR #1349, opened 2018-03-16
Add Enum.filter/2 lesson (#1342)

wilbertliu:
Hello folks 👋🏻,

This PR is the real work based on discussion from #1342.
Any thoughts?

Thanks! 🙏🏻

### DIFF
--- a/en/lessons/basics/enum.md
+++ b/en/lessons/basics/enum.md
@@ -1,5 +1,5 @@
 ---
-version: 1.3.0
+version: 1.4.0
 title: Enum
 redirect_from:
   - /lessons/basics/enum/
--- a/en/lessons/basics/enum.md
+++ b/en/lessons/basics/enum.md
@@ -143,6 +143,15 @@ Enum.max([], fn -> :bar end)
 :bar
 ```
 
+### filter
+
+`filter/2` returns collection's elements that match the condition from the given function.
+
+```elixir
+iex> Enum.filter([1, 2, 3, 4], fn(x) -> rem(x, 2) == 0 end)
+[2, 4]
+```
+
 ### reduce
 
 With `reduce/3` we can distill our collection down into a single value.  To do this we supply an optional accumulator (`10` in this example) to be passed into our function; if no accumulator is provided the first element in the enumerable is used:
--- a/en/lessons/basics/enum.md
+++ b/en/lessons/basics/enum.md
@@ -145,7 +145,7 @@ Enum.max([], fn -> :bar end)
 
 ### filter
 
-`filter/2` returns collection's elements that match the condition from the given function.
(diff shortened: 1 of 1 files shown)


en/lessons/basics/enum.md
```diff
@@ -143,6 +143,15 @@ Enum.max([], fn -> :bar end)
 :bar
 ```
 
+### filter
+
+`filter/2` returns collection's elements that match the condition from the given function.
```
doomspork: How about something like:

> The `filter/2` function enables us to filter the collection to include only those elements that evaluate to `true` using the provided function.

wilbertliu: Sounds clearer Sean! 😀 Will update then!

nscyclone: Thanks, @wilbertliu!
Should we bump the minor version of the lesson?

wilbertliu: Hello @nscyclone, I'm not sure I understand what you intended to ask. Could you please elaborate more? 🙏🏻

nscyclone: @wilbertliu, there is a variable called [`version`](https://github.com/elixirschool/elixirschool/pull/1349/files#diff-dfa183a1580b77baec65b2a1f09e3948L2) in the [front matter](https://jekyllrb.com/docs/frontmatter/) of every lesson/translation that helps to keep an eye on differences between original lessons and their translations. Usually when we introduce changes to an existing file we also increase `major`, `minor` or `patch` version. You can find a little more information about this in [`README`](https://github.com/elixirschool/elixirschool#translated-lesson).

In my opinion, in this case update of the `minor` version number would fit well to reflect the incoming changes, WDYT?

wilbertliu: @nscyclone Ah, I see. And yes, this should be `minor`. Who is in charge to change that though? Me? Thanks a lot for explaining! 🙂

nscyclone: @wilbertliu Sorry for a late reply! Would be great if you'd take care of that on your own. Thanks in advance!

wilbertliu: @nscyclone No problem mate! I think I'm done? 😊

wilbertliu:
Hello @doomspork 🙂

I know it's already merged, but I couldn't see the lesson on the production, which is https://elixirschool.com/en/lessons/basics/enum/

Is that expected? But I could see my photo at the `Contributors` section.

Thanks! 🙏🏻

doomspork:
@wilbertliu there was an issue with Heroku deployments I have to look into today.

nscyclone:
@wilbertliu Glad to see it [live](https://elixirschool.com/en/lessons/basics/enum/#filter)! Thanks once again!

PR #1357, opened 2018-03-21
mix-tasks.md: fix typos and inconsistencies

chadoh:
I also removed the `.` from the end of the shortdoc, which was my own editorial preference. None of the other task descriptions  given by `mix help` have a `.` at the end, why should this one?

### DIFF
--- a/en/lessons/basics/mix-tasks.md
+++ b/en/lessons/basics/mix-tasks.md
@@ -61,7 +61,7 @@ Now, in our **lib/hello.ex** file that Mix generated for us, let's create a simp
 ```elixir
 defmodule Hello do
   @doc """
-  Output's `Hello, World!` everytime.
+  Outputs `Hello, World!` every time.
   """
   def say do
     IO.puts("Hello, World!")
@@ -77,7 +77,7 @@ Let's create our custom Mix task. Create a new directory and file **hello/lib/mi
 defmodule Mix.Tasks.Hello do
   use Mix.Task
 
-  @shortdoc "Simply runs the Hello.say/0 command."
+  @shortdoc "Simply runs the Hello.say/0 function"
   def run(_) do
     # calling our Hello.say() function from earlier
     Hello.say()
@@ -110,6 +110,6 @@ $ mix help
 
 mix app.start         # Starts all registered apps
 ...
-mix hello             # Simply calls the Hello.say/0 function.
+mix hello             # Simply calls the Hello.say/0 function
 ...
 ```


en/lessons/basics/mix-tasks.md
```diff
@@ -61,7 +61,7 @@ Now, in our **lib/hello.ex** file that Mix generated for us, let's create a simp
 ```elixir
 defmodule Hello do
   @doc """
-  Output's `Hello, World!` everytime.
+  Outputs `Hello, World!` every time.
```
nscyclone: Good catch!

en/lessons/basics/mix-tasks.md
```diff
@@ -77,7 +77,7 @@ Let's create our custom Mix task. Create a new directory and file **hello/lib/mi
 defmodule Mix.Tasks.Hello do
   use Mix.Task
 
-  @shortdoc "Simply runs the Hello.say/0 command."
+  @shortdoc "Simply runs the Hello.say/0 function"
```
nscyclone: I find this change good but the previous version [has been copied across multiple translations](https://github.com/elixirschool/elixirschool/search?utf8=%E2%9C%93&q=shortdoc). So I think we should update them as well in order to remain consistent.

chadoh:
You mean the period/full stop, or the `s/command/function/`?

en/lessons/basics/mix-tasks.md
chadoh: You mean the period/full stop, or the `s/command/function/`?

I changed it here because I saw that it reads "function" in the example output below. We could alternatively make both say "command".

nscyclone: In places where the whole `@shortdoc` is copied from the English lesson we could introduce both `s/command/function/` and period removal. If I am not mistaken, those are 12 + the English one. With the others we have the following situation:

* `de` and `sk` already use `function` (`Funktion`/`funkciu`);
* `gr` and `ru` use `command` (`εντολή`/`команду`).

I'd suggest to only remove the trailing period from `de`, `sk` and `gr`, and I can take care of the Russian one on my own if you don't mind.
After all those changes only `gr` will differ from the others and I am not sure what should we do about that. Create an issue, ping Greek translators here for help or simply ignore it?

eksperimental: bump the minor version of the English lesson. that way translators can find out the change.
If the translations you are updating `function` with `command` are the latest version, bump up the version there as well, if not simply update and leave the version as is.

chadoh:
Not worth the effort

doomspork:
No worries @chadoh, I took care this on https://github.com/elixirschool/elixirschool/pull/1404 👍

PR #1367, opened 2018-03-28
Add extra note to "function clause pinning" section

jordan-brough:
To show that the reassignment of the `greeting` variable is local to the
function and doesn't affect the outer scope.


doomspork:
Thank you and welcome @jordan-brough 😁

PR #1371, opened 2018-03-29
fixes out-dated URL for "Expressions allowed in guard clauses"

sgnl:
Noticed that this URL was out of date and found the URL.

### DIFF
--- a/bg/lessons/basics/control-structures.md
+++ b/bg/lessons/basics/control-structures.md
@@ -91,7 +91,7 @@ iex> case {1, 2, 3} do
 "Will match"
 ```
 
-Проветеве официалната документация за [Допустими изрази в предпазващи клаузи](http://elixir-lang.org/getting-started/case-cond-and-if.html#expressions-in-guard-clauses).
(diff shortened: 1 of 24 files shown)


bg/lessons/basics/control-structures.md
```diff
@@ -91,7 +91,7 @@ iex> case {1, 2, 3} do
 "Will match"
 ```
 
-Проветеве официалната документация за [Допустими изрази в предпазващи клаузи](http://elixir-lang.org/getting-started/case-cond-and-if.html#expressions-in-guard-clauses).
+Проветеве официалната документация за [Допустими изрази в предпазващи клаузи](https://hexdocs.pm/elixir/master/guards.html).
```
nscyclone: Thanks, @sgnl!
I am not sure whether we should link to the `master` docs directly. Isn't [`stable`](https://hexdocs.pm/elixir/guards.html) a better choice?

doomspork: I think that's a good idea @nscyclone 👍

sgnl:
Thanks for the review. I've updated the URL to be:

`https://hexdocs.pm/elixir/guards.html#list-of-allowed-expressions`

doomspork:
Thank you @sgnl and @nscyclone 👍

PR #1372, opened 2018-03-29
Allow plug ex. to set plug_status on exception

justanotherdot:
Without `use Plug.ErrorHandler` every raised exception would default to
a status code of 500. This resolves that. The example could probably
also include the overwriting of the `handle_errors/2` callback but it
may not be necessary especially considering that it looks like Plug is
(thankfully) trying to get people to not use exceptions for regular
control flow. The example will fallback to body text of "Something went
wrong", however.


doomspork:
@justanotherdot this is a good start for now.  Thanks and welcome 😁

PR #1374, opened 2018-03-30
[EN] Semicolon replaced with a dot

rafalbig:
Hi All,

I think I noticed something that might be considered as a small mistake. Since the phrase ` You can find their implementation` starts with upper Y it might be a good idea to replace semicolon with a dot.

Hope this helps.

Cheers,
Rafal


doomspork:
Thank you @rafalbig 👍

Issue #1029, opened 2017-03-26
OTP Supervisors section needs clarification


ChuckIrvine:
To preface things, I really like your "Elixir School". Nice simple examples. I despise having to wade though someone's very complex example to get to the gist of what is being explained. Thanks!!!

Goto:
https://elixirschool.com/lessons/advanced/otp-supervisors/

The context for the code snippet below is unclear to me. Where should it be placed in the code for SimpleQueue

###########
import Supervisor.Spec
children = [
  worker(SimpleQueue, [], [name: SimpleQueue])
]
{:ok, pid} = Supervisor.start_link(children, strategy: :one_for_one)
############

doomspork:
@ChuckIrvine thanks for reaching out!  You're right, this is unclear and something we should improve.  Were you able to find the answer you needed?

9mm:
I'd like to add to this that it looks like the syntax for doing a proper supervisor changed after 1.5, and the method mentioned in this whole section is deprecated (anything with `import Supervisor.Spec` if I'm not mistaken).

I'm super confused trying to find a new/good example that isn't this really complicated docs page...https://hexdocs.pm/elixir/master/Supervisor.html

chuckhousley:
The Ecto page also uses the Supervisor.Spec module.  Should the examples be rewritten completely or should there be code snippets for pre-v1.5 like in the [ecto docs](https://github.com/elixir-ecto/ecto/blob/master/guides/Getting%20Started.md#adding-ecto-to-an-application)?

doomspork:
@9mm and @chuckhousley we should update the code to reflect the newer syntax. Would one of you like to take the lead on that?

chuckhousley:
If there are no objections I'd like to take it

9mm:
@chuckhousley  its all you

doomspork:
Thank you both!

notactuallypagemcconnell:
@chuckhousley @doomspork closing this as the merged PR resolves the issue

PR #1378, opened 2018-04-01
Update supervisor examples to Elixir 1.6

chuckhousley:
Addresses #1029 

The Ecto section was easy enough to update, but I expanded a lot on the OTP Supervisors page.  I added a more complete tutorial using the SimpleQueue module to show that the module is automatically started.  I also tried to summarize the new DynamicSupervisor without just copying over the official docs.

Some questions:
1. The project's _config.yml file has the Elixir version set to 1.5.2.  Should that be updated to 1.6 since DynamicSupervisor was added in that version?  Or is that version kept for another reason?

2. I see that the markdown files themselves are versioned to track which files need translations.  I did not increment the version number, what should they be set to with these updates?

3. Any improvements or clarifications?


doomspork:
@chuckhousley sorry for the delays, thank you for this wonderful addition 💜

Issue #1295, opened 2018-01-06
[PT] Keep up with current state for Portuguese


pragmaticivan:
met## Básico
- [x] Básico
- [x] Coleções
- [x] Enum - @pragmaticivan
- [x] Pattern Matching - @brunoliveira8
- [x] Estruturas de Controle - @pedrosnk
- [x] Funções - @jrichocean ~~#1302~~
- [x] Operador Pipe - @jrichocean ~~#1302~~
- [ ] Composição - @pedrosnk #1300
- [x] Mix - @pedrosnk ~~#1301~~ 
- [ ] Sigils - @jrichocean
- [ ] Documentação - @jrichocean
- [ ] Testando - @renatosuero
- [x] Comprehensions - @thiamsantos 
- [ ] Strings - @brunoliveira8
- [x] Tarefas Mix Customizadas - @brunoliveira8 ~~#1321~~
- [x] IEx Helpers

## Avançado
- [ ] Interoperabilidade com Erlang - @mememori
- [x] Tratamento de Erros
- [x] Executáveis - @thiamsantos 
- [ ] Concorrência - @mememori
- [ ] Concorrência OTP - @mememori
- [ ] Supervisores OTP - @mememori
- [x] Metaprogramação - @thiamsantos
- [x] Projetos Guarda-chuva
- [x] Especificações e tipos
- [x] Comportamentos
- [x] GenStage
- [x] Protocols

## Específico
- [x] Plug
- [ ] Ecto - @mememori
- [ ] Elixir embutido (EEx) - @renatosuero
- [ ] Erlang Term Storage (ETS) - @mememori
- [x] Mnesia - @thiamsantos
- [x] Debugging

## Biblioteca
- [x] Guardian (Basics)
- [ ] Poolboy - @mememori
- [x] Benchee
- [x] Bypass

pragmaticivan:
basic - enum (https://github.com/elixirschool/elixirschool/pull/1297/files)

brunoliveira8:
basic - pattern matching (https://github.com/elixirschool/elixirschool/pull/1298)

pedrosnk:
basic - Estruturas de Controle

doomspork:
Thank you @pragmaticivan, @brunoliveira8, @pedrosnk for helping ❤️

jrichocean:
I can do basic - pipe operator & functions

doomspork:
That would be wonderful @jrichocean 😁

[...]

PR #1381, opened 2018-04-03
[pt] update umbrella projects 

thiamsantos:
ref #1295


anderkonzen:
Thanks again @thiamsantos!

PR #1385, opened 2018-04-06
[RU] Fix for broken images 

BEaStia:
**This PR is for issue(#1386).**

I've found that in English version in `debugging` section there're some screenshots:

But in Russian version they look like:

I've changed urls in Russian version according to urls in English version


doomspork:
Thank you and welcome @BEaStia!

PR #1391, opened 2018-04-20
add parens to function call

cadebward:
I'm getting this error when running the example:

```
warning: variable "phrase" does not exist and is being expanded to "phrase()", please use parentheses to remove the ambiguity or change the variable name
```

I'm pretty new to Elixir so feel free to judge :)

### DIFF
--- a/en/lessons/basics/functions.md
+++ b/en/lessons/basics/functions.md
@@ -120,7 +120,7 @@ When we don't want other modules accessing a specific function we can make the f
 
 ```elixir
 defmodule Greeter do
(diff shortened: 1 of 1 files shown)


eksperimental:
I think we should bump the PATCH version number

doomspork:
@cadebward could you please bump the patch version in the lesson meta data?  Alternatively you could apply this change across all languages in which case the version bump may not be necessary.

doomspork:
Thank you @cadebward 😀

PR #1399, opened 2018-04-24
[EN] Mention that O(n) means "linear time"

eksperimental:
/cc @elixirschool/english


doomspork:
Thank you @eksperimental 👍

PR #1398, opened 2018-04-23
[TR]  Remove the "redirect_from" directive from all.

yasinyaman:
@eksperimental , all removed.

PR #1403, opened 2018-04-25
[tr] Remove redirect_from directive

eksperimental:
It does what #1398 tried to do.
/cc @yasinyaman


doomspork:
Thank you @eksperimental 💪

PR #1362, opened 2018-03-23
Improve accessibility

nscyclone:
* links must have a `title`;
* images must have an `alt`;
* external links using `[target="_blank"]` must have `[rel="noopener"]` to deny access to `window.opener`;
* it is recommended not to limit viewport's `maximum-scale`.

<details>
<summary>Before:</summary>

before

</details>

<details>
<summary>After:</summary>

after

</details>

PR #1407, opened 2018-04-25
Reintroduce improvements to the share buttons from #1362

nscyclone:
A small subset of changes from #1362 was reverted by recent updates. This PR returns them back.


doomspork:
Thank you @nscyclone!

PR #1409, opened 2018-04-25
Add Elixir & Erlang version to sidebar

### LINKED ISSUE #1380: Add the version info in the left panel
Hi, I think adding the version info for this tutorial in the left panel is a good idea, doesn't it?

When I read this tutorial, I'm not sure which version of Elixir this tutorial teaches.

Maybe we can provide a tutorial list for the different versions of Elixir.

doomspork:
Closes #1380 

I'd like feedback on where this should go, I'll bug @burden for help styling it once we can agree on placement.

### DIFF
--- a/_includes/section-sidebar-menu.html
+++ b/_includes/section-sidebar-menu.html
@@ -5,6 +5,7 @@
     <h2>Menu</h2>
     <p>{{ locale.project.description }}</p>
   </header>
+  <span>Elixir {{site.elixir.version}} - Erlang/OTP {{site.erlang.OTP}} [erts-{{site.erlang.erts}}]</span>
   <ul>
 
   {% for sections in site.contents[language] %}


_includes/section-sidebar-menu.html
```diff
@@ -5,6 +5,7 @@
     <h2>Menu</h2>
     <p>{{ locale.project.description }}</p>
   </header>
+  <span>Elixir {{site.elixir.version}} Erlang/OTP {{site.erlang.otp}} [erts-{{site.erlang.erts}}]</span>
```
nscyclone: I don't see the OTP version on [the screenshot](https://github.com/elixirschool/elixirschool/pull/1409#issuecomment-384401062). Most likely the reason is case-sensitivity of the config keys. Shouldn't it be `{{site.erlang.OTP}}` instead?

doomspork: image

_includes/section-sidebar-menu.html
```diff
@@ -5,6 +5,7 @@
     <h2>Menu</h2>
     <p>{{ locale.project.description }}</p>
   </header>
+  <span>Elixir {{site.elixir.version}} Erlang/OTP {{site.erlang.OTP}} [erts-{{site.erlang.erts}}]</span>
```
eksperimental: We need something to separate Elixir version from Erlang/OTP other than just a space.
maybe a `-`

doomspork: image

PR #1418, opened 2018-04-26
Improve code block styles

doomspork:
image


eksperimental:
I dont like it
 Its too think. Try a weight of 40p

doomspork:
@eksperimental the current font doesn't support much in the weight of weights, it's either bold or not.  Another option is to make the font just a tad bit bigger.

image

eksperimental:
That looks better now

PR #1428, opened 2018-05-03
Translation Report

doomspork:
This is a continuation of the work @gemantzu started in #1426.  These changes pick up where he left off today (2018-05-02):

- Break the translation plugin out into it's own generator and make it pageless
- Move custom plugins into `elixir_school` directory to make namespace
- Mark missing translations as `missing`
- Move `translation_report` configuration option up to `sections`
- Reorder sidebar and fix markup
- Add `priority` to plugins for ordering

What I have left to do is add support for interlang at the top of the menu.  This isn't generated for us because this plugin comes after the other generator since we're dependent on the tree.

PR #1423, opened 2018-04-28
Custom Redirect

doomspork:
@nscyclone / @eksperimental take a peek.  Works locally for me.

Closes #1282 #1366 #1420

PR #1431, opened 2018-05-03
Rename Appendix to Blog to clear up confusion

doomspork:
I sure do love naming things but with a project like Elixir School being mindful of non-native speakers is important it appears "Appendix" isn't exactly clear to most visitors so we're going to keep the name but rename references to it "Blog".

Thoughts @elixirschool/developers ?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,10 +1,10 @@
 # Setup
 baseurl: /
 url: https://elixirschool.com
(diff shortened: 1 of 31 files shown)


doomspork:
@nscyclone I'm going to hold off on merging this until #1428 and #1423 merge if that's cool 👌

PR #1434, opened 2018-05-06
percentage shown in translation report heading

gemantzu:
One might argue on the points given, please discuss.

### DIFF
--- a/_plugins/elixir_school/translation_report.rb
+++ b/_plugins/elixir_school/translation_report.rb
@@ -93,8 +93,6 @@ module ElixirSchool
         0.7
       when 'major'
         0.5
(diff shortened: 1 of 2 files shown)


_plugins/elixir_school/translation_report.rb
```diff
@@ -41,35 +41,63 @@ def build_interlang(site, lang, locales)
...
+      case version_severity
+      when 'none'
+        1.0
+      when 'patch'
+        0.9
+      when 'minor'
+        0.7
+      when 'major'
+        0.5
+      when 'missing'
```
doomspork: You could drop off `missing` here since the default is already `0`

PR #1436, opened 2018-05-08
fix acronym

rutaka-n:
diaLyzer. not diayzer =)

### DIFF
--- a/en/lessons/specifics/debugging.md
+++ b/en/lessons/specifics/debugging.md
@@ -11,7 +11,7 @@ Bugs are an inherent part of any project, that's why we need debugging. In this 
 
 # Dialyxir and Dialyzer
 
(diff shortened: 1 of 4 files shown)


doomspork:
Thank and welcome @rutaka-n!

eksperimental:
Please bump version, and update translated versions if they share the same version, if not leave the version as is regardless the change.

doomspork:
@eksperimental why?  If the change has been applied to all the languages why bother with a version bump?

eksperimental:
It is not good practice. If you changed something, its better to update the version, even if its trivial as this. Greek version didnt get updated by the way.
What happens if somebody started translating that same version. They will never know there was a change introduced.
The one exception to this rule would be not to bumo the version is there is a stylistic change or a spelling mistake that doesn't get translated/carried over to other languages

gemantzu:
Ow I figured that it was ok in the other translations, if it is not by all means go ahead and bump the version!

PR #1450, opened 2018-05-25
[PT] Translate global definitions

thiamsantos:


_data/locales/pt.yml
```diff
@@ -23,9 +23,9 @@ version_messages:
...
-  translation_title: Translation Title
-  original_version: Original Version
-  translation_version: Translation Version
-  version_severity: Version Severity
+  section: Seção
+  original_title: Título Original
+  translation_title: Título da Tradução
+  original_version: Versão Original
+  translation_version: Versão da Tradução
+  version_severity: Gravidade da Versão
```
anderkonzen: Hi @thiamsantos , what do you think about "Severidade" instead? To me, "gravidade" resembles something alarming, in the sense of something dangerous (physically). Severity here I think it's more related to something critical, serious.

thiamsantos: Yeah, "Severidade" fits better.

Issue #487, opened 2016-07-03
Mention that GenEvent handlers runs in a single process in a single synchronous loop


heri16:
I think it is good to mention that GenEvent handlers are run sequentially in a single process/pid, and that this will eventually become a bottleneck. The better example would be to dispatch a message from the event_handler to another Process. For example, using GenServer.cast(main_handling_pid, event) for concurrency.

Additionally, If backpressure is desired to prevent overloading Handler GenServer mailboxes, then use GenEvent.ack_notify, together with GenServer.call. All of this is a stop-gap measure to avoid bad OTP patterns until GenBroker is released.

doomspork:
@heri16 would you be open to updating the lesson?

brain-geek:
Solved in #1458

heri16:
Thanks @brain-geek !

PR #1458, opened 2018-05-30
Improve GenServer definition to state that it's single process

brain-geek:
Okay, this may be discussable overall, but this may be improvement.

#487 is a source where I took the idea from, and it makes some sense.

Let's discuss if you think that it's a bad change. :)

### DIFF
--- a/en/lessons/advanced/otp-concurrency.md
+++ b/en/lessons/advanced/otp-concurrency.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.2
+version: 1.0.3
 title: OTP Concurrency
 redirect_from:
   - /lessons/advanced/otp-concurrency/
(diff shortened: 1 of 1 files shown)


pjhampton:
Slightly unrelated, but I think we should split out the OTP lessons into their own section. Talk with various ex devs it seems that OTP is pretty difficult to grasp for many

doomspork:
@pjhampton so: Basics, Advanced, Specifics, OTP?

I'm all for better organization or lessons.  It might help us see areas we should elaborate on.

pjhampton:
yeah @doomspork 

I think this because when I look around nobody (except seasoned / advanced) guys are making use of this stuff. I think it's fair to assume that OTP is harder to learn / understand and requires it's own dedicated area where people can spend time in.

Thoughts @elixirschool/developers ?

doomspork:
I could get behind that.  Think you and I could identify and tackle lessons for the other parts of OTP we haven't covered?

pjhampton:
Yeah, just to add another footnote here - I think we're good at explaining the _what_ and _how_, but not the _why_ or _when_. I think we should:

1. Have a dedicated section for OTP
2. Have a simple and a difficult example
3. Explain good use cases
4. Advise on spotting OTP opportunities

As for identifying - yeah, I am up for that. As for writing lessons, my time isn't going to free up until next weekend which sucks and I'm still finishing the ExUnit lesson :(

doomspork:
@pjhampton once #1423 is merged we can make some changes to it to support these changes (redirecting old links to the new otp section) and start on this 👍

Issue #798, opened 2016-11-14
Function lesson: Add warning that function headers are matched in order of appearance


holsky:
One of the things that cost me a lot of time early on, was that I wrote my functions matching in the wrong order, with the broadest pattern on top so that the finer ones below were never reacher. Maybe it would be a good idea to add a small word of warning to the Pattern Matching part of the functions lesson?

ybur-yug:
Can you give some examples of this? I can see it as an issue but have never run into it myself @holsky.

doomspork:
I like this idea @holsky, ordering is important.  Any suggestions on how we might address this?  Would you like to take a stab at adding a note?

holsky:
Will do, probably by end of next week!

PR #1459, opened 2018-05-30
Add pattern matching ordering note/example

brain-geek:
We have this issue from 2016 #798 which actually mentioned this part of content.

@doomspork Could you look through it as a native speaker? I think the wording is a bit off, but I can't word it the proper way. 

Thanks!


doomspork:
@brain-geek this reads fine to me 👍

Issue #586, opened 2016-08-05
Description of import and require should explain why you use one vs the other


scouten:
After reading http://elixirschool.com/lessons/basics/modules/#import and http://elixirschool.com/lessons/basics/modules/#require, I don't really understand _why_ I would choose one over the other.

scouten:
BTW, I am overall _very_ impressed with elixir-school. Please take my issues and PRs as suggestions for how to improve something that is already very effective at its mission. I'm very new to Elixir and I'm finding this to be one of the fastest ways to grok the language. :+1:

doomspork:
@scouten thank you for all of your generous feedback!  I _very_ much appreciate it, it's through involvement like your's we are able to improve the resource for everyone!

brain-geek:
I believe that #1461 addressed this issue fully.

PR #1461, opened 2018-06-01
Rewrite require definition in modules lesson

brain-geek:
As someone might have noticed, I'm going through old Issues (2016-ish, so if original author wanted to fix this they had plenty of time).

The next one is #586 . I rewrote `require` definition to be somewhat less ambiguous.

Actually, looks like `import` definition was somewhat incorrect before this change:
Official doc: `# Import functions from Foo so they can be called without the `Foo.` prefix` 
Our definition: `want to import functions and macros`

### DIFF
--- a/en/lessons/basics/modules.md
+++ b/en/lessons/basics/modules.md
@@ -1,5 +1,5 @@
 ---
-version: 1.2.0
+version: 1.3.0
 title: Modules
 ---
 
@@ -148,7 +148,7 @@ end
 
 ### `import`
 
-If we want to import functions and macros rather than aliasing the module we can use `import/`:
+If we want to import functions rather than aliasing the module we can use `import`:
 
 ```elixir
 iex> last([1, 2, 3])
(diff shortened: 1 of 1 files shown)


en/lessons/basics/modules.md
```diff
@@ -193,7 +193,7 @@ import List, only: :macros
 
 ### `require`
 
-Although used less frequently `require/2` is nonetheless important.  Requiring a module ensures that it is compiled and loaded.  This is most useful when we need to access a module's macros:
+You could use `require` to tell Elixir you're going to macros from other module. The slight difference with `import` is that it allows using macros, but not functions from the specified module:
```
nscyclone: It feels like a verb is missing before `macros` in `you're going to macros`..

doomspork: I think we want to add in "use"

brain-geek:
@doomspork @nscyclone Thanks, guys. Applied your suggestions.

PR #1463, opened 2018-06-20
A look at Elixir configuration

doomspork:
A new blog post that seeks to clear up some confusion around configuration in Elixir.

Feedback — I want it!

### DIFF
--- a/_posts/2018-07-17-configuration-demystified.md
+++ b/_posts/2018-07-17-configuration-demystified.md
@@ -1,0 +1,162 @@
+---
+author: Sean Callan
+author_link: https://github.com/doomspork
+categories: general
+date: 2018-07-17
+ayout: post
+title: Configuration Demystified
+excerpt: >
+  We attempt to clear up some confusion around configuration by looking at the different types, the roles they play, and a different approach we could take.
+---
+
+There's been a lot of discussion about configuration in the community lately.
+We thought this would be an opportune time to discuss configuration and how best to handle it within an Elixir application.
+It is surprising to see how a small change to our applications configuration can eliminate much of the headaches others are experiencing.
+
+### Configuration types
+
+Before we go any further, let's look at the two configuration types and the roles they play.
+
+__Runtime Configuration__
+
+If you've ever used a system environment variable to configure some part of your application, then you're familiar with runtime configuration.
+As the name suggests, this is the configuration for an application at the time it is run.
+These are the values we can expect to change as we deploy our build artifacts to different systems.
+
+__Build-time Configuration__
+
+Our build-time configuration, sometimes known as Application configuration, is something different and the difference, while subtle, can be a pitfall in certain situations.
+The difference shines when we consider that our code, and its configuration, is compiled into a build artifact we can distribute.
+
+We can say with certainty that no matter where our application is run there are certain things we want to remain constant; we intend to use the same `Logger` configuration regardless of where we deploy.
+If we're relying on dependency injection for testing, then we know for certain we don't want to use those dependencies in the final deliverable.
+They configure the function of our code.
+
+### How it's been done
+
+The frustration for many people has to do with the usage of `Application.get_env/2` and `System.get_env/1`.
+We'll jump right in by looking at a configuration that's common to many Elixir projects:
+
+```elixir
+use Mix.Config
+
+config :example_app, Data.Repo,
+  adapter: Ecto.Adapters.MySQL,
+  username: System.get_env("EXAMPLE_APP_USERNAME"),
+  password: System.get_env("EXAMPLE_APP_PASSWORD"),
+  hostname: System.get_env("EXAMPLE_APP_HOSTNAME"),
+  database: System.get_env("EXAMPLE_APP_DATABASE"),
+  pool_size: 10
+```
+
+Simple and harmless enough, right?
+Wrong!
+
+Our application's configuration, defined in `config.exs` and friends, is compiled when we generate build artifacts, like those produced by Distillery.
+That means those `System.get_env/1` functions need to be resolved at compile time.
+See the problem?
+Our application's compiled code is coupled to the configuration of the system where it was compiled.
+
+What if we want to generate the build artifact locally and run it elsewhere?
+What if there's an emergency and the value of `EXAMPLE_APP_HOSTNAME` has been updated?
+With this configuration our application needs to be recompiled for changes to take effect.
+
+Let's illustrate the concept using colors to differentiate changes:
+
+![elixir-config-recompile](https://user-images.githubusercontent.com/73386/41503026-d8a66ef4-7185-11e8-95fa-37598f6a56ff.png)
+
+Here we see that our runtime values are different, which required us to recompile our code.
+This results in a new build artifact and updated configuration for Runtime B.
+We've managed to couple the runtime configuration and code together.
+To see changes in our environment reflected in our code recompilation is unavoidable.
+For those using releases this combination of configuration types often times requires additional libraries to fill the gaps.
+
+At its root, the problem is the conflation of two separate concepts: build-time and runtime configuration.
+
+### A new approach
+
+Now let's look at another approach to configuration that enables changes to the runtime, without requiring our code to be recompiled:
+
+```elixir
+defmodule ExampleApp.Repo do
+  use Ecto.Repo, otp_app: :data
+
+  def init(_, opts) do
+    {:ok, build_opts(opts)}
+  end
+
+  defp build_opts(opts) do
+    system_opts = [
+      database: System.get_env("EXAMPLE_APP_DATABASE"),
+      hostname: System.get_env("EXAMPLE_APP_HOSTNAME"),
+      password: System.get_env("EXAMPLE_APP_PASSWORD"),
+      username: System.get_env("EXAMPLE_APP_USERNAME")
+    ]
+
+    Keyword.merge(opts, system_opts)
+  end
+end
+```
+
+Here `Repo.init/2`, called on start, is used to update our configuration with values from the current system environment without the need to recompile anything.
+
+Now we could generate our build artifacts locally and run them elsewhere.
+How would the aforementioned scenario with `EXAMPLE_APP_HOSTNAME` changes play out?
+An application restart would pull the latest value, no compilation necessary.
+
+Let's update our illustration from before to reflect this new approach:
+
+![elixir-config](https://user-images.githubusercontent.com/73386/41503027-d8b8ecc8-7185-11e8-8284-73d417fea6dc.png)
+
+Our runtime environments have changed but our application's configuration and build artifacts have not, nor should they need to.
+
+We've managed to decouple our code from our runtime configuration with the added bonus of a configuration that is explicit and lives alongside the code.
+
+### Bringing it together
+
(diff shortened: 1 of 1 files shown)


_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
+---
+author: Sean Callan
+author_link: https://github.com/doomspork
+categories: general
+date: 2018-06-18
+layout: post
+title:  Confusing Configuration
```
doomspork: Probably needs a better title.

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
+---
+author: Sean Callan
+author_link: https://github.com/doomspork
+categories: general
```
doomspork: Should this be classified under something other than `general`?  What categories should we have?

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
...
+As the name suggests, this is the configuration for an application at the time it is run.
+These are the values we can expect to change as we deploy our build artifacts to different systems.
+
+__Application Configuration__
+
+Application configuration is something different, and the difference, while subtle, can be a pitfall in certain situations.
+The difference shines when we consider that our code, and it's configuration, is compiled into a build artifact we can distribute.
+
+We can say with certainty that no matter where our application is run there are certain things we want to remain constant.
+No matter where we deploy our application, we want know we want to use the same `Logger` configuration.
```
anderkonzen: I think there is a typo in "... we want know we want to use ..."

_posts/2018-06-18-configuration.md
burden: We're not using the `categories` feature yet (or `tags`). At a later date we could perhaps use it to group similar posts?

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
...
+date: 2018-06-18
+layout: post
+title:  Confusing Configuration
+excerpt: >
+  We attempt to clear up some confusion around configuration by looking at the different types, the roles they play, and a different approach we could take.
+---
+
+There's been a lot of discussion about configuration in the community lately.
+We thought this would be an opportune time to discuss configuration and how best to handle it within an Elixir application.
+It is surprising to see how a small change to our applications configuration can eliminate much of the headaches others are experiencing._
```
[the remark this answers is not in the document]
doomspork: Typo, good catch!

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
...
+__Runtime Configuration__
+
+If you've ever used a system environment variable to configure some part of your application, then you're familiar with runtime configuration.
+As the name suggests, this is the configuration for an application at the time it is run.
+These are the values we can expect to change as we deploy our build artifacts to different systems.
+
+__Application Configuration__
+
+Application configuration is something different, and the difference, while subtle, can be a pitfall in certain situations.
+The difference shines when we consider that our code, and it's configuration, is compiled into a build artifact we can distribute.
```
michaelstalker: Nitpick: "it's configuration" should be "its configuration"

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
...
+  pool_size: 10
+```
+
+Simple and harmless enough, right?
+Wrong!
+
+Our application's configuration, defined in `config.exs` and friends, is compiled when we generate build artifacts, like those produced by Distillery.
+That means those `System.get_env/1` functions need to be resolved at compile time.
+See the problem?
+Our application's code is coupled to our system configuration.
```
michaelstalker: Maybe it would be clearer to say that our application's _compiled_ code is coupled to the configuration of the system that compiled the code. I don't want to split hairs, but the use of `System.get_env/1` will always make the _behavior_ of the application dependent upon some system configuration, and I'd bet some people see behavior and code as synonymous.

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
...
+
+Let's illustrate the concept using colors to differeniate changes:
+
+![elixir-config-recompile](https://user-images.githubusercontent.com/73386/41503026-d8a66ef4-7185-11e8-95fa-37598f6a56ff.png)
+
+Here we see that our runtimes values are different, which required we recompile our code.
+This results in a new build artifact and updated configuration for Runtime B.
+We've managed to couple the runtime configuration and code together.
+To see changes in our environment reflected in our code recompilation is unavoidable.
+For those using Releases this combobulation of configuration types often times requires additional libraries to fill the gaps.
```
michaelstalker: Did you mean to capitalize "Releases"?

Will "combobulation" make it difficult to translate this article into other languages, since the word is slang? If so, "combination" may be a better fit.

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
...
+
+At it's root, the problem is the conflation of two separate concepts: application configuration and runtime configuration.
```
michaelstalker: ~it's~ its

I'm not sure if the distinction between application and runtime configuration is the right one. I might draw the difference between build-time configuration and runtime configuration instead, due to some ambiguity in what "application" means. Just my two cents 😄

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
...
+If you've ever used a system environment variable to configure some part of your application, then you're familiar with runtime configuration.
+As the name suggests, this is the configuration for an application at the time it is run.
+These are the values we can expect to change as we deploy our build artifacts to different systems.
+
+__Application Configuration__
+
+Application configuration is something different, and the difference, while subtle, can be a pitfall in certain situations.
+The difference shines when we consider that our code, and it's configuration, is compiled into a build artifact we can distribute.
+
+We can say with certainty that no matter where our application is run there are certain things we want to remain constant.
```
michaelstalker: > where our application is run

You mean a particular, compiled instance of an application, right? I don't think you're making a statement about how the application will run, no matter what the build-time configuration is. "Application" in this article seems to refer to the uncompiled code sometimes, and to a particular instance of compiled code at other times. Since you're assigning a technical meaning to "application configuration," it may be worth assigning a technical meaning to "application," defining it, and using it in only one sense of the word.

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,160 @@
...
+Wrong!
+
+Our application's configuration, defined in `config.exs` and friends, is compiled when we generate build artifacts, like those produced by Distillery.
+That means those `System.get_env/1` functions need to be resolved at compile time.
+See the problem?
+Our application's code is coupled to our system configuration.
+
+What if we want to generate the build artifact locally and run it elsewhere?
+What if there's an emergency and the value of `EXAMPLE_APP_HOSTNAME` has been updated?
+With this configuration our application needs to be precompile for changes to take effect.
```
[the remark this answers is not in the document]
dgmcguire: I think this should be `to be recompiled` no?  Precompilation implies it was compiled before the updated ENV was setup.

_posts/2018-06-18-configuration.md
doomspork: This is a very valid point, I will change it to "combination"

doomspork:
@nscyclone / @michaelstalker I believe I have incorporated all of your feedback.  Want to give this one last once over before I update the published date and merge?

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,162 @@
...
+Our application's configuration, defined in `config.exs` and friends, is compiled when we generate build artifacts, like those produced by Distillery.
+That means those `System.get_env/1` functions need to be resolved at compile time.
+See the problem?
+Our application's compiled code is coupled to the configuration of the system where it was compiled.
+
+What if we want to generate the build artifact locally and run it elsewhere?
+What if there's an emergency and the value of `EXAMPLE_APP_HOSTNAME` has been updated?
+With this configuration our application needs to be recompile for changes to take effect.
+
+Let's illustrate the concept using colors to differeniate changes:
```
burden: > using colors to *differentiate* changes

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,162 @@
...
+  end
+end
+```
+
+When our application starts it will attempt to retrieve those system variables, removing the `nil` values, and finally merging the options defined by our application configuration with the runtime configuration, giving precedence to the runtime options.
+
+Now we can use the `dev.exs` and `test.exs` files we're so accustomed to while also ensuring our final build artifact will be correctly set up, thus making configuration of deployments a breeze.
+
+What do you think of this approach?
+We'd love to hear your thoughts!
```
burden: Maybe suggest where those thoughts are best directed-- issues, twitter, other...?

_posts/2018-06-18-configuration.md
burden: How about something along the lines of "Configuration Demystified" or even "Best Practices"?

_posts/2018-06-18-configuration.md
```diff
@@ -0,0 +1,162 @@
...
+__Runtime Configuration__
+
+If you've ever used a system environment variable to configure some part of your application, then you're familiar with runtime configuration.
+As the name suggests, this is the configuration for an application at the time it is run.
+These are the values we can expect to change as we deploy our build artifacts to different systems.
+
+__Build-time Configuration__
+
+Build-time, or Application, configuration is something different, and the difference, while subtle, can be a pitfall in certain situations.
+The difference shines when we consider that our code, and its configuration, is compiled into a build artifact we can distribute.
```
burden: I'm going to suggest another line here to make it a paragraph. Markdown is running these two sentences together making it pretty confusing.

_posts/2018-06-18-configuration.md
doomspork: We need to figure this out TBH.

_posts/2018-07-12-configuration-demystified.md
```diff
@@ -0,0 +1,163 @@
...
+
+__Runtime Configuration__
+
+If you've ever used a system environment variable to configure some part of your application, then you're familiar with runtime configuration.
+As the name suggests, this is the configuration for an application at the time it is run.
+These are the values we can expect to change as we deploy our build artifacts to different systems.
+
+__Build-time Configuration__
+
+Build-time, or Application, configuration is something different, and the difference, while subtle, can be a pitfall in certain situations.
```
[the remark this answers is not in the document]
michaelstalker: I believe the commas are correct as they're written. "Or Application" is a somewhat parenthetical explanation of "Build-time," and the clause needs commas to set it apart. Both build-time and application are adjectives for configuration. "Build-time, or Application configuration" makes it sound like "Build-time" is one thing, and "Application configuration" is something else.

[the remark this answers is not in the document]
doomspork: > simplifying the sentence structure would make it easier to read.

I suspect the phrase does not translate well (@gemantzu is Greek).  I'll fix that tonight so we can get this merged.

doomspork:
Thank you @burden, @michaelstalker, and @gemantzu 🎉

PR #1464, opened 2018-06-20
Cleans up the blog title url

gemantzu:


doomspork:
Any thoughts on this change @elixirschool/developers / @burden?  I'd like to make it before we have a lot of blogs and have to worry about redirecting.

PR #1471, opened 2018-07-20
Fix instructions on path to save verify_request.ex

gjastrab:
Path was missing top level namespace folder of the application.


gjastrab:
Certainly, a few of the versions were only `0.9.x` so I bumped those to `0.10.0`

doomspork:
Thank you @gjastrab 👍

gjastrab:
No sweat, thanks for your work on elixir school, I've found it a great resource!

doomspork:
You're very welcome @gjastrab!  We're always open to feedback on how to make it even better 😁

PR #1473, opened 2018-07-24
last_commit as a link on report

gemantzu:
This is based on the fact that english lesson has the version on second line. For example, en/index.md breaks it.


eksperimental:
Yes, in English.
Thanks

PR #1484, opened 2018-08-19
Update earmark and ex_doc package versions

adamfeldman:
The package versions used in the lesson result in errors when running `mix docs`

### DIFF
--- a/en/lessons/basics/documentation.md
+++ b/en/lessons/basics/documentation.md
@@ -181,8 +181,8 @@ Assuming all is well and we're seeing the output above, we are now ready to set 
 
 ```elixir
   def deps do
-    [{:earmark, "~> 1.2.5", only: :dev},
-    {:ex_doc, "~> 0.19.1", only: :dev}]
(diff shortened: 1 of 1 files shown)


en/lessons/basics/documentation.md
```diff
@@ -181,8 +181,8 @@ Assuming all is well and we're seeing the output above, we are now ready to set
 
 ```elixir
   def deps do
-    [{:earmark, "~> 0.1", only: :dev},
-    {:ex_doc, "~> 0.11", only: :dev}]
+    [{:earmark, "~> 1.2.5", only: :dev},
```
doomspork: Should we match these on PATCH or MINOR (`~> 1.2.5` vs `~> 1.2`)?

gemantzu: Minor makes more sense for a guide, so we won't have to update that often.

doomspork:
Thank you @adamfeldman and @gemantzu!

PR #1485, opened 2018-08-20
Simplified Chinese translation updates

zillou:


doomspork:
Thank you and welcome @zillou!!

Issue #1481, opened 2018-08-09
Ecto cast/4 misused


Hugo-Hache:
The Ecto lesson tells that 

> we use cast/4 to convert our parameters to a changeset from a set of required and optional fields

 and in examples uses it like this 

`|> cast(params, @required_fields, @optional_fields)`

However ecto/4 [documentation](https://hexdocs.pm/ecto/Ecto.Changeset.html#cast/4) refers the third parameter as permitted and the last as options. So I suppose the lesson should use it like this:

```
|> cast(params, @required_fields ++ @optional_fields)
|> validate_required(@required_fields)
```

doomspork:
@Hugo-Hache good catch, Ecto has changed since the lesson was written.  Would you be open to submitting a PR to update it?

PR #1487, opened 2018-08-21
Update Ecto lesson with cast/3

Hugo-Hache:
As mentioned in #1481, Ecto cast arity changes in recent releases. 

I updated the examples and the description of the process.


doomspork:
Thank you so much @Hugo-Hache 😁

Issue #1432, opened 2018-05-03
New Jekyll plugin to generate translation boilerplate


doomspork:
https://jekyllrb.com/docs/plugins/#commands

```ruby
module ElixirSchool
  class NewTranslation < Jekyll::Command
  end
end
```

PR #1488, opened 2018-08-24
Merge interlang into locales

nscyclone:
- [x] `cn` and `tw` were renamed to corresponding ISO language codes in `interlang.yml` and `contributors.yml`
- [x] Language names were moved from `interlang.yml` into corresponding `locales` files
- [x] `README` section about `interlang` and code relying on `interlang.yml` were updated accordingly


eksperimental:
Im not too sure about this change. When a new language is added, the translator will have to modified every locale file for the other languages to include the new entry. This is foing to

eksperimental:
Im not too sure about this change. When a new language is added, the translator will have to modified every locale file for the other languages to include the new entry. This is going to lead to peole missing some, it will be hard to track and language names are not going to be translated

gemantzu:
This is a good concern. To be honest, now that you mentioned it, both solutions seem to me a bit bad. Are we looking at this the wrong way? Do we actually need this? I mean, to me it just adds noise to someone who wants to get a translation working, it feels like a chore.

eksperimental:
Its not noise, its the chance for prople who dont speak English to navigate to  other translated versions of the guide if they can read the language

eksperimental:
My main concern this this guide and its translations is that you still need to be able to understand English to read it, which ofcourse defeats the purpose of translating it

eksperimental:
Also, to comment that IIRC I was one who introduced this feature, and I did it intentionally all in one file for the reasons stated above.

gemantzu:
What I described as noise, is the fact that every translator has to translate all other language names in their respective language, plus his language name in all other languages. I don't see why this should be the case. Each translation should have their translated local name, which makes sense for translators and visitors and probably nothing else. Why should someone care how their language name is translated in whatever other language? And besides that, how can someone be sure that whatever translator he used for it, returned the correct result? Unless we want the other language maintainers to add that language to their respective field to make sure that it is correct.

I might be wrong here, and I hope that I did not offend your work in any way, but my understanding is that since we want ppl to start contributing more, we should make their lives easier, not harder. 

Get in the shoes of a newcomer for a minute here. You find this project and you think you should contribute to it. You make the folders / files needed, you translate the locales, and then you stumble upon interlang. You add all the other languages translated in yours, and then you start adding your language to the others. Since you might not know how your language is spelled in Vietnamese or Indonesian or traditional Chinese, google translate to the rescue. How possible is it that you will not get bored out of your mind midway filling this file? I know I would be. I am not a role model regarding patience though, so I might still be wrong. Also, add the feeling of uneasiness that comes with using a translation tool _for a language you know nothing about_.

eksperimental:
To make it clear, interlang is about translating every language name into your language, and not your language name into every other language. Hence the confusion

doomspork:
I love that there's a discussion going on here ❤️ 

I've read this over and now I'm wondering something:  do we need interlang?  

> You add all the other languages translated in yours, and then you start adding your language to the others. Since you might not know how your language is spelled in Vietnamese or Indonesian or traditional Chinese, google translate to the rescue. 

This got me thinking.  Do we need to know what all the other languages are in Russian for instance? 
 Or would it be simpler to have 1 list of all available languages but with their name translated into their native language?

```yml
interlang:
  en: English
  es: Español
  gr: Ελληνικά
  id: Bahasa Indonesia
  ja: 日本語
  th: ภาษาไทย
  ru: Русский
```

gemantzu:
This is exactly what I was saying @doomspork, I think me not being native English speaker just got in the way and made things more complicated.

doomspork:
Which part @eksperimental?

This portion is what I had in mind: 

image

This would be achieved with that singular file, right?  These appear to be the language in it's native language/script: "Čeština Euskara فارسی 한국어 Magyar Norsk Română Srpski Srpskohrvatski"

doomspork:
The "More than 1,000,000 articles" part makes me think of some changes we could make to the homepage to quickly highlight the translation completions... 🤔

eksperimental:
@doomsport the interlang sidebar on the left

nscyclone:
Wow, I am glad to see such an interesting discussion here! Sorry for coming late to the party again.

To be honest I don't understand what exactly went wrong with this PR. The problem that is currently being discussed wasn't actually brought by these changes. Let me elaborate this. The only thing I did to interlang is moving it from a separate file to the corresponding locale. This way allows us to store all the site-related translation in a single place. Before this change we still had to modify the English interlang file when we were adding a new language `xx`:

1. add `locales/xx.yml`;
2. add a section with `xx` language to `interlang.yml`;
3. add `xx` language into `en` section of `interlang.yml`.

The third step just wasn't documented anywhere, so it might feel that I've added a brand new additional configuration step but I really did not.

@doomspork's idea inspired by Wiki seems like a great compromise for me too. I also think that we could talk more about #1432, because that might make the whole process easier and more straight-forward.

eksperimental:
@nscyclone sorry, you are totally right, I got it wrong since it's not the duty of the new language to add the entry in every language but to only add it to the English language.
I think we are good as it is then. the only thing that needs to be implemented is add the version for the locale files so translators can be notified of the new changes.

Issue #711, opened 2016-10-08
Mnesia guide extensions


doomspork:
PR #697 (thanks @brunogirin!) adds additional content to the mnesia lesson, we'll want to update the translations.
## Translation

Translation progress:
- [x] en
- [x]  ru /cc @brain-geek / @nscyclone
- [x]  ko /cc @devleoper
- [ ]  jp /cc @3100
- [ ]  no /cc @therobinsj
- [x]  pl /cc @Koziolek
- [ ]  sk /cc @michalvalasek
- [ ]  fr /cc @julienXX
- [ ]  cn /cc @cizixs
- [ ]  es /cc @aleandros / @jonahoffline / @eze1981
- [ ]  pt /cc @pragmaticivan
- [ ]  vi /cc @joneslee85
- [ ]  it /cc @lucidstack  / @sformisano
- [ ]  my /cc @alkaha
- [ ]  id /cc @rezaprima

anderkonzen:
I'll take the PT translation for this one

PR #1494, opened 2018-08-28
[zh-hans] Mnesia lesson translation

zillou:
Chinese translation for #711


doomspork:
Thank you thank you @zillou!

PR #1495, opened 2018-08-31
update ecto lesson

gritsenkob:
add 'a' to end of 
@required_fields ~w(username encrypted_password email)
and
@optional_fields ~w()
lines. Without it validate_required(@required_fields) throw error: " expects field names to be atoms, got: " string name of field.

### DIFF
--- a/en/lessons/specifics/ecto.md
+++ b/en/lessons/specifics/ecto.md
@@ -142,8 +142,8 @@ defmodule ExampleApp.User do
     timestamps
   end
 
-  @required_fields ~w(username encrypted_password email)
-  @optional_fields ~w()
+  @required_fields ~w(username encrypted_password email)a
+  @optional_fields ~w()a
 
   def changeset(user, params \\ :empty) do
     user


en/lessons/specifics/ecto.md
```diff
@@ -142,8 +142,8 @@ defmodule ExampleApp.User do
     timestamps
   end
 
-  @required_fields ~w(username encrypted_password email)
-  @optional_fields ~w()
+  @required_fields ~w(username encrypted_password email)a
+  @optional_fields ~w()a
```
doomspork: What do you think about dropping the optional_fields all together since this is empty?  If you wouldn't mind, could you make this change:

Changing these lines: 
```elixir
  @required_fields ~w(username email password password_confirmation)
  @optional_fields ~w()

  def changeset(user, params \\ :empty) do
    user
    |> cast(params, @required_fields ++ @optional_fields)
```

To this:
```elixir
  @required_fields ~w(username email password password_confirmation)

  def changeset(user, params) do
    user
    |> cast(params, @required_fields)
```

Removing the optional_fields and the default `:empty` params.  Neither of those are used in the lesson and I think add to the confusion.

zillou: @doomspork One thing @gritsenkob wants to fix here is the `required_fields` in the example is a list of strings, but Ecto expects the field name to be atoms.

I suggest to change the `@required_fields` literal to the most straightforward way: `[:username, :encrypted_password, :email]`, the sigils `~w` introduces an unnecessary overhead to beginners.

doomspork: @zillou that is a good idea, `~w()` and `~w()a` can be confusing!

doomspork: @gritsenkob could you make this final change?  I'd love to merge your PR 😁

doomspork:
Thank you again @gritsenkob!  I've made the changes to the module attributes directly on master 👍

PR #1496, opened 2018-09-01
[zh-hans] Update typespec Chinese translation

zillou:


doomspork:
Thank you, thank you, thank you @zillou!

PR #1498, opened 2018-09-02
[zh-hans] Minor version updates

zillou:


doomspork:
Thank you thank you @zillou 😁

PR #1500, opened 2018-09-13
Fix contributor's alias  (@brain-geek)

nscyclone:


doomspork:
Good catch @nscyclone 👍

PR #1513, opened 2018-09-24
fix plug lesson, about the router

t-kusakabe:
> 手始めに、lib/plug/router.exというファイルを作り、以下をコピーしましょう:

When making router.ex, there is an instruction as above.

But originally should be /lib/example/router.ex

Otherwise there is a discrepancy with the description in lib / example.ex.

Please use this if you like.


doomspork:
Thank you very much @t-kusakabe ❤️

riseshia:
Thanks!!

PR #1525, opened 2018-10-02
Fix blog post

notactuallypagemcconnell:


notactuallypagemcconnell:
Not sure why this shows the whole thing being different and not just the edit adding the git clone line at the beginning.

gemantzu:
Thank you @notactuallypagemcconnell

Issue #1519, opened 2018-09-30
New Ecto Section & Lessons


doomspork:
After many requests we've decided to proceed with a section dedicated to Ecto, going above and beyond our existing lessons.

Here are the current lessons and subsections we've come up with, these are fluid and open to discussion.  Ideally each lessons equates to a single PR so others may participate.

_New!_ __Ecto Section__

- [x] Basics (lesson)
  - What is Ecto?
  - What isn’t Ecto?
  - Overview of adapters
  - Working with Ecto
  - Migrations
  - Schemas
- [x] Associations
  - Belongs to
  - Has many
  - Many to Many
  - Saving associations
  - Retrieving associated data
    - Preload etc 
- [x] Changesets
  - Basic changeset usage
  - Look at validation
  - Custom validations
  - Changeset design approaches
    - `changeset/2` vs `create_changeset/1` & `update_changeset/2`
- [x] Querying Data
  - Essentials of from/belongs to/where
  - Changesets and not having to touch the database
  - DSL vs functions
  - SQL functions (basics, a la count)
  - More in-depth association querying
- [ ] Advanced Querying
  - Complex joins
  - Fragments
  - Ecto.Multi
- [ ] Query Composition
  - Piped querying
  - Functional approach
  - Comparison

gj:
@doomspork to start, I'll split the Querying Data / Advanced Querying lessons with @notactuallypagemcconnell

jeremyjung:
I can take the basics section.

svileng:
I suspect it won't be much work for me to re-se my [short article on Ecto.Multi](https://medium.com/heresy-dev/a-brief-guide-to-ecto-multi-9c8ea0c729f0), just need to re-read it once again, make any changes if necessary and open a PR. Feel free to ping me if you don't see a PR by end of next week :)

A suggestion for the `Associations` chapter: working with embedded schemas. I remember I struggled a bit when I started using embedded schemas, not just setting them up, but also when updating & replacing `embeds_many` relationships. If no one picks that up, I can definitely write something up, but I'm a bit short on time, so not sure when that would happen 😞

doomspork:
@jeremyjung I think we can re-use some of our existing lesson and then delete it once yours is ready.  The new Ecto section will replace the single Ecto lesson by expanding on the content.

[...]

PR #1539, opened 2018-10-07
[WIP] Add Ecto Basics Lesson

jeremyjung:
Initial version of ecto basics lesson based off of the official ecto getting started guide.  Seeking feedback so I better understand the type of content that's expected for this guide.  Related to (#1519)

### DIFF
--- a/en/lessons/specifics/ecto_basics.md
+++ b/en/lessons/specifics/ecto_basics.md
@@ -1,0 +1,196 @@
+---
+version: 1.3.0
+title: Ecto
+---
+
+Ecto allows you to query and store data into databases using its own query language.  It is typically used with relational databases such as PostgreSQL.
+
+If you've worked with an Object Relational Mapper (ORM) like ActiveRecord, Entity Framework, or Hibernate many of the concepts in Ecto will be similar.
+
+{% include toc.html %}
+
+### Adapters
+
+Ecto supports different databases through the use of adapters.  A few examples of adapters are:
+
+* PostgreSQL
+* MySQL
+* SQLite
+
+In this tutorial, we'll configure Ecto to use a PostgresQL adapter.
+
+### Working with Ecto
+
+We'll cover 3 parts of Ecto in this tutorial:
+
+* Repository
+    * Connects to the database
+* Migration
+    * Describes how to create or update tables in the database
+* Schema
+    * Maps information in database tables to structs
+
+To start, create an application with a supervision tree.  
+```shell
+$ mix new friends --sup
+$ cd friends
+```
+
+This creates an elixir application named `Friends`.
+
+Add the ecto and postgrex package dependencies to your `mix.exs` file.
+
+```elixir
+  defp deps do
+    [
+      {:ecto, "~> 2.0"},
+      {:postgrex, "~> 0.11"}
+    ]
+  end
+```
+
+Fetch the dependencies using
+
+```shell
+$ mix deps.get
+```
+
+#### Creating a Repository
+
+A repository in Ecto maps to a datastore such as our Postgres database.  All communication to the database will be done using this repository.
+
+Set up a repository by running:
+```shell
+$ mix ecto.gen.repo -r Friends.Repo
+```
+
+This will generate the configuration required in `config/config.exs` to connect to a database including the adapter to use.  This is the configuration file for our `Friends` application
+
+```elixir
+config :friends, Friends.Repo,
+  adapter: Ecto.Adapters.Postgres,
+  database: "friends_repo",
+  username: "postgres",
+  password: "",
+  hostname: "localhost"
+```
+
+This configures how Ecto will connect to the database.  Note how we chose the `Ecto.Adapters.Postgres` adapter.
+
+It also creates a `Friends.Repo` module inside `lib/friends/repo.ex`
+
+```elixir
+defmodule Friends.Repo do
+  use Ecto.Repo, otp_app: :friends
+end
+```
+
+We'll use the `Friends.Repo` module to query the database. We also tell this module to find its database configuration information in the `:friends` Elixir application.
+
+Next, we'll setup the `Friends.Repo` as a supervisor within our application's supervision tree in `lib/friends/application.ex`.  This will start the Ecto process when our application starts.
+
+```elixir
+  def start(_type, _args) do
+    # List all child processes to be supervised
+    children = [
+      Friends.Repo,
+    ]
+
+  ...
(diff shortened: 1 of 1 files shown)


notactuallypagemcconnell:
Will give this a review today after work @jeremyjung! Thank you!

notactuallypagemcconnell:
@jeremyjung one immediate note -- 
We are making a move to make a whole section on Ecto itself rather than make all these new pieces be part of Ecto's base lesson as it exists. So we want to actually make this a new subsection `Ecto` that has sections per each of the items identified in [the original issue](https://github.com/elixirschool/elixirschool/issues/1519).

This would essentially entail creating an `Ecto` directory inside of `specifics` that then includes a top level lesson on `01_basics.md` (01 so that it logically shows first -- CC @doomspork to make sure thats the structure we are shooting for).

Going to look at the diff and give some commentary if I have any now! Thanks again for contributing ❤️

en/lessons/specifics/ecto_basics.md
```diff
@@ -0,0 +1,196 @@
+---
+version: 1.3.0
+title: Ecto
+---
+
+Ecto allows you to query and store data into databases using its own query language.  It is typically used with relational databases such as PostgreSQL.
+
+If you've worked with an Object Relational Mapper (ORM) like ActiveRecord, Entity Framework, or Hibernate many of the concepts in Ecto will be similar.
```
notactuallypagemcconnell: I would disagree with this personally. I think the language of the existing lesson is more fitting

> Ecto is an official Elixir project providing a database wrapper and integrated query language. With Ecto we’re able to create migrations, define schemas, insert and update records, and query them. (it doesnt have a section I can hyperlink to but scroll up just a bit [here](https://github.com/elixirschool/elixirschool/blob/master/en/lessons/specifics/ecto.md#setup) to see the original.

notactuallypagemcconnell: To elaborate on my disagreement: this follows a repository pattern while AR (what I believe most people who come to us with the ruby-heavy background of Elixir learners in a lot of cases) disregards it in favor of abstraction that reads a little more nicely but has a lot of magic. I do not know the others mentioned, though.

en/lessons/specifics/ecto_basics.md
```diff
@@ -0,0 +1,196 @@
...
+### Schemas
+
+Schemas define how data from a database table or view maps to an elixir struct.  Tables are created using the plural form of the word but schemas should use the singular form.
+
+We'll create a *person* schema to match our people table.  This way, when we query our database, we can receive a collection of *person* elixir structs back.
+
+Let's create the schema at `lib/friends/person.ex`:
+
+```elixir
+defmodule Friends.Person do
```
notactuallypagemcconnell: A lot of the time the naming pattern of modules matches the database table. There are times that this doesnt apply, lets say we had Teachers and Students, but all of them were `users`, but this goes into more complex use cases and relationships. Maybe make them match since this is the basics section? I personally think that makes the introduction here more straightforward. But I dont feel strongly, this is just observation from my end.

en/lessons/specifics/ecto_basics.md
jeremyjung: I agree with you that they are not the same.  However, I came at it from more of a high level perspective.  If we take Ecto and AR as an example-- they both:
* Provide a way to modify and create database schemas using migrations
* Provide a query language that will ultimately be translated to the database's native query language based on the adapter used.
* Take the result of the query and maps it to a native data type (object in AR, struct in ecto)

I'm fine with putting the existing lesson's explanation back though.

en/lessons/specifics/ecto_basics.md
jeremyjung: To make sure I understand correctly, are you suggesting that `defmodule Friends.Person` would be `defmodule Friends.People` and file would be `people.ex` in this case?

jeremyjung:
Thanks for the feedback @notactuallypagemcconnell !

Understood regarding the subsections comment.  I would like some more general feedback regarding the structure of the lesson.  Here are some questions I have.
* Is the level of detail right or is it too simplistic? Too verbose?
* Will the series of lessons be standalone or will they all build on the same application?
* Should the lesson guide the user step by step through building an application or just serve as a reference for the concepts of ecto?

notactuallypagemcconnell:
> Is the level of detail right or is it too simplistic? Too verbose?

We try to be concise, but informative. I think this is walking right in line with that.

> Will the series of lessons be standalone or will they all build on the same application?

I don't know if we have a final call for that. My Guy says standalone because people often come to us as a resource for quick reference. CC @doomspork on this one 

> Should the lesson guide the user step by step through building an application or just serve as a reference for the concepts of ecto?

Here I would fall on the side of having the steps (if we are keeping the lessons isolated and standalone) but not necessarily going into explaining the boilerplate, if that makes sense to you. There are other parts of ES that can guide them through the subtleties of mix and why we might user a supervised app etc.

en/lessons/specifics/ecto_basics.md
notactuallypagemcconnell: I would have this be `defmodule Friends.Person` with a filename `lib/friends/person.ex` with a table name `people` in the schema definition bit. The Schema is conventionally singular to avoid complexities of pluralization with ecto. Person vs People is an example of this.

doomspork:
@jeremyjung and @notactuallypagemcconnell apologizes for the delays 😰

I would agree with @notactuallypagemcconnell's comments.  Since the different Ecto lessons are being done by different folks, I think separate code for now makes the most sense.  Once we've finalized all the lessons I'm 💯on board coming back through and looking for a way to unify them through a single project.

We don't do many project continuations for the reasons @notactuallypagemcconnell mentioned but I think as the lessons mature and evolve we can find ways to incorporate that.

en/lessons/specifics/ecto_basics.md
```diff
@@ -0,0 +1,194 @@
+---
+version: 1.3.0
```
doomspork: @jeremyjung / @notactuallypagemcconnell since this is an entirely new lesson (even though a take on the original), what do you think about starting over at `1.0.0`?

cc @gemantzu / @brain-geek / @nscyclone for comment from the translator perspective.

doomspork: @gemantzu I believe the plan is to move this into the ecto subsection as the basics.

@notactuallypagemcconnell / @SophieDeBenedetto correct me if I'm mistaken.

PR #1543, opened 2018-10-11
add localized homepage links

thth:
1. Changed "Elixir School" in the header to link to the respective localized home page (e.g. `href="/ja/`) instead of simply `href="/"`.
2. Adapted the "Menu" heading in the sidebar into an "Elixir School" link to the localized home page. (As the sidebar is a ToC, a homepage link should definitely be on there, but one is currently missing. The "Menu" heading, meanwhile, is entirely superfluous but also perfectly positioned and styled as a title / homepage link. 😄)

### DIFF
--- a/_includes/section-sidebar-menu.html
+++ b/_includes/section-sidebar-menu.html
@@ -7,7 +7,9 @@
   </header>
   <span>Elixir {{site.elixir.version}} - Erlang/OTP {{site.erlang.OTP}} [erts-{{site.erlang.erts}}]</span>
   <ul>
-
+    <li>
+      <a href="/{% if page.lang %}{{ page.lang }}/{% endif %}" class="{% unless locale.disable_transform %} up {%endunless%}{% if page.layout == "home" %}active{% endif %}">{{ site.data['locales'][language]['sections']['home'] }}</a>
+    </li>
   {% for sections in site.contents[language] %}
     {% assign section = sections[0] | strip %}
     {% assign chapters = sections[1] %}
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -14,7 +14,7 @@
 				<div class="inner">
 					<!-- Header -->
 					<header id="header">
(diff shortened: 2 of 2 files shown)


doomspork:
Thank you and welcome @thth!  I'm going to give @burden an opportunity to review this and provide any feedback, he's been handling much of our design.

burden:
Nice! The banner was always meant to be the proverbial "Home" link, and as such I think it's a **great** idea to add the localized homepage link to the banner. Keep that.

Unfortunately, I don't think changing the "Menu" header is a good idea, as it would break semantics (the HTML elements provide context to the search engines). To further explain, each page has a H1 element (ex: lesson's title, or blog's title), then H2 elements ("Menu", and "Table of Contents").

That being said, if everyone agrees that we could use a "Home" link, I'd like to propose that we locate it above the Basics chapter. 

What do the @elixirschool/developers think?

doomspork:
@burden I think that sounds like a great idea, would you mind working with @thth to make that happen?  😁

thth:
If, for a home page link on the sidebar, the semantics of the `h2` heading is an issue, would simply changing it to an `a` tag while retaining the same styling and position as the current "Menu" heading be a sensible solution?

Otherwise, @burden's proposal of inserting a home page link as a `li` listing above "Basics" would also make sense imo.

burden:
Thanks so much for your updates @thth 🎉

I'm not a big fan of having "Elixir School" both in the Sidebar, and in the Content section of every page. Let's go with the _Home link as sidebar listing_ option because the header serves a purpose in addition to the caption "Menu".

notactuallypagemcconnell:
We definitely should fix the history on this before merging. I can do it tomorrow if you cant or dont have the time @thth

thth:
Okay, I figured out the git issues, my bad on that.

Should actually be good to go now!

doomspork:
Thank you again @thth!  We really appreciate your continued involved 😁

PR #1553, opened 2018-10-13
Ecto: Associations lesson

SophieDeBenedetto:
Draft of the Ecto Associations lesson requested by this issue: https://github.com/elixirschool/elixirschool/issues/1519

Looking for feedback on:
1. Structure of lesson:
  a. Does it make sense to organize associations into these three types?
    * belongs to/has many, belongs to/has one, many-to-many
  b. Does it make sense to build out each type of relationship and only then get into inserting data and querying it? Or would it make sense to describe how to insert and query data for each type of relationship before moving on to the next one?

2. Direction of content: This post is more instructional, i.e. "this is how we do preload data". Not sure if that is preferred to a deeper dive in which we take a look at _how_ stuff works, i.e. "This is what is happening under the hood when we use Ecto.preload". 

Lastly, what do people think about putting out more advanced lessons on 
* Saving associated data: `build_assoc` vs. `put_assoc` vs. `cast_assoc`
* Working with nested relationships. For example the "has many through" relationships. Ex) a post has many comments and a comment belongs to a commenter. How would we build out the relationships between these three entities? How would we insert such data and query data. In particular how can you preload such nested associated schemas. 

I still need to: 
* Edit for grammar, spelling, etc. 
* Make sure all referenced documentation is properly linked to

### DIFF
--- a/_data/contents.yml
+++ b/_data/contents.yml
@@ -36,6 +36,7 @@ advanced:
 ecto:
   - basics
   - changesets
+  - associations
 specifics:
   - plug
   - eex
--- a/en/lessons/specifics/ecto/associations.md
+++ b/en/lessons/specifics/ecto/associations.md
@@ -1,0 +1,498 @@
+---
+version: 1.0.0
+title: Associations
+---
+
+In this section we'll learn how to use Ecto to define and work with associations between our schemas.
+
+{% include toc.html %}
+
+## Set Up
+
+We'll be building off of the app, `Example`, from the previous lessons. You can refer to the setup [here](./basics.md) for a quick refresher.
+
+## Types of Associations
+
+There are three types of associations we can define between our schemas. We'll look at what they are and how to implement each type of relationship.
+
+### Belongs To/Has Many
+
+We're adding some new entities to our example app's domain model so that we can catalogue our favorite films. We'll start with two schemas: `Movie` and `Character`. We'll implement a "has many/belongs to" relationship between these two schemas: A movie has many characters and a character belongs to a movie.
+
+#### The Has Many Migration
+
+Let's generate a migration for `Movie`:
+
+```console
+mix ecto.gen.migration create_movies
+```
+
+Open up the newly generated migration file and define your `change` function to create the `movies` table with a few attributes:
+
+```elixir
+# priv/repo/migrations/*_create_movies.exs
+defmodule Example.Repo.Migrations.CreateMovies do
+  use Ecto.Migration
+
+  def change do
+    create table(:movies) do
+      add :title, :string
+      add :tagline, :string
+    end
+  end
+end
+```
+
+#### The Has Many Schema
+
+We'll add a schema that specifies the "has many" relationship between a movies and its characters.
+
+```elixir
+# lib/example/movie.ex
+defmodule Example.Movie do
+  use Ecto.Schema
+
+  schema "movies" do
+    field :title, :string
+    field :tagline, :string
+    has_many :characters, Example.Character
+  end
+end
+```
+
+The `has_many/3` macro doesn't add anything to the database itself. What it does is use the foreign key on the associated schema, `characters`, to make a movie's associated characters available. This is what will allow us to call `movie.characters`.
+
+#### The Belongs To Migration
+
+Now we're ready to build our `Character` migration and schema. A character belongs to a movie, so we'll define a migration and schema that specifies this relationship.
+
+First, generate the migration:
+
+```console
+mix ecto.gen.migration create_characters
+```
+
+To declare that a character belongs to a movie, we need the `characters` table to have a `movie_id` column. We want this column to function as a foreign key. We can accomplish this with the following line in our `create_table/1` function:
+
+```elixir
+add :movie_id, references(:movies)
+```
+So our migration should look like this:
+
+```elixir
+# priv/migrations/*_create_characters.exs
+defmodule Example.Repo.Migrations.CreateCharacters do
+  use Ecto.Migration
+
+  def change do
+    create_table(:characters) do
+      add :name, :string
+      add :movie_id, references(:movies)
+    end
+  end
+end
+```
+
+#### The Belongs To Schema
+
+Our schema likewise needs to define the "belongs to" relationship between a character and its movie.
+
+```elixir
+# lib/example/character.ex
+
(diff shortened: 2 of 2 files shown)


en/lessons/specifics/ecto/02_associations.md
```diff
@@ -0,0 +1,672 @@
...
+
+## Table of Contents
+
+* Set up
+* Types of associations
+  * Belongs to/has many
+  * Belongs to/has one
+  * Many to Many
+* Saving associated data
+  * `build_assoc`
```
SophieDeBenedetto: I'm not sure if `build_assoc` vs. `put_assoc` is the right way to outline it. `put_assoc` is for any relationships that implement a `belongs_to` and `put_assoc` is for the `many_to_many`. `put_assoc` is not the only way  to save such data. You can also use `cast` assoc. I didn't include anything on cast assoc here since I felt like it was running a little long/would have required more of a deep dive but I'm open to including it. 

So with this in mind, might make more sense to outline more like:
* Saving data with `belongs_to`
* Saving data with `put_assoc`

en/lessons/specifics/ecto/02_associations.md
```diff
@@ -0,0 +1,672 @@
...
+  id: 1,
+  tagline: "Something about video games",
+  title: "Ready Player One"
+}
+```
+
+We can see that a new actor was created with an ID of "2" and the attributes we assigned it.
+
+## Querying For Associated data
+### Preloading
```
SophieDeBenedetto: How in depth is it appropriate do to on the topic of querying? Maybe it should include some basic queries: "give me all the movies featuring X actor" etc. Or is that content a better fit for the querying lesson?

notactuallypagemcconnell:
This is awesome! Thanks @SophieDeBenedetto -- I'll get to replying to this tonight after work :)

doomspork:
@SophieDeBenedetto wow, this is awesome!  Thank you and welcome 😁

I will review this today or tomorrow and give you my thoughts.  Thank you again!

en/lessons/specifics/ecto/02_associations.md
notactuallypagemcconnell: I think this makes sense unless I'm missing something myself -- feel free to rock it out in that direction

en/lessons/specifics/ecto/02_associations.md
notactuallypagemcconnell: I think sticking to the simplest basic examples is good for queries here. Query the data such that we get a real result, but we can save the complexities of composing a nontrivial query for other lessons.

en/lessons/specifics/ecto/02_associations.md
```diff
@@ -0,0 +1,773 @@
...
+  },
+  %Moviedb.Actor{
+    __meta__: #Ecto.Schema.Metadata<:loaded, "actors">,
+    id: 2,
+    movies: #Ecto.Association.NotLoaded<association :movies is not loaded>,
+    name: "Gary"
+  }
+]
+```
+
```
notactuallypagemcconnell: We definitely should cut out the extra newlines + `dd` at the bottom here haha

SophieDeBenedetto:
thanks for the feedback @notactuallypagemcconnell! I'll have these changes for you guys by mid-week.

SophieDeBenedetto:
@notactuallypagemcconnell its ready for another look now.

doomspork:
@notactuallypagemcconnell sorry I didn't see this ping sooner.  I'm okay with putting that in the styleguide but I don't want to delay @SophieDeBenedetto's merge any longer.  What do you think about merging and doing a follow-up PR for that?

en/lessons/specifics/ecto/02_associations.md
```diff
@@ -0,0 +1,672 @@
...
+```elixir
+# mix.exs
+defp deps do
+  [{:ecto, "~> 2.0"},
+   {:postgrex, "~> 0.11"}]
+end
+```
+
+Install your dependencies with `mix deps.get`.
```
midnio: Not sure, but I think you should show the how output looks like to users to be sure.

en/lessons/specifics/ecto/02_associations.md
```diff
@@ -0,0 +1,727 @@
...
+  distributor: nil,
+  id: 1,
+  tagline: "Something about video games",
+  title: "Ready Player One"
+}
+```
+
+We can see that a new actor was created with an ID of "2" and the attributes we assigned it.
+
+## Querying For Associated data
```
notactuallypagemcconnell: Going to move this to next lesson after discussions with @SophieDeBenedetto and @gj

en/lessons/specifics/ecto/02_associations.md
doomspork: @midnio for the sake of brevity we often exclude this output because it _will_ change slightly as PATCH versions are released, which ultimate leads to a lot confusion and/or PRs updating versions.

SophieDeBenedetto:
Do we want to merge this in or wait until we have other sections under `specifics/ecto/` are ready to go? At some point I imagine we'll want to do away with the existing Ecto lesson, currently in `specifics/ecto.md`?

SophieDeBenedetto:
@gj thanks for giving it such a close read! I incorporated pretty much all of your feedback.

en/lessons/specifics/ecto/03_associations.md
```diff
@@ -0,0 +1,567 @@
...
+  __meta__: #Ecto.Schema.Metadata<:built, "movies">,
+  actors: #Ecto.Association.NotLoaded<association :actors is not loaded>,
+  characters: #Ecto.Association.NotLoaded<association :characters is not loaded>,
+  distributor: #Ecto.Association.NotLoaded<association :distributor is not loaded>,
+  id: nil,
+  tagline: "Something about video games",
+  title: "Ready Player One"
+}
+
+iex> movie = Repo.insert!(movie)
```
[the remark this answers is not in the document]
SophieDeBenedetto: I'm going to skip this for now because I'll need to rebuild the repo with this code just to run that one line to get the output. Unless you want to fake it in which case be my guest :)

SophieDeBenedetto:
@doomspork updated to use `Example` app domain and removed section on setting up a new app.

Issue #1465, opened 2018-06-21
Add dark theme


sotojuan:
Would be great to add a toggle for a dark theme. I'm not a designer not good with colors but would be happy to implement the actual toggle button and behavior.

doomspork:
Hello @sotojuan!  This is something I think a great deal of people would love to have.  @burden and myself have discussed this a bit but have never come to a consensus on approach.  If you have any thoughts or could point us towards any projects you think might be useful, that would be tremendously helpful!

nscyclone:
Thanks to everyone involved into this! The dark theme looks really cool.
I suppose that the issue can be closed since all the changes are already available.

PR #1554, opened 2018-10-15
Create a Toggle-able Dark Theme

jereinhardt:
Thought I'd try to take a swing at #1465.  This PR changes the color scheme to create a dark theme for the site.  It also adds a toggle button at the bottom of the sidebar that allows users to switch back and forth between dark and default themes.

### DIFF
--- a/_assets/javascripts/main.js
+++ b/_assets/javascripts/main.js
@@ -278,6 +278,24 @@
 
 				});
 
+		// Theme
+			var cookieName = 'elixirschooltheme';
+
+			function toggleTheme(event) {
+				event.preventDefault();
+				var currentTheme = $.getCookie(cookieName);
+				if ( $.getCookie(cookieName) == '' ) {
+					currentTheme = 'default';
+				}
+				if ( currentTheme == 'default' ) {
+					document.cookie = cookieName + '=dark';
+				} else {
+					document.cookie = cookieName + '=default';
+				}
+				location.reload();
+			}
+
+			$("#changeThemes").on('click', toggleTheme);
 	});
 
 })(jQuery);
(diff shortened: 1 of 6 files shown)


doomspork:
@jereinhardt this is awesome!  I'll take a peek this week.  I've asked @burden to also review this, he was working on a dark theme as well 👍

burden:
@jereinhardt it appears great minds think alike :) I was also working on a theme as @doomspork mentioned.  It even looks like we started in the same place — same css, slightly different. Nevertheless, I wanted to share some thoughts and hoped we could discuss them. 

The first thing that jumped out to me is that it appears that the cookie's path is set to the page where the button was pressed.  We want the theme choice to be reflected throughout the visitor's experience on the site, but the current implementation is limited to per-page.  I don't believe this is by design, is it?
screen shot 2018-10-15 at 2 45 50 pm

The other thing I think we should address is whether we need an extra CSS file or not. Instead of making everyone download a slightly modified version of the original CSS, we should be adding onto `main.scss` to minimize the download, which is what I did over on my branch. Do you want to take a peek? https://github.com/elixirschool/elixirschool/pull/1555

I'm hoping we can figure out a way to incorporate our changes together. I spaced on jQuery's cookie implementation, which could be something for you to improve.

doomspork:
@jereinhardt I'd hate for your work to not be included.  @burden is your branch complete?  Is there anything that's left that @jereinhardt could help with?

I hate when folks work doesn't get merged 🙁

jereinhardt:
No worries.  Honestly, doing the work and then comparing it to @burden's helped me learn a few things, so it's not for nothing 😄

doomspork:
@jereinhardt I hope we can find other opportunities for you to contribute!  I really do appreciate you taking the time to open this PR 😁

PR #1555, opened 2018-10-15
Dark theme

burden:
- Dark theme that inverts most colors, and softens the rest.
- Toggle theme at bottom of sidebar.
- Theme will persist if back button is used.
- Does not require page refresh.

screen shot 2018-10-15 at 2 32 00 pm
screen shot 2018-10-15 at 2 34 45 pm

### DIFF
--- a/_assets/javascripts/js.cookie.js
+++ b/_assets/javascripts/js.cookie.js
@@ -1,0 +1,165 @@
+/*!
+ * JavaScript Cookie v2.2.0
+ * https://github.com/js-cookie/js-cookie
+ *
+ * Copyright 2006, 2015 Klaus Hartl & Fagner Brack
+ * Released under the MIT license
+ */
+;(function (factory) {
+	var registeredInModuleLoader = false;
+	if (typeof define === 'function' && define.amd) {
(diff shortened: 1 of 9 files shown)


doomspork:
Do we want to have the bottom towards the top of the menu instead of bottom, so it's easier to find?

burden:
> Do we want to have the button towards the top of the menu instead of bottom, so it's easier to find?

I went with the bottom because the top was seeming a little too busy. We already have the languages up top, then "Menu", then a slogan, followed up by the OTP version before getting to any actual menu. I didn't want to push the menu down any further.

That said, I was going to eventually try and tackle moving it under the hamburger icon, as per  @sotojuan's suggestion in https://github.com/elixirschool/elixirschool/issues/1465

> Is there anything left to do? I'd really like to find a way to incorporate @jereinhardt's work. I hate for anyone's contribute to not be included.

I feel ya, but everything here is complete. At some point it would be a good idea to circle back and remove that js.cookies.js lib, since jQuery can do the same thing. Wouldn't mind if @jereinhardt wanted to tackle that :)

PR #1563, opened 2018-10-19
Ecto Changesets chapter

svileng:
Changesets chapter for #1519.

I mostly followed the topics outlined in #1519 ; I decided not to have dedicated section for `Changeset design approaches`,  but rather try to sneakily introduce this kind of thinking, alongside the other content. Not sure if it was successful, so I'd love to hear people's feedback 😆 

Any feedback is of course welcome 😄

### DIFF
--- a/en/lessons/specifics/ecto/changesets.md
+++ b/en/lessons/specifics/ecto/changesets.md
@@ -1,0 +1,277 @@
+---
+version: 1.0.0
+title: Changesets
+---
+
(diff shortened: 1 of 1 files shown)


doomspork:
@svileng could you change the filename to exclude the `03_` prefix?

svileng:
Changed to `changesets.md` 👍

doomspork:
Thank you @svileng!  I need to make a couple configuration changes and then the change will be live on the site 😁

PR #1582, opened 2018-10-24
Clarify frontpage language to note that right side entries are of the blog

notactuallypagemcconnell:
Right now, if a new visitor comes it reads 'Recent Posts`juxtaposed with the book content of the left side once opening the index.
It is clear this is reference material present and to be consumed, but what is the right side and lower?
We do not translate blog entries at this time, we do concentrate on the book-style-content for obvious reasons.
This makes the boundary more clear.

### DIFF
--- a/_includes/section-home-blog-posts.html
+++ b/_includes/section-home-blog-posts.html
@@ -1,7 +1,7 @@
 <!-- Section -->
 <section>
   <header class="major">
-    <h2>Recent Posts</h2>
+    <h2>Blog</h2>
   </header>
   <div class="posts">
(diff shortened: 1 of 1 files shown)


notactuallypagemcconnell:
Lol @ me for a commit msg longer than the diff of the change.

notactuallypagemcconnell:
Also, I know @gemantzu is already doing a ton but if you have any interest in organizing effort to translate blog entries as well please ping me privately and we can discuss and bring a proposal up publicly here if we decide it worth the effort. There is already a lot going on here, but if we could get that working and people would find value I'd be happy to contribute however possible.

PR #1583, opened 2018-10-24
Fix table headers' text color in dark theme

nscyclone:
<details>
<summary>Table headers are invisible in dark theme:</summary>

Animation representing the problem

</details>


doomspork:
Thank you @nscyclone and @gemantzu 😁

PR #1586, opened 2018-10-30
[style] Make inline code padding smaller

zillou:
Old style:

image

New style:

image

### DIFF
--- a/_assets/stylesheets/base/_typography.scss
+++ b/_assets/stylesheets/base/_typography.scss
@@ -147,12 +147,6 @@
       margin: 0;
       font-size: 80%;
     }
-      padding: 0.2em 0.4em;
-      margin: 0;
-      font-size: 95%;
-      border-radius: 3px;
-      border: none;
-    }
 	}
 
 	pre {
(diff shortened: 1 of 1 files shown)


_assets/stylesheets/base/_typography.scss
```diff
@@ -141,6 +141,14 @@
 		font-size: 1.15em;
 		margin: 0 0.25em;
 		padding: 0.25em 0.65em;
+
+    &.highlighter-rouge {
```
doomspork: @zillou could we change this to:

```suggestion
    &.highlighter-rouge {
      padding: 0.2em 0.4em;
      margin: 0;
      font-size: 80%;
    }
```

doomspork: Doesn't look like that worked as intended @zillou, the result is broken code 😞

zillou: Fixed with 8ca0d23

doomspork:
Thank you @zillou 😁

PR #1593, opened 2018-11-01
Close ticks at latest addition

midnio:
I just realized I didn't close ticks for the code example, sorry.


gemantzu:
Thank you @midnio

PR #1599, opened 2018-11-11
Update ecto lesson for Ecto 3

zillou:
I noticed there's another plan to provide more detailed lessons for Ecto in https://github.com/elixirschool/elixirschool/issues/1519, but before we finish that, I think it's better to update the original ecto lesson so the reader can still follow along.


doomspork:
Wonderful idea @zillou!  You are _awesome_!  I hope you're having a great weekend 😁

PR #1600, opened 2018-11-11
Add the reason why first argument of and... cant be any value

midnio:


en/lessons/basics/basics.md
```diff
@@ -210,6 +210,8 @@ iex> not 42
 ** (ArgumentError) argument error
 ```
 
+Note: The reason why first argument of `and`, `or` or `not` can not be any value except booleans is Erlang does not evaluates the first argument in `and`, `or` or `not` so it will want boolean, but second argument will be evaluated. (Elixir is built on Erlang VM (BEAM))
```
doomspork: Could we rephrase this slightly?  Maybe something like:

> Note: It's important to remember that the first arguments in `and`, `or` and `not` _must be_ a boolean.  The reason for this is the first argument is evaluated by Erlang whereas the second argument is evaluated by Elixir.

midnio:
Just did it! @doomspork

doomspork:
Thank you @midnio!

PR #1605, opened 2018-11-15
[PT] Updating basics for 1.2.0

lucaspbordignon:


pt/lessons/basics/basics.md
```diff
@@ -210,6 +210,8 @@ iex> not 42
 ** (ArgumentError) argument error
 ```
 
+Nota: Elixir é baseada na máquina virtual da linguagem Erlang(BEAM). Se você é um desenvolvedor Erlang, `and` e `or` são mapeados para `andalso` e `orelse`, os quais não avaliam o primeiro argumento e avaliam o segundo.
```
anderkonzen: Thanks @lucaspbordignon! This is a nice contribution, keeping the lessons updated!

I have two suggestions. First, I believe "Elixir é basead**o** na..." is better, since `Elixir` is treated as a masculine noun throughout the Portuguese translations. Second, there is a missing space here: "Erlang(BEAM)" -> "Erlang (BEAM)".

lucaspbordignon: Thank you so much for the review! I agree with both, they were just syntax mistakes, updates done! 😄

Issue #1611, opened 2018-11-18
New content: OTP + Distribution


SophieDeBenedetto:
Distributed OTP
- [ ] Introduction to distribution: why do we want to be able to do this?
- [ ] Intro to communicating between nodes 
    
  * what is a node?
  * synchronous communication with Node.spawn_link

- [ ] Asychronous Distributed Tasks
  
  * What are distributed supervised tasks and why do we want them?
  * Let's build a simple command line chatting app that allows users to chat with one another from the iex console. 
  - [ ] Generate a new supervisor app
  - [ ] Add the Task Supervisor to the supervision tree 
  - [ ] Defining an interface module: sending messages to a remote node via the Task Supervisor
  - [ ] Defining an interface module: receiving/responding to messages
  - [ ] See it in action! start two named iex nodes and start chatting!
  - [ ] Testing distributed code

SophieDeBenedetto:
I'll have a draft ready for this soon

PR #1612, opened 2018-11-18
OTP distribution lesson 

SophieDeBenedetto:
addresses issue #1611

### DIFF
--- a/en/lessons/advanced/otp-distribution.md
+++ b/en/lessons/advanced/otp-distribution.md
@@ -3,17 +3,17 @@ version: 1.0.0
 title: OTP Distribution
 ---
 ## Introduction to Distribution
-Elixir runs on the Erlang VM, which means it has access to to Erlang's powerful [distribution functionality.](http://erlang.org/doc/reference_manual/distributed.html)
-
-> A distributed Erlang system consists of a number of Erlang runtime systems communicating with each other. Each such runtime system is called a node.
-
 We can run our Elixir apps on a set of different nodes distributed across a single host or across multiple hosts. Elixir allows us to communicate across these nodes via a few different mechanisms which we will outline in this lesson.
 
 {% include toc.html %}
 
 ## Communication Between Nodes
 
-A node is any Erlang runtime system that has been given a name. We can start a node by opening up `iex` process and naming it:
+Elixir runs on the Erlang VM, which means it has access to to Erlang's powerful [distribution functionality.](http://erlang.org/doc/reference_manual/distributed.html)
+
+> A distributed Erlang system consists of a number of Erlang runtime systems communicating with each other. Each such runtime system is called a node.
+
+A node is any Erlang runtime system that has been given a name. We can start a node by opening up `iex` session and naming it:
 
 ```bash
 iex --sname alex@localhost
@@ -57,15 +57,15 @@ Hi, my name is Kate
 #PID<10507.132.0>
 ```
 
-#### A Note of I/O and Nodes
+#### A Note on I/O and Nodes
 
-Notice that, although `Kate.say_name` is getting executed on the remote node, it is the local, or calling, node that receives the `IO.puts` output. That is because the local node is the **group leader**. The Erlang VM manages I/O via processes. This allows us to send execute I/O tasks, like `IO.puts`, across distributed nodes. These distributed processes are managed by the I/O process group leader. The group leader is always the node that spawns the process. So, since our `alex` node is the one from which we called `spawn_link`, that node is the group leader and the output of `IO.puts` will be directed to the standard output stream of that node.
+Notice that, although `Kate.say_name` is getting executed on the remote node, it is the local, or calling, node that receives the `IO.puts` output. That is because the local node is the **group leader**. The Erlang VM manages I/O via processes. This allows us to execute I/O tasks, like `IO.puts`, across distributed nodes. These distributed processes are managed by the I/O process group leader. The group leader is always the node that spawns the process. So, since our `alex` node is the one from which we called `spawn_link`, that node is the group leader and the output of `IO.puts` will be directed to the standard output stream of that node.
 
 #### Responding to Messages
 
 What if we want the node that receives the message to send some *response* back to the sender? We can use a simple `receive` and [`send`](https://hexdocs.pm/elixir/Process.html#send/3) setup to accomplish exactly that.
 
-We'll have our `alex` node spawn a link to the `kate` node and give the `kate` node an anonymous function to execute. That anonymous function will listen fro the receipt of a particular tuple describing a message and the PID of the `alex` node. It will respond to that message by `send`-ing back a message to the PID of the `aliex` node:
+We'll have our `alex` node spawn a link to the `kate` node and give the `kate` node an anonymous function to execute. That anonymous function will listen for the receipt of a particular tuple describing a message and the PID of the `alex` node. It will respond to that message by `send`-ing back a message to the PID of the `alex` node:
 
 ```elixir
 iex(alex@localhost)2> pid = Node.spawn_link :kate@localhost, fn ->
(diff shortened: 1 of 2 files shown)


en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
+---
+version: 1.0.0
+title: OTP Distribution
+---
+## Introduction to Distribution
```
doomspork: Could you add whitespace before and after headers please?

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+Let's open up another node in another terminal window:
+
+```bash
+iex --sname kate@localhost
+iex(kate@localhost)1>
+```
+
+These two nodes can send messages to one another using `Node.spawn_link/1`.
+
+### Communicating with [`Node.spawn_link/1`](https://hexdocs.pm/elixir/Node.html#spawn_link/2)
```
doomspork: I'm not sure we should put links in our headers since these are themselves turn into links in the ToC, this will likely cause some funky issues.

Let's mention the documentation later in the lesson when it makes a little more sense.  We don't typically push the docs on folks until they've gotten through a section 😁

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+
+Elixir runs on the Erlang VM, which means it has access to to Erlang's powerful [distribution functionality.](http://erlang.org/doc/reference_manual/distributed.html)
+
+> A distributed Erlang system consists of a number of Erlang runtime systems communicating with each other. Each such runtime system is called a node.
+
+A node is any Erlang runtime system that has been given a name. We can start a node by opening up `iex` session and naming it:
+
+```bash
+iex --sname alex@localhost
+iex(alex@localhost)1>
```
doomspork: We don't show `iex` line numbers (I disabled them via config for myself), it's mentioned in the styleguide.

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+#PID<10507.132.0>
+```
+
+#### A Note on I/O and Nodes
+
+Notice that, although `Kate.say_name` is getting executed on the remote node, it is the local, or calling, node that receives the `IO.puts` output. That is because the local node is the **group leader**. The Erlang VM manages I/O via processes. This allows us to execute I/O tasks, like `IO.puts`, across distributed nodes. These distributed processes are managed by the I/O process group leader. The group leader is always the node that spawns the process. So, since our `alex` node is the one from which we called `spawn_link`, that node is the group leader and the output of `IO.puts` will be directed to the standard output stream of that node.
+
+#### Responding to Messages
+
+What if we want the node that receives the message to send some *response* back to the sender? We can use a simple `receive` and [`send`](https://hexdocs.pm/elixir/Process.html#send/3) setup to accomplish exactly that.
```
doomspork: If this is referring to the function `send`, it should include the arity 👍

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+iex(alex@localhost)4> send(pid, {:hi, self()})
+{:hi, #PID<0.106.0>}
+iex(alex@localhost)5> flush()
+:sup?
+:ok
+```
+
+#### A Note On Communicating Between Nodes on Different Networks
+If you want to send messages between nodes on different networks, we need to start the named nodes with a shared cookie:
+
```
doomspork: Extra whitespace here.

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+iex --sname alex@localhost --cookie secret_token
+```
+
+```bash
+iex --sname kate@localhost --cookie secret_token
+```
+
+Only nodes started with the same `cookie` will be able to successfully connect to one another.
+
+#### `Node.spawn_link` Limitations
```
doomspork: Function references should include arity.  This applies here and in the paragraphs belong.

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+    Supervisor.start_link(children, opts)
+  end
+end
+```
+
+Now we know that wherever our application is started on a given node, the `Chat.Supervisor` is running and ready to supervise tasks.
+
+### Sending Messages with Supervised Tasks
+
+We'll start supervised tasks with the [`Task.Supervisor.async/4`](https://hexdocs.pm/elixir/master/Task.Supervisor.html#async/5) function.
```
doomspork: You have `/4` and link to `/5`, we should fix one or the other.

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+This function takes in four arguments:
+
+* The supervisor we want to use to supervise the task. This can be passed in as a tuple of `{SupervisorName, remote_node_name}` in order to supervise the task on the remote node.
+* The name of the module on which we want to execute a function
+* The name of the function we want to execute
+* Any arguments that need to be supplied to that function
+
+Out Chat application is pretty simple. It sends messages to remote nodes and remote nodes respond to those messages by `IO.puts`-ing them out to the STDOUT of the remote node.
+
+First, let's define a function, `receive_message`, that we want our task to execute on a remote node.
```
doomspork: `receive_message/1`

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+# lib/chat.ex
+defmodule Chat do
+  ...
+
+  def send_message(recipient, message) do
+    spawn_task(__MODULE__, :receive_message, recipient, [message])
+  end
+
+  def spawn_task(module, fun, recipient, args) do
+    Task.Supervisor.async(remote_supervisor(recipient), module, fun, args)
```
doomspork: Our code examples follow the [community styleguide](https://github.com/christopheradams/elixir_style_guide) for the most part, so is covered under starting pipelines with a bare value [here](https://github.com/christopheradams/elixir_style_guide#bare-variables).

We can refactor this to something like this:

```elixir
recipient
|> remote_supervisor()
|> Task.Supervisor.async(module, fun, args)
|> Task.await
```

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+Let's take a look at our other testing approach--configuring the application to behave differently in different environments.
+
+### Environment-Specific Applicaton Configuration
+
+The part of our code that tells `Task.Supervisor` to start a supervised task on a remote node is here:
+
+```elixir
+# app/chat.ex
+def spawn_task(module, fun, recipient, args) do
+  Task.Supervisor.async(remote_supervisor(recipient), module, fun, args)
```
doomspork: See note above about starting pipelines with a bare value.

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,430 @@
...
+```
+
+Lastly, we'll update our `Chat.remote_supervisor` function to look up and use the function stored in our new application variable:
+
+```elixir
+# lib/chat.ex
+defp remote_supervisor(recipient) do
+  Application.get_env(:chat, :remote_supervisor).(recipient)
+end
+```
```
doomspork: We typically wrap up our lessons with some small comment other than a code snippet, what do you think about adding something here to summarize what we've covered?

SophieDeBenedetto:
@doomspork thanks for your feedback! I _think_ I addressed everything 🚀

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,445 @@
...
+
+  def send_message(recipient, message) do
+    spawn_task(__MODULE__, :receive_message, recipient, [message])
+  end
+
+  def spawn_task(module, fun, recipient, args) do
+    recipient
+    |> remote_supervisor()
+    |> Task.Supervisor.async(module, fun, args)
+    |> Task.await
```
doomspork: This is function so it should include parenthesis: `Task.await()`

en/lessons/advanced/otp-distribution.md
```diff
@@ -0,0 +1,445 @@
...
+end
+```
+
+If we run our tests via `mix test`, we see it fail with the following error:
+
+```elixir
+** (exit) exited in: GenServer.call({Chat.TaskSupervisor, :moebi@localhost}, {:start_task, [#PID<0.158.0>, :monitor, {:sophie@localhost, #PID<0.158.0>}, {Chat, :receive_message_for_moebi, ["hi", :sophie@localhost]}], :temporary, nil}, :infinity)
+         ** (EXIT) no connection to moebi@localhost
+```
+
```
doomspork: Extra whitespace.

PR #1614, opened 2018-11-19
Ecto Subsection

doomspork:
As part of our new push to include an Ecto subsection we need to set up the configuration for that.  This PR seeks to do that in addition to a few other things:

- [X] Create `Ecto` direction
- [X] Move `Changeset` lesson into the `Ecto` subsection
- [X] Merge changes from `ecto_basics` into `ecto`
- [X] Move `Basics` into `Ecto`
- [X] Add new lessons and section to `config/contents.yml` 
- [X] Add Ecto subsection to `_data/locales/en.yml`

I consider this still a WIP because I'd like to test it more thoroughly locally to ensure there's no hiccups and to verify whether or not I need to make changes to the translation configurations.


_data/contents.yml
```diff
@@ -39,6 +39,9 @@ specifics:
   - ets
   - mnesia
   - debugging
+ecto:
```
doomspork: @SophieDeBenedetto now that you're adding new lessons I thought you might like to know that this is the file we use to manage the order of our lessons.

This way we don't need to prefix filenames with a number (like `03_associations.md`).

SophieDeBenedetto: awesome, I would like to know that 😃

notactuallypagemcconnell: HA I never knew this and suggested the aforementioned naming convention. My bad! Great to know.

en/lessons/ecto/basics.md
```diff
@@ -0,0 +1,231 @@
+---
+version: 1.0.0
+title: Basics
+---
+
+Ecto is an official Elixir project providing a database wrapper and integrated query language. With Ecto we're able to create migrations, define schemas, insert and update records, and query them.
```
SophieDeBenedetto: "query them" -> "query for them"

en/lessons/ecto/changesets.md
```diff
@@ -22,27 +24,31 @@ iex> %Ecto.Changeset{}
...
 ```elixir
-iex> Ecto.Changeset.cast(%User{name: "Bob"}, %{}, [:name])
-#Ecto.Changeset<action: nil, changes: %{}, errors: [], data: #User<>,
+iex> Ecto.Changeset.cast(%Friends.Person{name: "Bob"}, %{}, [:name])
+#Ecto.Changeset<action: nil, changes: %{}, errors: [], data: #Friends.Person<>,
  valid?: true>
  ```
 
- The first parameter is the original data — an empty `%User{}` struct in this case. Ecto is smart enough to find the schema based on the struct itself. Second in order are the changes we want to make — just an empty map. The third parameter is what makes `cast/4` special: it is a list of fields allowed to go through, which gives us the ability to control what fields can be changed and safe-guard the rest.
+The first parameter is the original data — an empty `%User{}` struct in this case. Ecto is smart enough to find the schema based on the struct itself. Second in order are the changes we want to make — just an empty map. The third parameter is what makes `cast/4` special: it is a list of fields allowed to go through, which gives us the ability to control what fields can be changed and safe-guard the rest.
```
SophieDeBenedetto: The rest of the lesson refers to a User schema and struct. Should we update to `Friends.Person`?

doomspork: @SophieDeBenedetto I am in the process of doing that so these Ecto lessons will flow together.  That isn't something we've done previously but it's feedback we've heard and I think now is as good a time as any to start 😁 

Whatcha think?

doomspork: It does look like your PR here would need to be modified: https://github.com/elixirschool/elixirschool/pull/1553/.  On the upside we can eliminate all the boilerplate.  On the downside, we need to rethink some of the associations.

SophieDeBenedetto: Ah yea that makes sense @doomspork. So on the one hand I'm definitely down to update that lesson but I think @notactuallypagemcconnell and @gj are using the Movie app/domain model as the basis for their lesson on querying. I also think that its okay to have one or two different domain models/apps used throughout the section, as long as each section sticks to just one--whereas here in the changeset lesson we started out talking about `Friend`s and then switched to referring to `User`s. I also think there's some value in showing _some_ repetition, i.e. generating a new app and migrations, since the audience is beginners. 

Anyways I'll defer to the general consensus--what do others think?

doomspork: @SophieDeBenedetto the switch in Changesets is only a result of this PR being a WIP 😁  

The original changeset uses User and whereas basic lesson uses Friend.  I personal don't think we should be switching from Friend, to Users, to Movies in the span of 3-4 lessons, seems a bit much IMO.  

Another reason I'd rather we not switch between different data models throughout this section is that it requires each lesson have the same repetitive setup of schema, migrations, et al which is that we have successfully avoided throughout our other lessons.

Somewhat related sidenote:  Something to keep in mind is the majority of Elixir School's traffic are non-native English speakers.  Anything we can do to make concepts and ideas flow together between, the easier it is to translate or make sense of if English is not your primarily language.  Removing unnecessary verbage, duplicated content, etc all makes the job of translators easier.

doomspork: I should also add:  We can (and probably should) most certainly add additional data models to make sensible queries in the association.  I'd like to avoid a huge divergence from what folks are already putting together as they're working through the lessons.

Maybe it's as simple as the application from `Friends` to `Example` (our more traditional namespace for example code).  With that change it from go from `Friends.Person` to `Example.Person` which then allows us to do something like  `Example.Movies` with `has_many :actors, Example.Person`

doomspork:
@SophieDeBenedetto and @elixirschool/english this _should_ be ready for a final review 😁

Something to think about either to address now or in a follow-up lesson: `age` does not make sense.  It should be `birthdate`.

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,371 @@
...
+ data: #Person<>,
+ valid?: true
+>
+```
+
+## Validations
+
+We've learned how to create changesets and apply changes to them but how do we validate those changes?
+
+If we take our previous examples which lack validation, any changes to a person's name is considered valid which allows mean a blank string counts:
```
SophieDeBenedetto: "allows mean" -> "means"

en/lessons/ecto/basics.md
```diff
@@ -0,0 +1,232 @@
...
+
+```elixir
+  defp deps do
+    [
+      {:ecto, "~> 2.0"},
+      {:postgrex, "~> 0.11"}
+    ]
+  end
+```
+
```
gemantzu: Should we bump to latest versions?

en/lessons/ecto/basics.md
```diff
@@ -0,0 +1,232 @@
...
+```elixir
+config :example, Example.Repo,
+  adapter: Ecto.Adapters.Postgres,
+  database: "example_repo",
+  username: "postgres",
+  password: "",
+  hostname: "localhost"
+```
+
```
gemantzu: Isn't the default password for postgres, well, postgres? :laughing:

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,372 @@
...
+ data: #Person<>,
+ valid?: true
+>
+```
+
+What we see is the same valid changeset from before with no mention of the `"email"` value because it was not explicitly permitted.
+
+### `change/2`
+
+If the source of the data change is truthworthy `change/2` can be a viable alternative to `cast/4`.
```
gemantzu: `trustworthy,`
Am I just misguided on the use of comma or is it really needed here?

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,372 @@
...
+ action: nil,
+ changes: %{name: ""},
+ errors: [],
+ data: #Person<>,
+ valid?: true
+>
+```
+
+This won't work for a real application so let's look at how we can handle this with Ecto.
+Lucky for us, as with many of the Elixir libraries, Ecto comes with a slew of awesome features one of which is built-in validaitons.
```
gemantzu: Maybe change the wording for `Lucky for us` to something else? Like `Luckily for us, `, or `Thankfully`, or something else.

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,372 @@
...
+
+Above we introduced two new functions: [`get_field/3`](https://hexdocs.pm/ecto/Ecto.Changeset.html#get_field/3) and [`add_error/4`](https://hexdocs.pm/ecto/Ecto.Changeset.html#add_error/4).
+While we can probably guess what they're doing, let's walk through our validation anyways:
+
++ First we use `get_field/3` to retreive the value for `:name`.
+It's important to know that `get_field/3` will try to get the value from the changes first before falling back to the original data.
++ Next we downcase our string to make it easier to compare with our prohibited words.
++ We split it into parts (`"test user"` becomes `["test", "user"]`).
++ With `Enum.any?/2` we check if any part of our name is included in the prohibited words list.
++ Finally, if there we do have an unsafe name we rely on `add_error/3` to add a new error to our change.
```
gemantzu: `Finally, if we do have an unsafe name in there` ?

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,372 @@
...
+
+No surprise to us, our new validation rule works!
+
+## Adding changes programatically
+
+Sometimes you want to introduce changes to a changeset manually.
+We previously saw how to use `changes/2` to apply a map of changes to our changeset, now we'll look at a different approach using `put_change/3`.
+
+To demonstrate `put_change/3` we'll update our `changeset/2` function to allow `:name` to be blank and when it's blank set it to `"Anonymous"`.
+In order to accomplish this let us create a new `set_anonymous/1` function that accepts and returns a changeset as we did earlier:
```
gemantzu: `In order to accomplish this, `
A wild comma appears.

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,372 @@
...
+
+  if is_nil(name) or name == "" do
+    put_change(changeset, :name, "Anonymous")
+  else
+    changeset
+  end
+end
+```
+
+Nothing to difficult, right?  If the `:name` is blank or an empty string we want to put a change on to the changeset making `:name` equal to `"Anonymous"`.
```
gemantzu: Nothing too difficult, right?  If the `:name` is blank or an empty string, we want to put a change on to the changeset making `:name` equal to `"Anonymous"`.

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,372 @@
...
+
+With our new function in place, let's update see what `changeset/2` looks like now:
```
gemantzu: `let's update see`? `Let's see` probably.

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,372 @@
...
+  struct
+  |> cast(params, [:age, :name])
+  |> validate_required([:age])
+  |> validate_number(:age, greater_than: 0, less_than: 120)
+  |> validate_name_appropriate()
+  |> set_anonymous()
+end
+```
+
+Now if we don't pass `:name` then it will be set to `"Anonymous"` as expected:
```
gemantzu: `:name`,

en/lessons/ecto/basics.md
doomspork: Honestly I'm not sure.  I use Postgres.app and my is setup to be blank.

en/lessons/ecto/basics.md
```diff
@@ -0,0 +1,232 @@
...
+### Migrations
+
+To create and modify tables inside the postgres database Ecto provides us with migrations.
+Each migration describes a set of actions to be performed on our database, like which tables to create or update.
+
+Since our database doesn't have any tables yet, we'll need to create a migration to add some.
+The convention in Ecto is to pluralize our tables so for application we'll need a `peoples` table, so let's start there with our migrations.
+
+The best way to create migrations is the mix `ecto.gen.migration <name> task`, so in our case let's use:
```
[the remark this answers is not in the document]
doomspork: Good catch

en/lessons/ecto/basics.md
```diff
@@ -0,0 +1,232 @@
...
+```shell
+$ mix ecto.migrate
+```
+
+### Schemas
+
+Now that we've created our initial table we need to tell Ecto more about it, part of how we do that is through schemas.
+A schema is a module that defines mappings to the underlying database table's fields.
+
+While Ecto favors pluralize database table names, the schema is typically singular, so we'll create a `Person` schema to accomplany our table.
```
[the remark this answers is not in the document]
doomspork: and "accomplany" is spelled wrong.

en/lessons/ecto/changesets.md
svileng: Just some food for thought:

I cannot find the exact article / research around this at the moment, but I remember reading some advice on writing tutorials that suggested avoiding expressions like "Nothing too difficult, right?", "This is super easy", "Few basic / simple steps" etc. Basically implying that something is very easy or simple. The thinking behind this is:

1) Readers who struggle with the lesson, for whatever reason, will find this kind of expressions discouraging (e.g. "Why do I struggle if this is supposed to be easy?")

2) Readers who do well might will feel lower sense of achievement.

Since then I stopped using those when writing tutorials / documentation / etc.

Again, this is just some food for thought 😃 I tend to over-think things sometimes 😄

en/lessons/ecto/changesets.md
```diff
@@ -0,0 +1,372 @@
...
+## Schemaless changesets
+
+An exciting feature of `Ecto.Changeset` is the ability to apply changesets to more than just our Ecto schemas.
+With schemaless changes we can provide the same level of data validation and transformation, without the schema.
+
+While they are beyond the scope of these lessons, everything we need to know can be found in the official [schemaless changesets](https://hexdocs.pm/ecto/Ecto.Changeset.html#module-schemaless-changesets) docs.
+
+## Conclusion
+
+This is just the tip of the iceberg when it comes to the punch that `Ecto.Changeset`'s packs.
```
svileng: Just a note that `tip of the iceberg` and `packs a punch` are common expressions, but might be hard to translate in other languages, making translator's job more difficult 🤔

en/lessons/ecto/changesets.md
doomspork: That makes perfect sense @svileng, I think so much so we that add a note to our styleguide about it.

It may be a day or two before I can get back to these updates but when I do I will fix this 😁

en/lessons/ecto/changesets.md
doomspork: @svileng please call these out!  I _try_ not to use a lot informal local terms but sometimes I don't catch myself.  One thing I look for in these reviews is things like this!  I want to make translation as easy as we can 😁

SophieDeBenedetto:
@doomspork looks great to me but sadly the build broke :(

doomspork:
@gemantzu are you going to re-review this or should I fix the build and merge?

en/lessons/ecto/basics.md
```diff
@@ -0,0 +1,231 @@
...
+
+If you receive any errors, make sure that the configuration information is correct and that your instance of postgres is running.
+
+### Migrations
+
+To create and modify tables inside the postgres database Ecto provides us with migrations.
+Each migration describes a set of actions to be performed on our database, like which tables to create or update.
+
+Since our database doesn't have any tables yet, we'll need to create a migration to add some.
+The convention in Ecto is to pluralize our tables so for application we'll need a `peoples` table, so let's start there with our migrations.
```
gemantzu: `people` table

en/lessons/ecto/basics.md
```diff
@@ -0,0 +1,231 @@
...
+
+The best way to create migrations is the mix `ecto.gen.migration <name> task`, so in our case let's use:
```
gemantzu: task is still shown inside ``

en/lessons/ecto/changesets.md
```diff
@@ -22,27 +24,31 @@ iex> %Ecto.Changeset{}
 
 As you can see, it has some potentially useful fields, but they are all empty.
 
-For a changeset to be truly useful, when we create it, we need to provide a blueprint of what the data is like. `Ecto.Schema` is exactly that — a blueprint of all fields and their types. Let's look at a simple schema for a user:
+For a changeset to be truly useful, when we create it, we need to provide a blueprint of what the data is like.
+What better blueprint for our data than the schemas we've created the define our fields and types?
```
gemantzu: __that__ define

PR #1617, opened 2018-11-21
Add Plug.ErrorHandler to Plug Lesson

### LINKED ISSUE #1604: Add Plug.ErrorHandler content in plug lesson
When I updated the plug lesson I removed a portion of a code snippet that had `use Plug.ErrorHandler`. That was included without the `handle_errors` function definition and with no explanation. Might be nice to add a section to the Plug lesson on error handling using [`Plug.ErrorHandler`](https://hexdocs.pm/plug/Plug.ErrorHandler.html).

aturkewi:
Closes #1604 

- Adds `Plug.ErrorHandler` to the plug lesson
- Be consistent with use of `child_spec`
  - It was used in some places and not in others with no explanation. Made it consistent for the implicit use, but this can be changed if needed, I would just recommend we change it throughout the lesson if we go the explicit route.
- Update the `/upload` route to be a GET request so it's easier to test in browser and you don't need to setup a form or use another app to see that verify is working

cc @SophieDeBenedetto

### DIFF
--- a/en/lessons/specifics/plug.md
+++ b/en/lessons/specifics/plug.md
@@ -109,7 +109,7 @@ defmodule Example.Application do
 end
 ```
 
->Note: We do not have to call `child_spec` here, this function will be called by the supervisor starting this process. We simply pass a tuple with the module that we want the child spec built for and then the three options needed.
+_Note_: We do not have to call `child_spec` here, this function will be called by the supervisor starting this process. We simply pass a tuple with the module that we want the child spec built for and then the three options needed.
 
 This starts up a Cowboy2 server under our app's supervision tree. It starts Cowboy running under the HTTP scheme (you can also specify HTTPS), on the given port, `8080`, specifying the plug, `Example.HelloWorldPlug`, as the interface for any incoming web requests.
 
--- a/en/lessons/specifics/plug.md
+++ b/en/lessons/specifics/plug.md
@@ -211,13 +211,13 @@ defmodule Example.Plug.VerifyRequest do
   def init(options), do: options
 
   def call(%Plug.Conn{request_path: path} = conn, opts) do
-    if path in opts[:paths], do: verify_request!(conn.body_params, opts[:fields])
+    if path in opts[:paths], do: verify_request!(conn.params, opts[:fields])
     conn
   end
 
-  defp verify_request!(body_params, fields) do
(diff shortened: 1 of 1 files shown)


en/lessons/specifics/plug.md
```diff
@@ -80,7 +80,7 @@ It receives a `%Plug.Conn{}` connection struct as its first argument and is expe
 
 We need to tell our application to start up and supervise the Cowboy web server when the app starts up.
 
-We'll do so with the [`Plug.Cowboy.child_spec/1`](https://hexdocs.pm/plug_cowboy/Plug.Cowboy.html#child_spec/1) function.
+We'll do so with the [`Plug.Adapters.Cowboy.child_spec/1`](https://hexdocs.pm/plug_cowboy/Plug.Cowboy.html#child_spec/1) function.
```
SophieDeBenedetto: So this actually should be `Plug.Cowboy.child_spec/1`. Latest version of Cowboy (Cowboy2) [instructs users that `Plug.Adapters.Cowboy` is deprecated.](https://hexdocs.pm/plug/Plug.Adapters.Cowboy.html)

en/lessons/specifics/plug.md
```diff
@@ -109,6 +109,8 @@ defmodule Example.Application do
 end
 ```
 
+_Note_: We do not have to call `child_spec` here, this function will be called by the supervisor starting this process. We simply pass a tuple with the module that we want the child spec built for and then the three options needed.
```
SophieDeBenedetto: 👍

SophieDeBenedetto:
@aturkewi this looks GREAT! One small comment otherwise looks good to me. Thanks! 💃 
And thank you for all the other fixes rolled into this PR 🍨

en/lessons/specifics/plug.md
aturkewi: Thanks so much! Not sure how I missed that 🤦‍♂️

SophieDeBenedetto:
@doomspork I'd be happy to merge this in but not sure what the deploy process is? Or if we prefer to have more than one review before merging? Let me know!

gemantzu:
version bump?

SophieDeBenedetto:
@gemantzu ah good catch. I'll open a PR for that now. Thanks!

PR #1628, opened 2018-11-27
bump plug lesson version number

SophieDeBenedetto:


en/lessons/specifics/plug.md
```diff
@@ -1,5 +1,5 @@
 ---
-version: 2.0.0
+version: 2.1.0
```
doomspork: Why are we bumping this?  Did we miss it with a previous change?

doomspork: I see we did 👍

PR #1629, opened 2018-11-28
[zh-hans] fix content

wardenlym:
Hi, 
a little content fix.
now we use Plug.Cowboy definitely


doomspork:
Thank you and welcome @wardenlym!  

There's a couple other changes to the Plug lesson in addition to this one.  I'll defer to our @elixirschool/chinese team on how we want to handle this.  We'll need to bump the lesson version but perhaps that should wait until the error handler changes are incorporated.

PR #1634, opened 2018-11-28
Add section on functions and pattern matching.

meryldakin:
This is a PR to add clarity on pattern matching within function parameters. 

I'd like particular feedback on the following:
One thing I've seen in a few blog posts is that there's an order to the way incoming parameters are matched against the function argument.

For example, if we have 
```elixir
def hello(%{name: name} = person)
```
an incoming map would first match against person, then move on to matching `person` against `%{name: name}`. From what I could tell, the two parts of the argument are independently matching against incoming parameters; the order doesn't matter because they aren't pattern matching against each other. I would love to hear any clarifying thoughts on this part and if I've misunderstood.

For reference, an excerpt from one of the blogs:
> Elixir is pattern matching params first (the passed in map is the right-hand side, params becomes the left), then pattern matches user_id, as the left-hand side, against params which is now the right-hand side, like so.
-- [@turnandface - Pattern Matching in Elixir](https://medium.com/@turnandface/pattern-matching-in-elixir-743e71ceac92)

### DIFF
--- a/en/lessons/basics/functions.md
+++ b/en/lessons/basics/functions.md
@@ -113,6 +113,154 @@ iex> Greeter2.hello("Fred", "Jane")
 
 We've listed the function names in comments above. The first implementation takes no arguments, so it is known as `hello/0`; the second takes one argument so it is known as `hello/1`, and so on. Unlike function overloads in some other languages, these are thought of as _different_ functions from each other. (Pattern matching, described just a moment ago, applies only when multiple definitions are provided for function definitions with the _same_ number of arguments.)
 
+### Functions and Pattern Matching
+
+Behind the scenes, functions are pattern-matching the arguments that they're called with.
+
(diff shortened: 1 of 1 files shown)


michalvalasek:
Maybe we could put the advanced/in-depth info into som kind of boxes to make it clear for the beginners that this is something they can skip on the first pass and come back later?

kenspirit:
I really like the lesson on explaining on what pattern matching really is and how it works.  Understanding it help smooth a lot when reading other's Elixir code. It sure worths to have somewhere to have this.

I don't think this is advance if we do want to learn Elixir well.  It's basic.  But if it's lengthy, I think it might deserve a separate lesson.

@meryldakin, I always wonder there does have a order when the Elixir does pattern matching.  Because pattern matching always replace the variable with actual value on the right side of the = operator, and then try to find match for the variable on the left side of the =.

According to the booking "Programming Elixir >= 1.6"

>Elixir will only change the value of a variable on the left side of an equals sign—on the right a variable is replaced with its value.

PR #1641, opened 2018-12-03
add sophie to contributors

SophieDeBenedetto:


_data/contributors.yml
```diff
@@ -366,3 +366,16 @@
...
+  name: Sophie DeBenedetto
+  intro:
+    Sophie is an engineer and teacher at The Flatiron School. She loves teaching and learning and finding the Elixir School community felt like the perfect fit!
+  twitter: sm_debenedetto
+  homepage: https://thegreatcodeadventure.com
+  contributions:
+    - lesson
+    - blog
+  lessons: ['OTP Distribution', 'Plug']
+  articles:
```
doomspork: @burden I know it makes things a little crazy to unwire but could we figure this out automagically to avoid having to maintain a list of articles that might grow large?

PR #1652, opened 2018-12-11
Newline all sentences

doomspork:
@notactuallypagemcconnell I think we had decided to move forward with newlining sentences to avoid the 2 vs 1 space debate/discussion/changes (#1649 reminded me of it). This PR applies that change to all existing English lessons.

What do you think?


notactuallypagemcconnell:
💯 by me

PR #1671, opened 2019-01-02
Add StreamData lesson

michaelstalker:
I think I'm _finally_ done with this article :tada: I'd welcome feedback.

### DIFF
--- a/en/lessons/libraries/stream-data.md
+++ b/en/lessons/libraries/stream-data.md
@@ -9,35 +9,53 @@ An example-based unit testing library like [ExUnit](https://hexdocs.pm/ex_unit/E
 * You can write these tests without thinking through your requirements thoroughly.
 * These tests can be very verbose when you use several examples for one function.
 
-In this lesson we're going to explore how [StreamData](https://github.com/whatyouhide/stream_data) can help overcome some of these drawbacks.
-
+In this lesson we're going to explore how [StreamData](https://github.com/whatyouhide/stream_data) can help us overcome some of these drawbacks.
 
 {% include toc.html %}
 
-
 ## What is StreamData?
 
-[StreamData](https://github.com/whatyouhide/stream_data) is library that performs stateless property-based testing.
+[StreamData](https://github.com/whatyouhide/stream_data) is a library that performs stateless property-based testing.
 
-The StreamData library will run each test 100 times with random data. When a test fails, StreamData will try to shrink the input to the smallest value that causes the test failure. This can be helpful when you have to debug your code! If a 50-element list causes your function to break, and only one of the list elements is problematic, StreamData can help you identify the offending element.
+The StreamData library will run each test [100 times by default](https://hexdocs.pm/stream_data/ExUnitProperties.html#check/1-options), using random data each time. When a test fails, StreamData will try to [shrink](https://hexdocs.pm/stream_data/StreamData.html#module-shrinking) the input to the smallest value that causes the test failure. This can be helpful when you have to debug your code! If a 50-element list causes your function to break, and only one of the list elements is problematic, StreamData can help you identify the offending element.
 
 This testing library has two main modules. [`StreamData`](https://hexdocs.pm/stream_data/StreamData.html) generates streams of random data. [`ExUnitProperties`](https://hexdocs.pm/stream_data/ExUnitProperties.htm) lets you run tests against your functions, using the generated data as your input.
 
 You might be asking how you can say anything meaningful about a function if you don't know what your exact inputs are. Read on!
 
+## Installing StreamData
+
+First, create a new Mix project. Refer to [New Projects](https://elixirschool.com/en/lessons/basics/mix/#new-projects) if you need some help.
+
+Second, add StreamData as a dependency to your `mix.exs` file:
+
+```elixir
+defp deps do
+  [{:stream_data, "~> x.y", only: :test}]
+end
+```
+
+Just replace `x` and `y` with the version of StreamData shown in the library's [installation instructions](https://github.com/whatyouhide/stream_data#installation).
+
+Third, run this from the command line of your terminal:
+
+```
+mix deps.get
+```
 
 ## Using StreamData
 
 To illustrate the features of StreamData, we'll write a few simple utility functions that repeat values. Let's say we want a function like [`String.duplicate/2`](https://hexdocs.pm/elixir/String.html#duplicate/2), but one that will duplicate strings, lists, or tuples.
 
-
 ### Strings
 
-Let's duplicate strings first. What are some requirements for our function?
+First, let's write a function that duplicates strings. What are some requirements for our function?
 
 1. The first argument should be a string. This is the string that we'll duplicate.
-1. The second argument should be a non-negative integer. It shows how many times we'll duplicate the first argument.
-1. The function should return a string. This string zero or more of the is the first string, repeated over and over.
+2. The second argument should be a non-negative integer. It shows how many times we'll duplicate the first argument.
+3. The function should return a string. This new string is just the original string, repeated zero or more times.
+4. If the original string is empty, the returned string should also be empty.
+5. If the second argument is `0`, the returned string should be empty.
 
 When we run our function, we want it to look like this:
 
@@ -58,7 +76,6 @@ end
 
 The happy path should be easy to test with [ExUnit](https://hexdocs.pm/ex_unit/ExUnit.html).
 
-
 ```elixir
 defmodule RepeaterTest do
   use ExUnit.Case
@@ -73,8 +90,7 @@ end
 
 That's hardly a comprehensive test, though. What should happen when the second argument is `0`? What should the output be when the first argument is an empty string? What does it even mean to repeat an empty string? How should the function work with UTF-8 characters? Will the function still work with large input strings?
 
-You might say that my criticisms are unfair. After all, who would write just one example to cover this function? I'll grant your point. We could write more examples to test edge cases and large strings. However, let's see if we can use StreamData to test this function more rigorously without much more code.
-
+We could write more examples to test edge cases and large strings. However, let's see if we can use StreamData to test this function more rigorously without much more code.
 
 ```elixir
 defmodule RepeaterTest do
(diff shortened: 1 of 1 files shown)


en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+[StreamData](https://github.com/whatyouhide/stream_data) is library that performs stateless property-based testing.
+
+The StreamData library will run each test 100 times with random data. When a test fails, StreamData will try to shrink the input to the smallest value that causes the test failure. This can be helpful when you have to debug your code! If a 50-element list causes your function to break, and only one of the list elements is problematic, StreamData can help you identify the offending element.
+
+This testing library has two main modules. [`StreamData`](https://hexdocs.pm/stream_data/StreamData.html) generates streams of random data. [`ExUnitProperties`](https://hexdocs.pm/stream_data/ExUnitProperties.htm) lets you run tests against your functions, using the generated data as your input.
+
+You might be asking how you can say anything meaningful about a function if you don't know what your exact inputs are. Read on!
+
+
+## Using StreamData
```
SophieDeBenedetto: Can we add a bit on "getting started" that shoes you have to include the library in your mix file's dependencies?

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+title: StreamData
+---
+
+An example-based unit testing library like [ExUnit](https://hexdocs.pm/ex_unit/ExUnit.html) is a wonderful tool to help you verify that your code works the way you think it does. However, example-based unit tests have some drawbacks:
+
+* It can be easy to miss edge cases, since you're only testing a few inputs.
+* You can write these tests without thinking through your requirements thoroughly.
+* These tests can be very verbose when you use several examples for one function.
+
+In this lesson we're going to explore how [StreamData](https://github.com/whatyouhide/stream_data) can help overcome some of these drawbacks.
```
gemantzu: maybe `can help us overcome`?

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+
+
+{% include toc.html %}
+
+
+## What is StreamData?
+
+[StreamData](https://github.com/whatyouhide/stream_data) is library that performs stateless property-based testing.
```
gemantzu: is a library

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+    test "creates a new string, with the first argument duplicated a specified number of times" do
+      assert "aaaa" == Repeater.duplicate("a", 4)
+    end
+  end
+end
+```
+
+That's hardly a comprehensive test, though. What should happen when the second argument is `0`? What should the output be when the first argument is an empty string? What does it even mean to repeat an empty string? How should the function work with UTF-8 characters? Will the function still work with large input strings?
+
+You might say that my criticisms are unfair. After all, who would write just one example to cover this function? I'll grant your point. We could write more examples to test edge cases and large strings. However, let's see if we can use StreamData to test this function more rigorously without much more code.
```
gemantzu: Since this is not a blog post but a lesson, it might be better to remove the discussion part of this paragraph.

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+
```
gemantzu: remove second newline

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+        assert ??? == Repeater.duplicate(str, times)
+      end
+    end
+  end
+end
+```
+
+What does that do?
+
+* We replaced `describe` with [`property`](https://github.com/whatyouhide/stream_data/blob/v0.4.2/lib/ex_unit_properties.ex#L109). This lets us document the property we're testing.
```
gemantzu: you mean that you replaced `test` with `property`, right?

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+
+assert actual_length == expected_length
+```
+
+That's better than nothing, but it's not ideal. This test would still pass if our function generated random strings of the correct length.
+
+We really want to verify two things:
+
+1. Our function generates a string of the right length.
+1. The contents of the final string are the original string repeated over and over again.
```
gemantzu: 2.

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+  end
+end
+```
+
+We used `StreamData.list_of/1` and `StreamData.term/0` to create lists of random length, whose elements are any type.
+
+Like the property-based test for repeating strings, we compare the length of the new list with the product of the source list and `times`. The second assertion takes some explaining:
+
+1. We break the new list apart into several lists, each of which has the same number of elements as `list`.
+1. We then verify that each chunked list is equal to `list`.
```
gemantzu: 2.

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+  end
+end
+```
+
+...the StreamData version is actually shorter. StreamData also covers edge cases a developer might forget to test.
+
+
+### Lists
+
+Now, let's write a function that repeats lists. I want the function to work like this:
```
gemantzu: We want

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+  end
+end
+```
+
+How might we test this? Let's approach it a bit differently than we've done so far. For strings and lists, we asserted something about the length of the final data, and we asserted something about the contents of the data. Trying the same approach with tuples is possible, but the test code may not be as straightforward.
+
+Consider two sequences of operations you could perform on a tuple:
+
+1. Call `Repeater.duplicate/2` on the tuple, and convert the result to a list
+1. Convert the tuple to a list, and then pass the list to `Repeater.duplicate/2`
```
gemantzu: 2.

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+          |> Repeater.duplicate(times)
+
+        assert result_1 == result_2
+      end
+    end
+  end
+end
+```
+
+
```
gemantzu: remove second newline

gemantzu:
All in all, good job so far and welcome aboard. I have noted some minor typos / places that can be refined for you.

After reading it, the only part that got me thinking if it should change is the part where you wonder which testing methodology works best, but this might just be me.

I would also like to see a paragraph explaining that property testing is going to make the test suite slower.

en/lessons/libraries/stream-data.md
michaelstalker: Good catch! Those indefinite articles are important :grimacing:

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,397 @@
...
+To illustrate the features of StreamData, we'll write a few simple utility functions that repeat values. Let's say we want a function like [`String.duplicate/2`](https://hexdocs.pm/elixir/String.html#duplicate/2), but one that will duplicate strings, lists, or tuples.
+
+
+### Strings
+
+Let's duplicate strings first. What are some requirements for our function?
+
+1. The first argument should be a string. This is the string that we'll duplicate.
+1. The second argument should be a non-negative integer. It shows how many times we'll duplicate the first argument.
+1. The function should return a string. This string zero or more of the is the first string, repeated over and over.
```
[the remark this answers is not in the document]
michaelstalker: Yikes! My sentence wasn't even proper English. I'll fix this.

en/lessons/libraries/stream-data.md
michaelstalker: Thanks for the suggestion :+1: Would removing all sentences except the last two be better? I'll push up that change, and you can let me know what you think.

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,417 @@
...
+
+First, create a new Mix project. Refer to [New Projects](https://elixirschool.com/en/lessons/basics/mix/#new-projects) if you need some help.
+
+Second, add StreamData as a dependency to your `mix.exs` file:
+
+```elixir
+defp deps do
+  [{:stream_data, "~> x.y", only: :test}]
+end
+```
```
michaelstalker: This code snipped is mostly copied from the library's README. I linked to the README on line 38 below, but is any additional citation necessary here? I'm assuming not, since this is how you would install about any dependency, but please let me know if I should do something different. I don't want to plagiarize.

en/lessons/libraries/stream-data.md
```diff
@@ -0,0 +1,417 @@
...
+
+Just replace `x` and `y` with the version of StreamData shown in the library's [installation instructions](https://github.com/whatyouhide/stream_data#installation).
```
michaelstalker: This feels a little awkward to me, but I didn't want the lesson to go out of date if they released a new version of StreamData. What are your thoughts?

en/lessons/libraries/stream-data.md
notactuallypagemcconnell: I think with a link you are absolutely fine, the intent here is clear - way to be cautious/caring though!

en/lessons/libraries/stream-data.md
notactuallypagemcconnell: Great move IMO. I've thought about this for lessons before and didn't come up with this but it works for future proofing the lesson _and_ with how many translations we have, the work on updating just that version number compounds _quite_ quickly - so this seems minor but actually will save 10s of pull requests of the event of this being updated and assuming every language did a translation PR to keep it current.

en/lessons/libraries/stream-data.md
gemantzu: I would remove this part: `You might say that my criticisms are unfair. After all, who would write just one example to cover this function? I'll grant your point` and it's back to a lesson format :)

doomspork:
I am planning to release this Monday morning @michaelstalker 👍

PR #1690, opened 2019-01-15
fix spelling errors on distillery lesson

techgaun:
Saw couple of errors on spelling. Doing it from cellphone but hopefully everything is correct


doomspork:
Thank you @techgaun 😁

PR #1693, opened 2019-01-18
Typo in module name

kritik:
Greeter4 is not defined in this lesson and results which are returned match results from Greeter2


doomspork:
🎉 Thank you @kritik!

PR #1706, opened 2019-01-31
Don't suggest that the concept of an interface will be familiar to Ruby developers

timrogers:
The "Specifications and types" page compares Elixir's "specifications" to the concept of an "interface", suggesting that this might be a helpful analogy if you're familiar with Java or Ruby.

Ruby doesn't have a concept of interfaces in the language, so this seems like a bit of an odd thing to say. (Although the generic, abstract concept of an "interface", as in a "contract", is likely to be familiar to many programmers.)

This removes this reference to Ruby, just referring to Java instead.

### DIFF
--- a/en/lessons/advanced/typespec.md
+++ b/en/lessons/advanced/typespec.md
@@ -23,7 +23,7 @@ That means all information about a type will be ignored by the compiler, but cou
 
 ## Specification
 
-If you have experience with Java or Ruby you could think about specification as an `interface`.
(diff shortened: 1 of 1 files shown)


timrogers:
@doomspork Done!

doomspork:
Thank you very much @timrogers! 

If you're interested in helping us out more with Elixir School feel free to stop by our Slack: https://admissions.elixirschool.com/

PR #1718, opened 2019-02-13
Add excerpt to plug.router Heroku deployment post

notactuallypagemcconnell:
Self-explanatory. Excerpt is ```Want to put your app in the real world? Today we do it with Heroku!```


doomspork:
Thank you @notactuallypagemcconnell!

PR #1720, opened 2019-02-14
Retrieve all author's metadata from contributors.yml

tajchumber:
Retrieve the information of the author of a blog post from `contributors.yml`
if the author's website is unknow we fallback on their github page.

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -7,11 +7,7 @@ layout: default
 <section id="section-page">
   <header class="main">
     <h1 class="post-title">{{ page.title }}</h1>
-    {% if contributor.homepage %}
-      <p>{{ page.date | date_to_string }} · by <a href="{{ contributor.homepage }}">{{page.author}}</a></p>
-    {% else %}
-      <p>{{ page.date | date_to_string }} · by <a href="https://github.com/{{ contributor.github }}">{{page.author}}</a></p>
-    {% endif %}
+    <p>{% assign author_url = contributor.homepage | default: 'https://github.com/' | append: contributor.github %} {{ page.date | date_to_string }} · by <a href="{{ author_url }}">{{page.author}}</a></p>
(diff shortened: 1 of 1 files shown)


_layouts/post.html
```diff
@@ -1,10 +1,17 @@
...
 ---
+
+{% include function-getContributor type="name" search=page.author %}
+
 <section id="section-page">
   <header class="main">
     <h1 class="post-title">{{ page.title }}</h1>
-    <p>{{ page.date | date_to_string }} · by <a href="{{page.author_link}}">{{page.author}}</a></p>
+    {% if contributor.homepage %}
+      <p>{{ page.date | date_to_string }} · by <a href="{{ contributor.homepage }}">{{page.author}}</a></p>
```
doomspork: I _think_ we could do this without an `if` but I'm not sure that's necessarily more readable, what do you think?

```html
<p>
{% assign homepage = contributor.homepage | default: contributor.github %}
{{ page.date | date_to_string }} · by <a href="{{ homepage }}">{{page.author}}</a>
</p>
```

tajchumber: i'd call the variable `author_url`

but I prefer your version!

gemantzu: It seems cleaner, only thing is the way you do it, there will be a mistake in github url. You must build the github url in default (btw is there string interpollation in jekyll?).

tajchumber: i've committed the cleaner version 👍 thanks for the input all :)

tajchumber: > btw is there string interpollation in jekyll?).

@gemantzu [Liquid](https://shopify.github.io/liquid/) (the templating engine of Jekyll) has [append](https://shopify.github.io/liquid/filters/append/) and [prepend](https://shopify.github.io/liquid/filters/prepend/) functions

PR #1722, opened 2019-02-14
[zh-hans] Align ecto/changeset and add zh-hans contributors

kenspirit:


_data/contributors.yml
```diff
@@ -400,3 +400,23 @@
   homepage: https://kate-travers.com/
   contributions:
     - blog
+
+- github: zillou
```
doomspork: Thank you for doing this ❤️ 

Could you remove `intro`, `homepage`, and `twitter` as long as they are empty? I believe that will cause some issues.

kenspirit: It seems there are many other contributor section also have blank field as well, should I remove them all?

doomspork: Apparently we handle this in the templates, thanks for pointing that out @kenspirit 😁

zh-hans/lessons/ecto/changesets.md
```diff
@@ -213,7 +213,7 @@ end
 iex> User.changeset(%User{}, %{"name" => "Bob"})
 #Ecto.Changeset<
   action: nil,
-  changes: %{first_name: "Bob"},
+  changes: %{name: "Bob"},
```
doomspork: Care to do another PR applying this change to all languages impacted?

kenspirit: Will work on this after work tonight.

PR #1732, opened 2019-03-04
Reword confusing part in Associations lesson

brain-geek:
I was confused a lot when tried to read this first few times, so decided to improve it.

Also, current version states `Unlike adding the `movie_id` column to our `characters` table, this macro _doesn't_ add anything to the database.` While this is true, as in schema definitions don't change anything in database, it implies that we don't have `movie_id` field in schema, which is not correct - it's [defined by default](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/schema.ex#L1797) by `belongs_to` macros.

Edit: After the initial PR creaton I found some other places which could be refactored.

It's OK to squash these.

### DIFF
--- a/en/lessons/ecto/associations.md
+++ b/en/lessons/ecto/associations.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.1
+version: 1.1.0
 title: Associations
 ---
 
@@ -9,7 +9,7 @@ In this section we'll learn how to use Ecto to define and work with associations
 
 ## Set Up
 
-We'll be building off of the app, `Example`, from the previous lessons. You can refer to the setup [here](../basics) for a quick refresher.
+We'll start off with the `Example` app from the previous lesson. You can refer to the setup [here](../basics) for a quick refresher.
 
 ## Types of Associations
 
@@ -110,7 +110,7 @@ defmodule Example.Character do
 end
 ```
 
-Let's take a closer look at what the `belongs_to/3` macro does for us. Unlike adding the `movie_id` column to our `characters` table, this macro _doesn't_ add anything to the database. It _does_ give us the ability to access our associated `movies` schema _through_ `characters`. It uses the the foreign key of `movie_id` on the `characters` table to make a character's associated movie available when we query for characters. This is what will allow us to call `character.movie`.
+Let's take a closer look at what the `belongs_to/3` macro does for us. In addition to adding the foreign key `movie_id` to our schema, it also gives us the ability to access associated `movies` schema _through_ `characters`. It uses the foreign key to make a character's associated movie available when we query for them. This is what will allow us to call `character.movie`.
 
 Now we're ready to run our migrations:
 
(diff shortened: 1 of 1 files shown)


en/lessons/ecto/associations.md
```diff
@@ -9,7 +9,7 @@ In this section we'll learn how to use Ecto to define and work with associations
 
 ## Set Up
 
-We'll be building off of the app, `Example`, from the previous lessons. You can refer to the setup [here](../basics) for a quick refresher.
+We'll start off the `Example` app from the previous lesson. You can refer to the setup [here](../basics) for a quick refresher.
```
doomspork: "We'll start off with the"

en/lessons/ecto/associations.md
```diff
@@ -110,7 +110,7 @@ defmodule Example.Character do
 end
 ```
 
-Let's take a closer look at what the `belongs_to/3` macro does for us. Unlike adding the `movie_id` column to our `characters` table, this macro _doesn't_ add anything to the database. It _does_ give us the ability to access our associated `movies` schema _through_ `characters`. It uses the the foreign key of `movie_id` on the `characters` table to make a character's associated movie available when we query for characters. This is what will allow us to call `character.movie`.
+Let's take a closer look at what the `belongs_to/3` macro does for us. In addition to adding foreign key `movie_id` to schema, it also gives us the ability to access associated `movies` schema _through_ `characters`. It uses the foreign key to make a character's associated movie available when we query for them. This is what will allow us to call `character.movie`.
```
doomspork: How about: "In addition to adding the foreign key `movie_id` to our schema,"

brain-geek:
@doomspork Feel free to squash these, I shouldn't add anything more here.

en/lessons/ecto/associations.md
```diff
@@ -177,7 +177,7 @@ defmodule Example.Movie do
 end
 ```
 
-The `has_one/3` macro functions just like the `has_many/3` macro. It doesn't add anything to the database but it _does_ use the associated schema's foreign key to look up and expose the movie's distributor. This will allow us to call `movie.distributor`.
+The `has_one/3` macro functions just like the `has_many/3` macro. It allows use the associated schema's foreign key to look up and expose the movie's distributor. This will allow us to call `movie.distributor`.
```
doomspork: "It uses the associated schema's foreign key to"

en/lessons/ecto/associations.md
brain-geek: Yeah, thanks! I spent like 15 minutes trying to phrase the best option here.

brain-geek:
Thanks for review, @doomspork . Applied and squashed everything as well.

doomspork:
You're awesome @brain-geek! Hope all is well with you and yours. Always a pleasure to see you around 😁 

You should join us on Slack if you haven't already https://admissions.elixirschool.com 🙂

brain-geek:
Thanks, joined!

PR #1734, opened 2019-03-04
Whitespace/tiny fixes

brain-geek:
Once again as in #712 , I'm going through project texts via `proselint`.

Mostly, whitespace (double-space or trailing space) fixes, typos (i.e. capital letter) and small readability changes (i.e. merging two related paragraphs, removing `very`, etc).

I'm actually not sure whether should I update versions or not, as these changes don't affect the actual texts.

PR #1737, opened 2019-03-06
[JA] Translate Ecto/Querying

Tsuyoshi84:
I translated Ecto/Quering page.
Please review! @elixirschool/japanese


ja/lessons/ecto/querying.md
```diff
@@ -0,0 +1,443 @@
+---
+version: 1.0.0
+title: クエリ
+---
+
+{% include toc.html %}
+
+このレッスンでは、 `Example` アプリと [前のレッスン](./associations) で設定した映画カタログドメインを作成します。
```
Tsuyoshi84: I translated the sentence `movie-cataloguing domain` as `映画カタログドメイン`, but there should be a better translation I think. Let me know if you have any idea!

riseshia:
I wil check until this weekend :bow:

ja/lessons/ecto/querying.md
```diff
@@ -0,0 +1,443 @@
+---
+version: 1.0.0
```
riseshia: We may say this v1.0.1 https://github.com/elixirschool/elixirschool/pull/1734/files#diff-42858e1c36d5073e0f1f6c9b1b86d979 (It seems already fixed / no need to fix in Japanese..?

ja/lessons/ecto/querying.md
```diff
@@ -0,0 +1,443 @@
...
+
+```elixir
+iex> Movie |> last() |> Repo.one()
+```
+
+## 関連したデータのクエリ
+
+### プリロード
+
+`belongs_to` 、 `has_many` 、そして `has_one` マクロが公開する関連レコードにアクセスできるようにするために、関連したスキーマを _プリロード_ する必要があります。
```
riseshia: (maybe) マクロが公開する関連レコード -> マクロからの公開する関連レコード

ja/lessons/ecto/querying.md
riseshia: I have no better idea too. 🙈 
or 映画ドメイン may be enough in this case..

(maybe) 設定した映画カタログドメインを作成します -> 設定した映画カタログドメインを利用します

ja/lessons/ecto/querying.md
Tsuyoshi84: Oops, I didn't realize that the version has increased. It looks like the change has only removed unnecessary spaces, so there is nothing we have to do but just bump up the version I guess.

ja/lessons/ecto/querying.md
Tsuyoshi84: OK, then I'll just translate it as `映画ドメイン`!
`映画カタログドメイン` sounds kind of strange and confusing so it should be avoided anyway 🤔

riseshia:
🎉

PR #1739, opened 2019-03-10
Update theme toggler

tajchumber:
Adding `title` attribute to the `a` element.

This "Toggle Theme" will show in the form of a tooltip on mouse hover


doomspork:
Thank you @tajchumber 😁

PR #1753, opened 2019-03-20
fix: #main div stretching too much

### LINKED ISSUE #1751: Blog: Burger icon not opening sidepanel in blog
This is a weird situation, where If I close the sidepanel when I am at the index page, and click on a blog post below, when I am in the blog, the burger icon does not reshow the sidepanel (It tries to open it but fails). It is showing on both Firefox and Chromium on my system. On my android phone, it is working as intended.

tajchumber:
#main div stretches too much when very long strings without breaks are present, making the sidebar disappear out of view

Fixes #1751

### DIFF
--- a/_assets/stylesheets/layout/_main.scss
+++ b/_assets/stylesheets/layout/_main.scss
@@ -3,7 +3,7 @@
 	#main {
 		@include vendor('flex-grow', '1');
 		@include vendor('flex-shrink', '1');
-		width: 100%;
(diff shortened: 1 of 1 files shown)


gemantzu:
Thanks @tajchumber !!! The only remnant of this I can still see is the first time you open the specific blog post and try to close the side panel, it insta-closes instead of doing the animation one. If you open another one, it works properly.

PR #1760, opened 2019-03-24
`phx.gen.secret` generates a secret, not `phx.secret`

mjparrott:
Looks like the correct mix tasks for this is `phx.gen.secret`, not `phx.secret`. See https://hexdocs.pm/phoenix/phoenix_mix_tasks.html


doomspork:
Thank you @mjparrott!

PR #1762, opened 2019-03-25
Fix code snippets in Phoenix Live View post

mjparrott:
A couple of missing `)`s in the code in the post.

Also added instructions to run `npm install` after adding `phoenix_live_view` as a dependency, otherwise Phoenix can't find it when running the application.


doomspork:
Thank you so much for doing this @mjparrott! If you're interested in joining us on Slack feel free to head over to https://admissions.elixirschool.com, you're contributions will grant you access 😁

PR #1774, opened 2019-04-04
Fix typo in ecto doc

### LINKED ISSUE #1758: Compile error when running the migrations for the lesson: Ecto Associations - The Belongs To Migration
When we run the migrations for the lesson [Ecto association - The Belongs To Migration](https://elixirschool.com/en/lessons/ecto/associations/#the-belongs-to-migration) we get this error:

```bash
 ╭─elixir@6da418c1ce55 ~/workspace/friends  ‹associations*› 
╰─$ mix ecto.migrate                                                                                                                              
** (CompileError) priv/repo/migrations/20190323130013_create_characters.exs:5: undefined function create_table/2
    (elixir) src/elixir_locals.erl:107: :elixir_locals."-ensure_no_undefined_local/3-lc$^0/1-0-"/2
    (elixir) src/elixir_locals.erl:108: anonymous fn/3 in :elixir_locals.ensure_no_undefined_local/3
    (stdlib) erl_eval.erl:680: :erl_eval.do_apply/6
```

This happens because the migration code uses `create_table` instead of `create table`:

```elixir
# priv/migrations/*_create_characters.exs
defmodule Example.Repo.Migrations.CreateCharacters do
  use Ecto.Migration

  def change do
    create_table(:characters) do
      add :name, :string
      add :movie_id, references(:movies)
    end
  end
end
```

lccezinha:
Closes #1758

### DIFF
--- a/en/lessons/ecto/associations.md
+++ b/en/lessons/ecto/associations.md
@@ -72,7 +72,7 @@ First, generate the migration:
 mix ecto.gen.migration create_characters
 ```
 
-To declare that a character belongs to a movie, we need the `characters` table to have a `movie_id` column. We want this column to function as a foreign key. We can accomplish this with the following line in our `create_table/1` function:
+To declare that a character belongs to a movie, we need the `characters` table to have a `movie_id` column. We want this column to function as a foreign key. We can accomplish this with the following line in our `create table/1` function:
 
 ```elixir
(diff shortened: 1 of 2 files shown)


SophieDeBenedetto:
Thanks for catching this @lccezinha!

PR #1778, opened 2019-04-05
Fix LiveView post 

### LINKED ISSUE #1777: LiveView blog post fixes
1. fix the numbering on the installation steps
2. change the file name to 2019, it is 2018
3. `MyAppWeb.GithubDeployView.handle_info/2` matching on {:push_contents, repo} has probably a wrong variable assignment, as below we use the var `org` which is not assigned anywhere.
4. You have to make it more specific that the old template (`page/index.html.eex`) must be removed before this will be able to work, as the way it is now, with both files in the folder, it throws an error.

SophieDeBenedetto:
closes #1777

### DIFF
--- a/_posts/2019-03-18-phoenix-live-view.md
+++ b/_posts/2019-03-18-phoenix-live-view.md
@@ -37,7 +37,7 @@ Here's the functionality we're building:
 
 The following steps are detailed in Phoenix LiveView [Readme](https://github.com/phoenixframework/phoenix_live_view).
 
-1. Install the dependency in your `mix.exs` file:
+* Install the dependency in your `mix.exs` file:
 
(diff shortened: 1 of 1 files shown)


gemantzu:
@SophieDeBenedetto do you think it would make sense to also add the very trivial part of the MyApp functions, just for the sake of clarity / completion? SInce we are targeting people who are still learning, it might lead to someone having issues. It would probably be helpful to also add a small timeout to each one of them, so someone can see the button changes happening. What do you think?

doomspork:
@gemantzu if this is an already published blog post, let's not go back and make changes to it, these aren't likely read like lessons and there's no way to track a blog has been updated since you lost read it.

PR #1788, opened 2019-04-09
Localize `Menu` heading

lex111:


doomspork:
Sounds like a winner to me! I'm going to give others a chance to look this over but I'm all for this 💯

PR #1792, opened 2019-04-09
Upgrade benchee lessons for 1.0 + some additional content

PragTob:
* Most changes are for benchee 1.0 compatibility + updated
  output
* Added a section about memory measurements as I think they
are important
* added lots of more configuration options that we added over
time didn't want to fully explain all of them and referred to the
README for them. Would be worth considering not enumerating all
of them but just the most crucial ones and refer to the README
for others (I'd leave parallel out for instance), but didn't want
to break with the existing content for now

Was unsure as to what version bump this would be considered as
it's neither just a few sentences nor a full rewrite but as it
needs translator attention and better safe than sorry I opted for
the major increase.


doomspork:
Thank you @PragTob!!

doomspork:
This is _awesome_ thank you for coming back to this lesson @PragTob 😁

PragTob:
Thanks for doing elixir school :dancer:

PR #1793, opened 2019-04-09
Fix icon class for last commit on Report page

lex111:
| Before   | After    |
| -------- | -------- |
|image|image|

Typo :)


doomspork:
This is awesome @lex111! We still need to do some tweaking of that last commit. If memory serves me, @gemantzu found that on Heroku we're not actually getting the full history and so the last commit is wrong 😢

gemantzu:
Yes this is correct. Do you want me to disable it?

PR #1794, opened 2019-04-10
[RU] index: improvements

lex111:


doomspork:
@elixirschool/russian I am going to merge these since it's been a few days and no comment. If there's issues we can follow-up with new PRs 👍

PR #1796, opened 2019-04-10
Remove Google+ share icon

lex111:
As Google+ doesn't exist anymore, the link in social sharing list should be removed.


doomspork:
Thank you @lex111 😁

PR #1803, opened 2019-04-12
Add GitHub star button

lex111:
image

Do you think is it better to hide the counter?

### DIFF
--- a/_includes/share-icons.html
+++ b/_includes/share-icons.html
@@ -4,6 +4,7 @@
   {% else %}
     {% assign title = page.title %}
   {% endif %}
+  <li><iframe src="https://ghbtns.com/github-btn.html?user=elixirschool&repo=elixirschool&type=star&count=true" height="20" title="GitHub Stars" width="93" style="vertical-align: sub;"></iframe></li>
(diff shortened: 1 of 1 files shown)


doomspork:
@lex111 I think the star count is fine. It may help folks notice the addition too. 

Wonderful series of improvements, I'm so grateful ❤️

doomspork:
@lex111 since you're clearly interested in helping with more than the translations, I'll add you to the @elixirschool/developers team as well. We really appreciate all you've done 😁

lex111:
It would be cool if the ElixirSchool work on an Elixir-based site generator, rather than on Jekyll. I was... I was thinking aloud :smiley:

doomspork:
@lex111 we've talked about this. Unfortunately, the existing Elixir static site generators are not supported and would require a fair amount of work to get to the point that we could swap them in for Jekyll.

We've also kicked around the idea of using Phoenix and handling the generation of pages from the markdown, something akin to https://github.com/thoughtbot/high_voltage

lex111:
Yeah, I understand it all, It'd be  very cool :heart_eyes:

PR #1808, opened 2019-04-14
Add link rel canonical

lex111:
According to https://support.google.com/webmasters/answer/139066?hl=en

### DIFF
--- a/_includes/_head.html
+++ b/_includes/_head.html
@@ -26,6 +26,7 @@
     {% endif %}
   {% endfor %}
 
+  <link rel="canonical" href="{{ site.url }}{{ page.url }}">
   <link rel="apple-touch-icon" sizes="180x180" href="{{ site.url }}{% asset favicons/apple-touch-icon.png @path %}">
   <link rel="icon" type="image/png" sizes="32x32" href="{{ site.url }}{% asset favicons/favicon-32x32.png @path %}">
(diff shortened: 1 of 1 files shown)


doomspork:
@lex111 you're the man with a plan! I need to level up my SEO skills, any suggestions on where I could do that?

lex111:
@doomspork ha ha, I'm glad to hear that, but in fact, SEO is already good! Seems to me we need to improve the print version.

P.S. Found one "bug", why is the `gr` prefix used for the Greek translation, when should there be an `el` (see https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)?

lex111:
Oh, that's right! https://en.wikipedia.org/wiki/ISO_3166-1

However, for `hreflang` attribute we need to use the language code according to ISO 639, that is: `el`!

lex111:
Quick fix! https://github.com/elixirschool/elixirschool/pull/1809

gemantzu:
@lex111 you are asking the ancient questions here man, the cannon name for our country is Hellas, we call it Hellada (from our own Ελλάδα), but our annointed name `Greece` and we as `Greeks` from the ancient times seem to be stuck with us [check here](https://en.wikipedia.org/wiki/Name_of_Greece). The cannonical abbreviation should always be 'el', but these things are tough to change all over the place.

PR #1806, opened 2019-04-14
Fixing some commentaries that were meant to be code

ogabriel:
the codes of creation of the structs were like comments

PR #1819, opened 2019-04-23
Add CODEOWNERS to enable auto assignment reviews

lex111:
See https://help.github.com/en/articles/about-code-owners

Inspired by [Ruby team](https://github.com/ruby/www.ruby-lang.org/blob/master/.github/CODEOWNERS) 💎

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,39 @@
+# Order alphabetically.
+# Order is important. The last matching pattern takes the most precedence.
+
+# Default owners for everything in the repo.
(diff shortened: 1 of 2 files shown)


lex111:
@doomspork Yeah! I did not know about it either!

P.S. I don't get why that's a problem. https://github.com/elixirschool/elixirschool/pull/1806
We’re leave this PR for too long, and I don’t understand **why**.

doomspork:
@lex111 I responded. It's been a hectic couple of weeks. What you and @gemantzu said is good 😁

PR #1820, opened 2019-04-25
Fix Chinese teams in CODEOWNERS file

lex111:
My fault :(


doomspork:
I should have noticed, that's on me. Thanks for fixing this @lex111 😁

PR #1821, opened 2019-04-25
Update functions.md

quentusrex:
Fixed typo in error output


lex111:
@doomspork wait, but shouldn't we fix this in the appropriate translation files?

doomspork:
@lex111 yes sir, that's preferable. In this case (or the event the PR author doesn't have time to do it), we can just do a follow-up PR with those changes.

I can do that or you can 😁

lex111:
@doomspork stay back, I'll do it! :smiley:

doomspork:
Thank you again @quentusrex, @tajchumber, and @lex111 😁

PR #1830, opened 2019-05-02
[zh-hans] Fix wrong copy of example code

Maples7:


doomspork:
Thank you so much @Maples7 🎉

PR #1845, opened 2019-05-16
update testing.md

Tass-Jomo:
That was a grammatical error.


doomspork:
Thank you @Tass-Jomo and @tajchumber 😁

PR #1849, opened 2019-05-23
Provide more useful exampels

rustkas:
I think that using data which can not be changed is not clearly explain ideas of described rules. When we using a function for describing interpolation it is a good idea to use data that can be changed when that function was implemented. 
I offer a clearer explanation in the form of an improvement in the example used.


doomspork:
@rustkas welcome and great catch! Would you be so kind as to apply this change to all the translations where application?

rustkas:
Yes. I will add that change to all the translations where application.

doomspork:
Thank you @rustkas!

PR #1858, opened 2019-05-23
Provide more useful exampels

rustkas:
change input data of String.downcase


doomspork:
Thank you @rustkas and @julienXX 😁

PR #1968, opened 2019-09-01
Change uniq_by to simpler uniq

kodnin:
`uniq/1` is simpler than `uniq_by/2`, though the latter is more flexible. I think simplicity wins for these guides. We can also mention both.

### DIFF
--- a/en/lessons/basics/enum.md
+++ b/en/lessons/basics/enum.md
@@ -195,11 +195,11 @@ iex> Enum.sort([%{:count => 4}, %{:count => 1}])
 [%{count: 1}, %{count: 4}]
 ```
 
-### uniq_by
+### uniq
 
-We can use `uniq_by/2` to remove duplicates from our enumerables:
+We can use `uniq/1` to remove duplicates from our enumerables:
 
 ```elixir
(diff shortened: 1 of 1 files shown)


en/lessons/basics/enum.md
```diff
@@ -195,11 +195,11 @@ iex> Enum.sort([%{:count => 4}, %{:count => 1}])
 [%{count: 1}, %{count: 4}]
 ```
 
-### uniq_by
```
vbrazo: Why don't you add both?

tajchumber: @kodnin I agree with @vbrazo here
Depending on the use case you might choose which function works best for you :)

kodnin:
@vbrazo and @tajchumber, I updated the PR to include both.

kodnin:
@doomspork, that's exactly what I did. What should be changed?

doomspork:
Sorry @vbrazo the diff looked weird, on second glance I see you did add both. Thanks again for all your help 😁 

BTW have you joined us on Slack? Head over to http://admissions.elixirschool.com to request an invite. The source for Admissions can be found here if you're curious: https://github.com/elixirschool/admissions

PR #1973, opened 2019-09-03
Improve doctest line

kodnin:
Small improvement for the line where doctests are introduced.


tajchumber:
i've bumped the version @doomspork

PR #1974, opened 2019-09-05
Textual updates to Strings chapter

kodnin:
Some small textual updates to the Strings chapter.


doomspork:
Thank you @kodnin and @vbrazo

PR #1992, opened 2019-09-11
Releasing an umbrella app with mix release, config and docker

SophieDeBenedetto:
Super rough draft! One big question that I outlined below and i still need to write a conclusion.

### DIFF
--- a/_posts/2019-09-15-releasing-an-umbrella-app-with-docker-and-mix-release.md
+++ b/_posts/2019-09-15-releasing-an-umbrella-app-with-docker-and-mix-release.md
@@ -215,8 +215,6 @@ COPY mix.* /app/
 
 COPY apps/registrar/mix.exs /app/apps/registrar/
 COPY apps/registrar_web/mix.exs /app/apps/registrar_web/
-COPY apps/registrar_ipc/mix.exs /app/apps/registrar_ipc/
-COPY apps/registrar_workers/mix.exs /app/apps/registrar_workers/
 
 ENV MIX_ENV=prod
 RUN mix do deps.get --only $MIX_ENV, deps.compile
--- a/_posts/2019-09-15-releasing-an-umbrella-app-with-docker-and-mix-release.md
+++ b/_posts/2019-09-15-releasing-an-umbrella-app-with-docker-and-mix-release.md
@@ -179,7 +179,7 @@ end
 
 While developers often use user-defined module attributes as constants, its important to remember that _the value is read at compilation time and not at runtime._ Since the value of `Application.get_env(:registrar, :stripe_api_base_url)` (which comes from a system environment variable) is only present at _runtime_, using a module attribute here won't work!
 
-Instead, we'll use a function to dyanmically look up the value at runtime:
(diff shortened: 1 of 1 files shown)


_posts/2019-09-15-releasing-an-umbrella-app-with-docker-and-mix-release.md
```diff
@@ -0,0 +1,376 @@
...
+
+Recall that we're starting our release in `./bin/start` with that `start` command:
+
+```
+bin/registrar_umbrella start
+```
+
+This will execute the `rel/env.sh` file in turn. This file should contain a script that does the following:
+
+* If the command given to the release was `start`, run the migrations using `eval`.
```
SophieDeBenedetto: On the one hand, since `env.sh` runs when the release is started (for example via the `start` command) this feels like it _could_ be the right place to execute migrations. However, since a call to `eval` _also_ starts the release running, that would execute `env.sh` _again_, hence the need for the `if` condition below. So I'm not sure if running migrations belongs here or if it would be better to put in in `./bin/start` just before the call to `bin/RELEASE_NAME start`. Does anyone have an opinion here? The docs didn't really clear it up for me.

_posts/2019-09-15-releasing-an-umbrella-app-with-docker-and-mix-release.md
```diff
@@ -0,0 +1,376 @@
...
+
+  def get(url) do
+    HTTPoison.get(@stripe_api_base_url <> url)
+  end
+end
+```
+
+While developers often use user-defined module attributes as constants, its important to remember that _the value is read at compilation time and not at runtime._ Since the value of `Application.get_env(:registrar, :stripe_api_base_url)` (which comes from a system environment variable) is only present at _runtime_, using a module attribute here won't work!
+
+Instead, we'll use a function to dyanmically look up the value at runtime:
```
vbrazo: typo: `dyanmically` should be `dynamically`

_posts/2019-09-15-releasing-an-umbrella-app-with-docker-and-mix-release.md
```diff
@@ -370,5 +370,6 @@ fi
 
 And that's it!
 
-
 ## Conclusion
+
+With Elixir 1.9, we can build a release without the addition of any external dependencies––Elixir now natively provides us everything we need. We can configure multiple releases for our umbrella app, definining which child apps to start for a given release. We can configure runtime vs. build time environment variables _and_ we can even define customized start up scripts to do things like run our migrations. All in all, `mix release` provides us with a comprehensive and powerful set of tools.
```
vbrazo: typo: `definining` should be `defining`

_posts/2019-09-15-releasing-an-umbrella-app-with-docker-and-mix-release.md
doomspork: I don't have any thoughts. Is there anyone we could reach out to? Or maybe it might be worth putting a note into the blog post about this and solicit feedback?

PR #2008, opened 2019-10-01
Specify that Map update operator syntax only works for existing keys

delta1:
Add a note and example that Map update operator only works to update an _existing_ key, and an example to use Map.put/3 for a new key instead.

### DIFF
--- a/en/lessons/basics/collections.md
+++ b/en/lessons/basics/collections.md
@@ -185,3 +185,19 @@ iex> map = %{foo: "bar", hello: "world"}
 iex> %{map | foo: "baz"}
 %{foo: "baz", hello: "world"}
 ```
+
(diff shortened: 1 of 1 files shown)


delta1:
hey @doomspork ! it's a pleasure, thanks for this great resource and for the feedback and the warm welcome :D 

I have bumped the version - hope that helps!

doomspork:
Thank you @delta1 ❤️ both of your PRs were great additions. I hope you'll stick around and continue your awesome contributions.

If you're interested in joining us on Slack you can grab an invite over at http://admissions.elixirschool.com though you may need to give GitHub some time to update (we check for contributions to eliminate spam).

delta1:
Hey @doomspork you're most welcome, thank you too for the friendly atmosphere and this great project! I don't use slack all that much but I'll try hop on occasionally - and I'll definitely be around to contribute some more :)

PR #2029, opened 2019-10-06
[PT] Update Guardian 1.0.3

marcsvll:


pt/lessons/libraries/guardian.md
```diff
@@ -207,13 +242,15 @@ end
...
+Fazer o login e logout de uma sessão do browser é muito simples. 
+No controller de login:
 
 ```elixir
 def create(conn, params) do
   case find_the_user_and_verify_them_from_params(params) do
     {:ok, user} ->
-      # Use access tokens. Other tokens can be used, like :refresh etc
+      # Use access tokens. 
+      #Other tokens can be used, like :refresh etc
```
anderkonzen: Minor typo: missing space

marcsvll: Hi @anderkonzen! 
I used the [English](https://github.com/elixirschool/elixirschool/blob/master/en/lessons/libraries/guardian.md) text as a reference.

PR #2050, opened 2019-10-10
Add a unique index for the one to many relation

rmoorman:
It could make sense to include a unique index on :movie_id on the distributors table as there may only be one for each movie.


doomspork:
Thank you and welcome @rmoorman! As @tajchumber mentioned, could you jump the version at the top of the file from 1.1.0 to 1.2.0? This helps the translators keep track of changes

rmoorman:
@tajchumber like this? If there is something else I can/need to do, please let me know :-)

rmoorman:
You are very welcome!

PR #2053, opened 2019-10-11
Swap "pair" for "my_app"

mertonium:
I was following along with this tutorial and got a compilation warning because the my config referenced an application (`pair`) that was not available.
I assume the basis for this post was an app called "pair" and this was just leftover 😄 

(I also updated a reference in the example output just for completeness.)

Thanks for your work!


tajchumber:
thanks for fixing this @mertonium 
now that you have contributed to `elixirschool` you can join our slack channel here: https://admissions.elixirschool.com

PR #2054, opened 2019-10-11
Corrected example for change/2

emjwivell:
Example referenced change/2, but example was showing cast/4.


doomspork:
Thank you @emjwivell!

PR #2055, opened 2019-10-12
en/lessons/ecto/querying.md > fix few minor issues with code listings

martosaur:
I was translating this article and noticed few issues with examples:

1. 
>We use the `Ecto.Query.select/3` function to specify the select statement portion of our query. If we want to select only certain fields, we can specify those fields as a list of atoms or by referencing the struct's keys. Let's take a look at the first approach

here we talk about a _function_ `Ecto.Query.select/3` and that we can pass a list of keywords... but in example it's actually a `from` macro with keyword query:
```
iex> query = from(Movie, select: [:title])
#Ecto.Query<from m in Example.Movie, select: [:title]>
```
2. 
>The following query will preload associated records in a _separate_ query.

>We can see that the above line of code ran _two_ database queries. One for all of the movies, and another for all of the actors with the given movie IDs.

We're talking about making DB queries and exactly at this moment debug logs are turned off in example. So I'm adding those.

### DIFF
--- a/en/lessons/ecto/querying.md
+++ b/en/lessons/ecto/querying.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.1
+version: 1.0.2
 title: Querying
 ---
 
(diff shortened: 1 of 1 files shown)


lex111:
@SophieDeBenedetto, please take a look at this :arrow_up: .

PR #2061, opened 2019-10-14
[en] Update ecto to 3.x

marocchino:


tajchumber:
thanks @marocchino ! :)

PR #2065, opened 2019-10-16
[PT] Update basics/collections (to 1.3.0)

g13ydson:
- Add newlines
- Add new content

### DIFF
--- a/pt/lessons/basics/collections.md
+++ b/pt/lessons/basics/collections.md
@@ -1,5 +1,5 @@
 ---
-version: 1.2.3
+version: 1.3.0
 title: Coleções
 ---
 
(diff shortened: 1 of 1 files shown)


pt/lessons/basics/collections.md
```diff
@@ -172,3 +172,19 @@ iex> map = %{foo: "bar", hello: "world"}
 iex> %{map | foo: "baz"}
 %{foo: "baz", hello: "world"}
 ```
+
+**Nota**: esta sintaxe funciona apenas para atualizar uma chave que já exista no mapa! Se a chave não existir, um `KeyError` será gerado.
```
anderkonzen: I think "... uma chave que já existe no mapa! ..." would translate better.

g13ydson: You're right, @anderkonzen. I made the corrections. Thanks: :+1:

Issue #2071, opened 2019-10-20
Syntax issue for output in Structs examples


MrL1605:
There is issue in some of examples of Structs where the output starts with `#` instead of `%`.
Eg. in [Module.md Line 86](https://github.com/elixirschool/elixirschool/blame/master/en/lessons/basics/modules.md#L86)
```elixir
#Example.User<name: "Sean", roles: [], ...>
```
should be
```elixir
%Example.User<name: "Sean", roles: [], ...>
``` 
Highlighter also shows error as shown below in the screenshot.

![Screenshot from 2019-10-20 18-35-13](https://user-images.githubusercontent.com/7734245/67160511-3786a080-f36f-11e9-8178-45dfd648d6bc.png)

Also, let me know If I can contribute.
I would like to fix this as part of Hacktober Fest.

doomspork:
This is great @MrL1605! Please do contribute a fix, we'd love to have your help ❤️ 

If this change can be applied to translated lessons, let's do that as well 😁

MrL1605:
Sure, I would do it for all languages.

Also, I did a grep and also found same issue in all mentions of [Ecto -> Changesets](https://github.com/elixirschool/elixirschool/blame/master/en/lessons/ecto/changesets.md#L21).
Should I fix that as well?

doomspork:
That'd be awesome @MrL1605! In the interest of Hacktoberfest if you wanted to submit this as 2 different PRs I'd be happy to accept both:

1. Correcting the `basics/module` lesson across all languages
2. Correcting the `changeset` lesson across all languages

MrL1605:
> if you wanted to submit this as 2 different PRs I'd be happy to accept both:

That would be too easy. I would like to earn the T-shirt the hard way. :smiling_imp:

SophieDeBenedetto:
Hey @MrL1605! Thanks so much for your contributions 🚀 🚀 🚀 
Since this issue seems to be resolved by #2073, I'm going to close this issue.

PR #2073, opened 2019-10-20
Fixed output of IEx for Structs from # to %

MrL1605:
Fixed the output of iex where Structs were prefixed with `#` instead of `%` in some files.

Related Issue: #2071 

List of structs that I found had the problem.
 - Example.User
 - Echo.Changeset
 - Ecto.Schema.Metadata
 - Ecto.Association.NotLoaded
 - Ecto.Query
 - Friends.Person
 - Example.Movie
 - Example.Actor
 - MyCoolWebApp.Accounts.User

I think I have covered all of them, let me know if there are any missing.
Also I have not tried each one of them in IEx, but assumed every Struct needs this fix. 
Mostly because I'm yet to complete my basics tutorials, so didn't know how to import Ecto.

Made sure to update the version of every file edited. 
Incremented only the PATCH part of it.

> Note: I have also edited posts, but I'm not sure If I should do that.

### DIFF
--- a/_posts/2018-10-10-ecto-multi.md
+++ b/_posts/2018-10-10-ecto-multi.md
@@ -19,7 +19,7 @@ Everything starts with a `%Multi{}` struct. You can create a new Multi calling t
 
 ```elixir
 iex> Ecto.Multi.new()
-%Ecto.Multi{names: #MapSet<[]>, operations: []}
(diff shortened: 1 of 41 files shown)


doomspork:
This is great, thank you @MrL1605!

PR #2094, opened 2019-10-24
Add a small reference to the pipe operator

Snackomat:
One thing that really bothered me throughout learning is that the pipe operator wasn't mentioned during the first few chapters in the basics. Therefore I propose a change in order to make the lives of newcomers easier. This change introduces a reference to the pipe operator which should be enough to get newcomers to understand what's going on.

### DIFF
--- a/en/lessons/basics/basics.md
+++ b/en/lessons/basics/basics.md
@@ -101,6 +101,7 @@ false
 ```
 
 The booleans `true` and `false` are also the atoms `:true` and `:false`, respectively.
+*The pipe operator |> passes the result of an expression as the first parameter of another expression. We will explore this more in-depth in chapter 7: "Pipe Operator"*
 
 ```elixir
 iex> true |> is_atom
--- a/en/lessons/basics/basics.md
+++ b/en/lessons/basics/basics.md
@@ -101,12 +101,11 @@ false
 ```
 
 The booleans `true` and `false` are also the atoms `:true` and `:false`, respectively.
-*The pipe operator |> passes the result of an expression as the first parameter of another expression. We will explore this more in-depth in chapter 7: "Pipe Operator"*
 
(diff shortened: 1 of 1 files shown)


en/lessons/basics/basics.md
```diff
@@ -101,6 +101,7 @@ false
 ```
 
 The booleans `true` and `false` are also the atoms `:true` and `:false`, respectively.
+*The pipe operator |> passes the result of an expression as the first parameter of another expression. We will explore this more in-depth in chapter 7: "Pipe Operator"*
```
doomspork: I'm wondering if we should change these to `is_atom(true)` and `is_atom(false)` since we often discourage using pipes for a single function calls anyways and instead add a small section on the pipe operator in this lesson.

What do you think @Snackomat?

Snackomat: I think that would really help newcomers aswell, because (atleast for me) that would've clarified it way more and also suits the way I expect a language to behave.

Snackomat:
I do also propose removing pipe operators from chapter 3: Enums and chapter 6: functions, or re-ordering the lessons accordingly.

tajchumber:
@Snackomat some of the changes are gone, can you fix your commits please?

en/lessons/basics/basics.md
doomspork: @snackomat we'll get this merged in and then if you're up for it, I'd love for you to submit a PR for the pipe section 😁

doomspork:
@tajchumber we're good 👍 @Snackomat updated the PR to remove their comment and change the `is_atom/1` to no longer use pipes.

Per https://github.com/elixirschool/elixirschool/pull/2094#discussion_r344495507 we are going to want to add a section on pipes

PR #2114, opened 2019-10-31
[ES] Add Spanish translation of Ecto changesets lesson

erickgnavar:


doomspork:
@erickgnavar I'm going to merge these. After 12 days I suspect no one on @elixirschool/spanish has time so we should not hold off on merging.

PR #2122, opened 2019-11-02
[zh-hant] Translation of libraries/poolboy

lithiumpie:
- Translate libraries/poolboy


doomspork:
Thank you @lithiumpie :purple_heart:

PR #2128, opened 2019-11-14
update documentation : error syntax in deps code for earmark and ex_doc

magiknono:
syntax was wrong and not working


doomspork:
Please see #2129

PR #2129, opened 2019-11-14
update documentation : error syntax in deps code for earmark and ex_doc

magiknono:
syntax was wrong and not working

### DIFF
--- a/en/lessons/basics/documentation.md
+++ b/en/lessons/basics/documentation.md
@@ -198,8 +198,8 @@ In the `mix.exs` file, add the two required dependencies to get started: `:earma
 
 ```elixir
   def deps do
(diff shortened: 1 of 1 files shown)


doomspork:
@magiknono hello and welcome! 👋 Something must be off with how this was copied and pasted into your editor because the syntax in this lesson is indeed correct. If we do `$ mix new <project name>` and take a peek at `mix.exs` we'll see that the syntax in this lesson , a list of tuples with keyword lists for options, is what we'll see:

image

If we update our `mix.exs` to reflect this change we get a compilation error 
image
image

I'm going to close this PR but I'd be happy to help you troubleshoot the situations you ran into.

PR #2149, opened 2019-12-16
Add timezone operations paragraph to DateTime lesson

### LINKED ISSUE #2062: Need update DateTime lesson
https://github.com/elixir-lang/elixir/commit/b8b1a4ba8af8256567f00e6d9f7fa056881a8ec1

Since elixir 1.8.0, DateTime start to support time zone.

brain-geek:
Closes #2062

### DIFF
--- a/en/lessons/basics/date-time.md
+++ b/en/lessons/basics/date-time.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.2
+version: 1.1.0
 title: Date and Time
 ---
 
@@ -10,8 +10,6 @@ Working with time in Elixir.
 ## Time
 
 Elixir has some modules which work with time.
-Though it needs to be noted that this functionality is limited to working with UTC timezone.
-
 Let's start with getting the current time:
 
 ```elixir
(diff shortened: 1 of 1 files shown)


en/lessons/basics/date-time.md
```diff
@@ -92,8 +90,13 @@ It does not have the limitations noted in `NaiveDateTime`: it has both time and
 But be aware about timezones. The official docs state:
 
 ```
-You will notice this module only contains conversion functions as well as functions that work on UTC.
-This is because a proper DateTime implementation requires a time zone database which currently is not provided as part of Elixir.
+Many functions in this module require a time zone database. By default, it uses the default time zone database returned by Calendar.get_time_zone_database/0, which defaults to Calendar.UTCOnlyTimeZoneDatabase which only handles "Etc/UTC" datetimes and returns {:error, :utc_only_time_zone_database} for any other time zone.
```
SophieDeBenedetto: Instead of the three backtics to denote a quote, markdown recommends using the `>` carrot. Then, within that block quote, let's use the single backtic to wrap `Calendar.get_time_zone_database/0`, `Calendar.UTCOnlyTimeZoneDatabase` and `{:error, :utc_only_time_zone_database}` and any other code blocks.

brain-geek:
@SophieDeBenedetto Thanks for the review! I've applied your suggestion.

PR #2150, opened 2019-12-16
Fix my info in contributors page

brain-geek:


brain-geek:
@lex111 yup. Thanks for the review!

PR #2151, opened 2019-12-17
Fix text-indent bug in Google Chrome

ezrafree:
This fixes a bug in Google Chrome (and other Chromium-based browsers) which causes the `text-indent` and `text-overflow` (and possibly by extension `overflow`) properties to behave differently in Chrome version 77.0.3865.75 and above.

As a result, the fallback text of "Toggle" is displaying behind the hamburger icon that toggles the display of the sidebar menu.

[This StackOverflow post](https://stackoverflow.com/questions/58035496/chrome-update-affects-css-text-overflow-in-relation-to-text-indent) seems to describe the issue.

I've tested these changes in Chrome and Firefox on Ubuntu 19.10.

### DIFF
--- a/_assets/stylesheets/layout/_sidebar.scss
+++ b/_assets/stylesheets/layout/_sidebar.scss
@@ -126,7 +126,7 @@
 			overflow: hidden;
 			position: absolute;
 			text-align: center;
-			text-indent: -9999px;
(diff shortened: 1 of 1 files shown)


lex111:
@ezrafree thank you for this!

PR #2156, opened 2019-12-23
Update basics.md

roostercrab:
Age change is inconsistent. If that is because the value wasn't saved then it is confusing for new people.


doomspork:
Good catch @roostercrab! Do you want to see if this issue exists in translations and fix those as well? If not I can do so in a subsequent PR.

brain-geek:
@doomspork Let's merge this one and fix translations in a separate PR.

doomspork:
Done @brain-geek, good idea :+1: I hope you're well!

Thank you again @roostercrab!

PR #2168, opened 2020-01-18
Improve moduledoc example

### LINKED ISSUE #1696: More info in documentation
Ijust learn elixir and got in stuck in topic [Documentation](https://github.com/elixirschool/elixirschool/blob/master/en/lessons/basics/documentation.md#documenting-modules) I run code in iex like in example but cannot get same result
```
iex(27)> defmodule Greeter do
...(27)>   @moduledoc """
...(27)>   Provides a function `hello/1` to greet a human
...(27)>   """
...(27)> 
...(27)>   def hello(name) do
...(27)>     "Hello, " <> name
...(27)>   end
...(27)> end
warning: redefining module Greeter (current version defined in memory)
  iex:27

{:module, Greeter,
 <<70, 79, 82, 49, 0, 0, 4, 208, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 141,
   0, 0, 0, 15, 14, 69, 108, 105, 120, 105, 114, 46, 71, 114, 101, 101, 116,
   101, 114, 8, 95, 95, 105, 110, 102, 111, 95, ...>>, {:hello, 1}}
iex(28)> c("greeter.ex")
** (ArgumentError) could not find files greeter.ex
    (iex) lib/iex/helpers.ex:180: IEx.Helpers.c/2
iex(28)> h Greeter
Greeter was not compiled with docs
```

It seems that I need to save this module in the file and compile with doc allowed. I think this part needs more explanation for beginners in Elixir.

brain-geek:
Fixes #1696

### DIFF
--- a/en/lessons/basics/documentation.md
+++ b/en/lessons/basics/documentation.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.3
+version: 1.1.0
 title: Documentation
 ---
 
@@ -56,6 +56,7 @@ end
 ```
 
 We (or others) can access this module documentation using the `h` helper function within IEx.
+We can see this for ourselves if we put our `Greeter` module into a new file, `greeter.ex` and compile it:
 
 ```elixir
 iex> c("greeter.ex")
@@ -67,6 +68,8 @@ iex> h Greeter
 
 Provides a function hello/1 to greet a human
 ```
+
+_Note_: we don't need to manually compile our files as we did above if we're working within the context of a mix project. You can use `iex -S mix` to load the IEx console for the current project if you're working in a mix project.
 
 ### Documenting Functions
 


en/lessons/basics/documentation.md
```diff
@@ -56,6 +56,7 @@ end
 ```
 
 We (or others) can access this module documentation using the `h` helper function within IEx.
+To check that - put the contents of the previous snippet to `greeter.ex` file in the same directory you started IEx from, and run it there:
```
doomspork: How about:

```suggestion
We can see this for ourselves if we put our `Greeter` module into a new file, `greeter.ex` and compile it:
```

en/lessons/basics/documentation.md
```diff
@@ -68,6 +69,8 @@ iex> h Greeter
 Provides a function hello/1 to greet a human
 ```
 
+But if you're working in existing project, you don't need to include file manually, as we did with `c("greeter.ex")` in this example.
```
doomspork: I'm not entirely sure what we want to say here. Can you help me understand this a little better?

brain-geek: I intended something like this:

> Line `c("greeter.ex")` is due to our current environment, you won't need it in a proper mix project.

brain-geek: Any ideas how to rephrase it to sound good?

doomspork: Maybe we should include this as a "note" like we do in other lessons, something like:

```
_Note_: we don't need to manually compile our files as we did above if we're working within the context of a mix project.
```

How is this @elixirschool/english?

SophieDeBenedetto: Maybe we should add a note on running `iex -S mix` to load the IEx console for the current project if you're working in a mix project? We could explain that approach vs. the `load IEx, compile the file` approach described here for one-off Elixir files.

PR #2178, opened 2020-01-21
Fix spelling errors at the [EN] index and [UK] index

dokugo:
I've accidentally found some little spelling errors at the index pages of [EN] and [UK] site versions.


brain-geek:
@dokugo Hm... Are you sure both of those are problems?

```
Some of these translations include: ... 
```

```
Переведено на некоторые языки, включая: ... Русский
```

```
Деякі з цих перекладів включають: ... Український
```

I would agree on the russian change, it sounds better. But I'm not sure `Український` sounds better then original `Українською`. I guess, `Українську` would be better. @elixirschool/ukrainian , what do you think?

dokugo:
@brain-geek Well, I'm actually not so sure about the ukrainian change too. I guess I should have fixed the russian spelling only, so I removed the ukrainian part.

lex111:
@dokugo thanks!

PR #2197, opened 2020-02-25
Add translation report reference to README.md

Auralcat:
The [translation report (example for Portuguese)](https://https://elixirschool.com/pt/report/) lists the pages that need to have their content updated or that need to be translated from scratch.

Having the address for the report inside the README.md file along with the other translation instructions makes it easier for people new to the project to pick up a page and translate it, or update the contents of a translation in a given language. What do you think? :)

## Some questions
- [ ] Is there any more information we should add in the README file regarding the translation report?
- [ ] Are there any extra steps involving the translation report that people should do before their translation is able to be merged?

This is my first contribution to the project, thanks in advance! 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,9 @@ To run locally, you need both Ruby and Bundler installed.
 
 1. Each of the languages has a folder in root of this repo. To start translating you need to copy a file from the English language to the corresponding folder in your language and start translating.
 
-2. Translated lessons must include page metadata.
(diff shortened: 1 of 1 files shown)


gemantzu:
Nice touch. Thanks @Auralcat ! I think you can make a change though, make the url links to be relative to the website (If I remember correctly, it should be `/pt/report`. Ain't that right, @doomspork ?

doomspork:
@gemantzu the README doesn't get rendered on the website. There is an index file inside each language folder that is the homepage.

Issue #2207, opened 2020-03-01
[FR] Update Basics/Enum


g-belmonte:
Update french version of Basics/Enum to version 1.5.0

PR #2208, opened 2020-03-01
Update french version of basics/enum

g-belmonte:
Updating french text for basics/enum section.
Issue: https://github.com/elixirschool/elixirschool/issues/2207


doomspork:
Thank you very much @g-belmonte!

PR #2211, opened 2020-03-05
Update what mix generates

siaw23:
`mix new example` doesn't generate  a `config` folder and `config/config.exs` anymore.


tajchumber:
please reopen if needed

PR #2212, opened 2020-03-06
Update what mix generates

siaw23:
`mix new example` doesn't generate  a `config` folder and `config/config.exs` anymore.

### DIFF
--- a/en/lessons/basics/documentation.md
+++ b/en/lessons/basics/documentation.md
@@ -148,8 +148,6 @@ $ mix new greet_everyone
 * creating .formatter.exs
 * creating .gitignore
 * creating mix.exs
-* creating config
-* creating config/config.exs
 * creating lib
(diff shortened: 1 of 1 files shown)


doomspork:
Hello and thank you again @siaw23! Would you like to incorporate this change into the translations or shall I do that in a follow up?

siaw23:
By that do you mean just to change the version number? I don't know how translations work with this :)

doomspork:
@siaw23 you could change the version number or you could apply this change to the other language's file since this generated output and itself does not get translated. Changing these other instances of the output: https://github.com/elixirschool/elixirschool/blob/master/ru/lessons/basics/documentation.md#exdoc

Does that make sense?

siaw23:
Maybe you can close this too then? @tajchumber

PR #2214, opened 2020-03-06
Fix wording

siaw23:


doomspork:
@siaw23 could you bump the version at the top of the file from "1.0.3" to "1.1.0"? That'll update the translation report so folks know these changes should be incorporated.

siaw23:
Should I do this for my other 2 PRs as well?

doomspork:
@siaw23 yes please! Alternatively, you could apply those changes to all translations since it's Mix output and not something actually translated. I'll defer to you.

Thanks so much for your help! I have been meaning to go back through the lessons and do some sprucing up.

PR #2215, opened 2020-03-06
Improve clarity

siaw23:


en/lessons/basics/iex-helpers.md
```diff
@@ -1,5 +1,5 @@
 ---
-version: 1.0.1
+version: 1.1.0
```
doomspork: I apologize for the confusing versioning we use @siaw23 but let's call this `1.0.2` since it is a change primarily for English. there isn't likely to be much need for translations to update.

PR #2216, opened 2020-03-07
Updates to reflect current behaviour of `mix new`

johnthreesixteen:
Since `mix new` no longer generates the `config/` directory as of Elixir v1.9.0, the bash outputs for `mix new` occurring in `basics/mix.md` and `advanced/umbrella-projects.md` ought to be updated to reflect the new behaviour

### DIFF
--- a/en/lessons/advanced/umbrella-projects.md
+++ b/en/lessons/advanced/umbrella-projects.md
@@ -1,5 +1,5 @@
 ---
-version: 1.0.2
+version: 1.0.3
 title: Umbrella Projects
 ---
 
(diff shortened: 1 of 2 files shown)


doomspork:
Thank you @johnthreesixteen! If you'd like to apply these changes to the existing translations in another PR that would be welcomed, if not I can address that :tada: 

Thanks again!

johnthreesixteen:
> Thank you @johnthreesixteen! If you'd like to apply these changes to the existing translations in another PR that would be welcomed, if not I can address that 🎉
> 
> Thanks again!

Thanks @doomspork. Kindly handle the application of the changes to the existing translations.

PR #2220, opened 2020-03-09
`Scheduler` module expects singular `health_check` option

jordelver:
This PR fixes a typo in the [lesson](https://elixirschool.com/en/lessons/libraries/bypass/) for [Bypass](https://github.com/PSPDFKit-labs/bypass/).

The test [here](https://github.com/elixirschool/elixirschool/blame/master/en/lessons/libraries/bypass.md#L215) passes in a `health_checks` option, but the receiving code actually requires `health_check` as seen [here](https://github.com/elixirschool/elixirschool/blame/master/en/lessons/libraries/bypass.md#L168):

```elixir
def init(opts) do
  # ...
  health_check = Keyword.get(opts, :health_check, HealthCheck)
  # ...
end
```

### DIFF
--- a/en/lessons/libraries/bypass.md
+++ b/en/lessons/libraries/bypass.md
@@ -212,7 +212,7 @@ defmodule Clinic.SchedulerTest do
   end
 
   test "health checks are run and results logged" do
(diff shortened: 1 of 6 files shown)


doomspork:
Thank you and welcome to the party @jordelver!

PR #2221, opened 2020-03-11
Update "mix new" output across all files

doomspork:
This is a follow up on #2212 to apply that change to all instances of `mix new`


doomspork:
Thank you @julienXX! I hope all is well with you

PR #2223, opened 2020-03-13
Added clarity on the previous lesson

davidkhardwick:
One of the lessons from early chapters in Elixir school was called 'example' so this led to some confusion for me.  Calling it the example 'Friends' app is more clear.


davidkhardwick:

PR #2224, opened 2020-03-13
Updated path to movie.ex

davidkhardwick:
Referenced 'example' instead of 'friends'


davidkhardwick:

PR #2225, opened 2020-03-13
Moved all 'example' references to 'friends'

davidkhardwick:
Consistently using the Friends app from the previous lesson.


davidkhardwick:

PR #2226, opened 2020-03-13
Make it more clear to create a new app

davidkhardwick:
Make it more clear that you want to create a new app, and not leverage the prior lesson's app `Friends`

### DIFF
--- a/en/lessons/ecto/associations.md
+++ b/en/lessons/ecto/associations.md
@@ -9,7 +9,7 @@ In this section we'll learn how to use Ecto to define and work with associations
 
 ## Set Up
 
-We'll start off with the `Example` app from the previous lesson. You can refer to the setup [here](../basics) for a quick refresher.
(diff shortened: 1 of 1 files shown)


doomspork:
Hello and welcome @davidkhardwick! We're excited to have you involved :tada: 

Could you help me understand the rationale behind this change? We intend for the user to re-use the same app for all of the Ecto lessons, not create a new one and redo work each lesson. 

I'm open to rewording things but the number of changes will be far greater than just these here. If we are to move from `Example` to something else, this should be reflected in all impacted lessons across the translations. 

@elixirschool/developers any thoughts?

SophieDeBenedetto:
Welcome @davidkhardwick! First of all thank you for your contribution :) 

Yep I agree with @doomspork that the intention here is to have people re-use the app they set up in previous lessons, and we link to that lesson so that folks who have yet to complete it can get up and running.

davidkhardwick:
Okay, I first thought the intention was to re-use the previous 'Friends' lesson, but then with so many references to 'examples', I thought the intention was to create a new app for this lesson.

I'll create a single PR that changes the references to 'Friends' instead of example

PR #2230, opened 2020-03-15
Adding ie> prompt

davidkhardwick:
I was typing out %Ecto.Query<from m in Example.Movie> because I didn't get that I only needed to type out the first two statements here.


SophieDeBenedetto:
LGTM! 🚀

PR #2232, opened 2020-03-16
Updating to reference 'Friends' instead of 'Example'

davidkhardwick:
Similar to my previous edit, I'm updating all the references to 'Friends' instead of 'Example' so that the student can continue on with the same example app created in the first section of the Ecto lessons.


doomspork:
Thank you @davidkhardwick!

PR #2240, opened 2020-03-20
[pt] The correct output message from Datasets.Hello

romenigld:
The correct output message from Datasets.Hello is "Hello, I'm the datasets" and not the :hello.
Because the file was modified for show that message.

### DIFF
--- a/pt/lessons/advanced/umbrella-projects.md
+++ b/pt/lessons/advanced/umbrella-projects.md
@@ -199,5 +199,5 @@ Consolidated Inspect
 Interactive Elixir ({{ site.elixir.version }}) - press Ctrl+C to exit (type h() ENTER for help)
 
 iex> Datasets.hello
-:world
+Hello, I'm the datasets
 ```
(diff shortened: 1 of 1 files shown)


pt/lessons/advanced/umbrella-projects.md
```diff
@@ -199,5 +199,5 @@ Consolidated Inspect
 Interactive Elixir ({{ site.elixir.version }}) - press Ctrl+C to exit (type h() ENTER for help)
 
 iex> Datasets.hello
-:world
+Hello, I'm the datasets
```
vbrazo: According to this file https://github.com/elixirschool/elixirschool/blob/master/en/lessons/advanced/umbrella-projects.md, you're missing the `:ok` output.

```suggestion
Hello, I'm the datasets
:ok
```

romenigld: Nice, I just don't watch the english version.

PR #2246, opened 2020-03-28
Update escripts.md

kamilpikula:
add parentheses as with #2243


tajchumber:
Thanks for this @kamilpikula !

PR #2247, opened 2020-03-29
Update the iex output

romenigld:
I updated the iex output.
And was need to put the match for `person =  %Friends.Person{...}`
for after call the ´person.name` and others

### DIFF
--- a/pt/lessons/ecto/basics.md
+++ b/pt/lessons/ecto/basics.md
@@ -208,15 +208,26 @@ pessoa:
 
 ```elixir
 iex> %Friends.Person{}
-%Friends.Person{age: 0, name: nil}
+%Friends.Person{
+  __meta__: #Ecto.Schema.Metadata<:built, "people">,
+  age: 11,
+  id: nil,
+  name: "Tom"
+}
 ```
 
 Como esperado, recebemos uma nova `Person` com o valor padrão aplicado a `age`.
 Agora, vamos criar uma pessoa "real":
 
 ```elixir
-iex> %Friends.Person{name: "Tom", age: 11}
-%Friends.Person{age: 11, name: "Tom"}
+iex> person = %Friends.Person{name: "Tom", age: 11}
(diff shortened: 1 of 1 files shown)


pt/lessons/ecto/basics.md
```diff
@@ -208,15 +208,26 @@ pessoa:
 
 ```elixir
 iex> %Friends.Person{}
-%Friends.Person{age: 0, name: nil}
+%Friends.Person{
```
SophieDeBenedetto: I'm not sure this output is correct. If we open up `iex` and simply create a new empty `Friends.Person` struct, then the resulting struct would not have any values for `name` or `age`

romenigld: The output of the struct is this, but the results are differents.
```
%Friends.Person{
  __meta__: #Ecto.Schema.Metadata<:built, "people">,
  age: 0,
  id: nil,
  name: nil
}
```

SophieDeBenedetto: Right, so shouldn't the output you added below have `id: 0` and name `""`?

romenigld: This was the result of my terminal.
with the elixir  1.10.0 version.

SophieDeBenedetto: Ah okay, do you mean to say your output is this?

```elixir 
%Friends.Person{
  __meta__: #Ecto.Schema.Metadata<:built, "people">,
  age: 0,
  id: nil,
  name: nil
}
```

If so, that makes sense and we should use that output on the lines below. But currently you have the name set as `"Tom"` on line 215 which I think is incorrect, it should be `nil`?

romenigld: yes it should be name: nil. Sorry!

SophieDeBenedetto: no prob! Look good!

PR #2251, opened 2020-03-30
Make RSS feed more discoverable

jordelver:
I wanted to subscribe to the blog and couldn't find the feed.

This PR:

* Adds a `<link>` tag to the RSS feed to make it discoverable to browsers (and users)
* Adds a link with icon to the share icons section of the header to make the feed more discoverable to users

Image 2020-03-30 at 3 24 45 pm

I couldn't find whether the `feed.xml` was localised. I'm guessing the future it would be nice if the feed was localised and then we could link to the correct feed for each language.


doomspork:
Thank you @jordelver :grin:

jordelver:
Thanks!

PR #2264, opened 2020-04-05
[JA] Fix typo in Functions and Pattern Matching

matsu4ki:


ja/lessons/basics/functions.md
```diff
@@ -1,5 +1,5 @@
 ---
-version: 1.2.0
+version: 1.2.1
```
riseshia, line 2: This version is for managing diff between origin lesson(en) and translated one.
Hence,  you don't need to bump up version if just typo fix which caused by translation)

marocchino, line 2:
```suggestion
version: 1.2.0
```

matsu4ki: I got it👍️

PR #2266, opened 2020-04-08
[pt] fix minor details

herminiotorres:
- add parentheses
- define underscore prefix in some assigned variables are don't use
- update Task example

### DIFF
--- a/pt/lessons/advanced/concurrency.md
+++ b/pt/lessons/advanced/concurrency.md
@@ -1,5 +1,5 @@
 ---
-version: 1.1.0
+version: 1.1.1
 title: Concorrência
 ---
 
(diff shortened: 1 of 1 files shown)


doomspork:
@elixirschool/developers we should migrate to GitHub actions :thinking: Travis seems to be getting stuck a lot these days.

doomspork:
Thanks for the contribution too @herminiotorres! I hope you're staying safe in these crazy times.

herminiotorres:
@doomspork my concerns about the GitHub actions it because sometimes they work well, in the past I had some not good experience using Github actions.

I'm staying safe and thanks for asking and cares about me, and I hope you're safe as too! :pray:

PR #2274, opened 2020-04-10
[pt] bump GenStage version

herminiotorres:
- update the gen_stage version
- add a _initial variable on start_link function examples
- change the children list itens format


thiamsantos:
@doomspork Could you help here? The build passed https://travis-ci.org/github/elixirschool/elixirschool/builds/673439958 but we are still not able to merge the PR.

PR #2313, opened 2020-05-17
Update mix-tasks.md

JorisKok:
Add information about how you can start your application when running a mix task. I've found myself looking for it from time to time.

### DIFF
--- a/en/lessons/basics/mix-tasks.md
+++ b/en/lessons/basics/mix-tasks.md
@@ -107,8 +107,6 @@ defmodule Mix.Tasks.Hello do
     Hello.say()
   end
 end
-    
-
 ```
 
 ## Mix Tasks in Action
--- a/en/lessons/basics/mix-tasks.md
+++ b/en/lessons/basics/mix-tasks.md
@@ -93,7 +93,9 @@ Within this function, we call our `Hello` module and the `say` function.
 
 ## Loading your application
 
-If you want to load your application before executing the script, you can do so with:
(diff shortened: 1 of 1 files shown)


en/lessons/basics/mix-tasks.md
```diff
@@ -91,6 +91,24 @@ On the second line, we introduce the `use Mix.Task` which brings the `Mix.Task`
 We then declare a run function which ignores any arguments for now.
 Within this function, we call our `Hello` module and the `say` function.
 
+## Loading your application
+
+If you want to load your application before executing the script, you can do so with:
```
brain-geek, line 96: Could you expand on this explanation? I'm not sure what does this mean, even given I'm not a 'student'.

doomspork: We also want to move away from "you" and use "we" as we do elsewhere. What about something like this:

```suggestion
Mix does not automatically start our application or any of its dependencies which is fine for many Mix task use-cases but what if we need to use Ecto and interact with a database? In that case we need to make sure the app behind Ecto.Repo has started. There are 2 ways for us to handle this: explicitly starting an app or we can start our application which in turn will start the others.

Let's look at how we can update our Mix task to start our application and dependencies:
```

doomspork: Any thoughts on this suggestion @elixirschool/english / @elixirschool/developers? I'd like to get @JorisKok change merged in.

JorisKok: > Could you expand on this explanation? I'm not sure what does this mean, even given I'm not a 'student'.

Sorry I didn't reply, totally missed it. 

Thanks @doomspork , approved your suggestion. I'm not a native English speaker, so you might be better at wording it.

starfish719:
https://github.com/elixirschool/elixirschool/pull/2313/files#diff-ce81ef91d95296ee6fc601cb0c462f42R2

bump up???

doomspork:
@starfish719 good catch. Let's get that fix :+1:

PR #2317, opened 2020-05-25
Fix sidebar disappearing

### LINKED ISSUE #2302: Menu randomly not showing  up on iPad
FA03AB02-B165-409E-94BC-D51C822CCE4F

thiamsantos:
Closes #2302

The problem was that the removal of the `is-loading` was happening inside a window load event. But the event listener was sometimes been added after the page was already loaded, causing it to never be triggered.

The proposed solution is to just call the setTimeout that remove the class right after, inside the window ready event listener to avoid the race conditions.

### DIFF
--- a/_assets/javascripts/main.js
+++ b/_assets/javascripts/main.js
@@ -16,6 +16,12 @@
 		'small-to-xlarge': '(min-width: 481px) and (max-width: 1680px)'
 	});
 
+	$(window).on('load', function() {
+		setTimeout(function() {
+			$('body').removeClass('is-loading');
+		}, 100);
+	});
+
 	$(function() {
 
 		var	$window = $(window),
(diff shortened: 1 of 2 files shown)


doomspork:
I'm admittedly not the best person to decide on frontend solutions but this looks good to me. I'll give the other @elixirschool/developers members a chance to comment before merging this.

Thanks for taking the initiative and time to open this PR @thiamsantos, it is so very much appreciated :grin:

_assets/javascripts/main.js
```diff
@@ -40,12 +40,9 @@
 				if ($theme === 'dark') {
 					toggleThemeIcon($elToggleTheme);
 				}
-				$window.on('load', function() {
```
tajchumber, line 43: the point of the `$window.on('load'` is to wait for the page to load before removing the class and now we're just removing the class without waiting. I think this is not the right approach

What do you think @thiamsantos ?

thiamsantos: @tajchumber This approach will inevitably incur in a race condition, as that is a high chance that the `load` event already happened before the `$window.on('load'` listener even been added. The option I could think of was to add the `is-loading` class directly at the html, with that we could remove the `$window.on('load'` listener from inside the `ready` listener, and be sure that the event will always be triggered. WDYT @tajchumber?

The change would mainly be
- Alter https://github.com/elixirschool/elixirschool/blob/master/_includes/_sidebar.html#L2 to `<div id="sidebar" class="is-loading">`
- Add outside of https://github.com/elixirschool/elixirschool/blob/master/_assets/javascripts/main.js#L19 

```js
$window.on('load', function() {
  setTimeout(function() {
    $('body').removeClass('is-loading');
  }, 100);
});
```

tajchumber: @thiamsantos yes! I think this could work!

thiamsantos: @tajchumber I pushed the changes, could you take another look?

tajchumber:
nice work @thiamsantos! 🎉

PR #2323, opened 2020-06-03
[PT] Improving the translation for concurrency lesson

elainenaomi:
Hi all! 👋 

I found some small issues in the Portuguese translation for the Concurrency Lesson.
Part of them I had to adapt the phrase to be more clear in Portuguese - Brazil because terms like `to match messages` can be a little confusing if we don't refer to `pattern matching`. 
Let me know if I have to create an issue or adapt a term to Portuguese - Portugal.

Thanks for this amazing project ✨ 

The examples from Elixir School helped a lot to start to understand Elixir, including the advanced topics 🙂

### DIFF
--- a/pt/lessons/advanced/concurrency.md
+++ b/pt/lessons/advanced/concurrency.md
@@ -1,5 +1,5 @@
 ---
-version: 1.1.2
+version: 1.1.1
 title: Concorrência
 ---
 
--- a/pt/lessons/advanced/concurrency.md
+++ b/pt/lessons/advanced/concurrency.md
@@ -1,11 +1,11 @@
 ---
-version: 1.1.1
+version: 1.1.2
 title: Concorrência
 ---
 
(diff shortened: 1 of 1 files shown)


pt/lessons/advanced/concurrency.md
```diff
@@ -1,19 +1,19 @@
 ---
-version: 1.1.1
+version: 1.1.2
```
thiamsantos, line 2:
```suggestion
version: 1.1.1
```
@elainenaomi Even though this is a improvement to the translation we should keep the same version as the original https://github.com/elixirschool/elixirschool/blob/master/en/lessons/advanced/concurrency.md. We only bump when the content from the original changes.

thiamsantos:
@elainenaomi Thanks for the improvements, the text is much better :)

pt/lessons/advanced/concurrency.md
elainenaomi: Oh! Ok! Thanks!

thiamsantos:
Thanks @elainenaomi :heart: :tada:

PR #2169, opened 2020-01-18
Improve error handling lesson

brain-geek:
Based on text from @holsky 

Fixes #799

PR #2330, opened 2020-06-16
Fix link 404 introduced with #2169

geekyi:


doomspork:
Thank you and welcome @geekyi!

PR #2331, opened 2020-06-19
Make erlang variables difference section a bit more descriptive.

ShPakvel:


en/lessons/advanced/erlang.md
```diff
@@ -128,14 +128,19 @@ iex> "Hello World" |> to_charlist() |> :string.words
 
 ### Variables
 
+In Erlang variables starts from uppercase and re-binding is not allowed.
```
SophieDeBenedetto, line 131: Just a minor grammatical suggestion:

> In Erlang, variables begin with an uppercase letter and re-binding is not allowed.

ShPakvel: Agree, much better. Thank you! :bow:

PR #2340, opened 2020-06-25
Tiny fix

esambo:
Thanks for the nice article!


esambo:
No, I think the article was fine and didn't seem dated

doomspork:
That's great to hear! Have a wonderful week @esambo

PR #2365, opened 2020-09-07
Add sentence about DB Roles to Ecto Basics

maxwolff:
Got a little stuck in the tutorial, turns out I didn't have a role called `postgres`. Thought it'd be nice to show how to fix that.

### DIFF
--- a/en/lessons/ecto/basics.md
+++ b/en/lessons/ecto/basics.md
@@ -71,8 +71,7 @@ config :friends, Friends.Repo,
   hostname: "localhost"
 ```
 
-This configures how Ecto will connect to the database. Either edit the username to match one of your existing database roles, or create a new one in your `psql` shell as follows: 
-`CREATE ROLE user WITH SUPERUSER CREATEDB LOGIN PASSWORD 'pass';`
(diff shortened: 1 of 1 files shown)


en/lessons/ecto/basics.md
```diff
@@ -71,7 +71,8 @@ config :friends, Friends.Repo,
   hostname: "localhost"
 ```
 
-This configures how Ecto will connect to the database.
+This configures how Ecto will connect to the database. Either edit the username to match one of your existing database roles, or create a new one in your `psql` shell as follows:
```
doomspork, line 74: Let's just note that you may need to configure the connection details to match your particular database's needs.

I'm afraid if we start going down the rabbit hole of SQL commands it won't just stop with this one, we'll want to add MySQL, MSSQL, etc.

doomspork (changes requested):

en/lessons/ecto/basics.md
maxwolff: ok, sure

doomspork (approved):

PR #2378, opened 2020-09-22
Adding Clipboard.js to help copying code samples

konung:
Adding styling and code from https://github.com/marcoaugustoandrade/jekyll-clipboardjs with small modifications

See screenshot for samples
Screen Shot 2020-09-21 at 23 35 07
Screen Shot 2020-09-21 at 23 35 32


doomspork:
@konung this is _awesome_ thank you so so much! I'll pull this down and make sure all is good before merging :tada:

konung:
No problem. 

The one thing to keep in mind, as I've shown in screenshots - this works the same for either time of code sample: 
1. Actual code sample
2. Copypaste from console/iex.

The only way to separate the two - is manually
1. Change the class name on "console" examples
2. Remove "iex>" from "console" examples.  <- this might be easier, as simple search/replace in Text Editor? And adding some kind of indicator that it's the output from iex?

doomspork (approved):

doomspork:
Hmm. #2 is probably the best option but I would definitely like to get more folks opinions on that. @konung would you be open to creating an issue to talk through a solution to this?

konung:
@doomspork By all means.

PR #2407, opened 2020-10-01
[pt] update Distillery lesson

herminiotorres:


pt/lessons/libraries/distillery.md
```diff
@@ -109,7 +111,7 @@ config :book_app, BookAppWeb.Endpoint,
 para isto:
 
 ```
```
thiamsantos, line 113:
```suggestion
```elixir
```

@herminiotorres Could you also add the "lang" tag to the code blocks, as it was also added to the original content?

thiamsantos (approved):

PR #2422, opened 2020-10-05
change atoms from :"_" to :_

felipelincoln:
It yields a warning.

```elixir
iex> :ets.match :testing, {:"$2", :"_"} 
#warning: found quoted atom "_" but the quotes are not required. Atoms made exclusively of Unicode letters, numbers, underscore, and @ do not require quotes
```


doomspork (approved):
Hello and welcome @felipelincoln! Thanks so much for the change :tada:

PR #2428, opened 2020-10-09
LiveView SVG Charts with the Contex Library

SophieDeBenedetto:
As part of my work on Pragmatic Bookshelf's upcoming LiveView book, together with @batate, I came across @mindok's Contex  library for server-side SVG chart rendering https://github.com/mindok/contex. The library plays really nicely with LiveView, even exposing some ability to attach Phoenix DOM elements bindings to chart SVG elements so that you can handle chart events in your LiveView. 

This blog post aims to:
* Generate some early excitement for and awareness of the upcoming LiveView book by illustrating some of the concepts and topics we will cover in the book. To this end, PragProg is offering a few free copies to Elixir School to give away to readers. This is mentioned in the post.
* Share some learnings about working with the Contex library and draw some attention to/support for this package


mindok:
Looks really good @SophieDeBenedetto. I have added an issue to allow optional specification of `phx-target` to the event handler. Also, there is a slightly more streamlined API for combining `BarChart` setup and `Plot` setup that was added a little later so it doesn't appear in all the examples. https://github.com/mindok/contex/blob/master/test/contex_plot_test.exs around line 39 shows an example. It is mentioned in the docs in passing around here https://hexdocs.pm/contex/Contex.Plot.html#new/5

Really appreciate the highlight article - what you've presented is pretty well exactly what I had in mind when I started out. There are a couple of other options out there (e.g. https://github.com/srowley/ggity - @srowley has made some great contributions to ContEx too, and https://github.com/elcritch/plotex). The goals and non-goals of the libraries are different - ContEx is probably most closely aligned with this kind of dashboard use case, but it may be useful to call out the other options to your readers. For example, ggity would be excellent for exploratory data analysis.

mindok:
PS - unless you want `send_update` to be a learning moment, I could probably turnaround the `phx-target` issue over the next 24 hrs...

srowley:
This is great! 

I did have one (unsolicited) comment, that frankly doesn't need to be addressed, but it touches on a cool part of the library so I will just mention it. This sentence:

> `Dataset` handles several different data structures by marshalling them into a consistent form for consumption by the chart plotting functions. 

is not technically accurate. `Dataset` doesn't transform different structures into a consistent form; it defines different accessor functions for retreiving the data from a given structure based on the type of structure - i.e., as it loops through a list of maps, it uses something like `Map.get(row, variable_name)` to get each value for a variable, whereas if the data is a list of tuples, it will use `elem(row, variable_index)`.

It doesn't matter at all for purposes of the article. I just think it is a great design feature that I would not have thought of; with this approach one doesn't loop through the data each time it is updated in order to transform it into a common format again, which is important for an application like a dashboard where it is likely updates will be frequent.

SophieDeBenedetto:
> This is great!
> 
> I did have one (unsolicited) comment, that frankly doesn't need to be addressed, but it touches on a cool part of the library so I will just mention it. This sentence:
> 
> > `Dataset` handles several different data structures by marshalling them into a consistent form for consumption by the chart plotting functions.
> 
> is not technically accurate. `Dataset` doesn't transform different structures into a consistent form; it defines different accessor functions for retreiving the data from a given structure based on the type of structure - i.e., as it traverses a list of maps, it uses something like `Map.get(row, variable_name)` to get each value for a variable, whereas if the data is a list of tuples, it will use `elem(row, variable_index)`.
> 
> It doesn't matter at all for purposes of the article. I just think it is a great design feature that I would not have thought of; with this approach one doesn't loop through the data each time it is updated in order to transform it into a common format again, which is important for an application like a dashboard where it is likely updates will be frequent.

Thanks for the clarification @srowley. That absolutely makes sense to me and is in-line with my understanding of the library. I did more or less pull that statement from the docs here https://hexdocs.pm/contex/Contex.Dataset.html, so it would probably be helpful to PR those docs you're interested 😄 

I'll make an update to the wording in this blog post 🚀

SophieDeBenedetto:
@mindok thanks for your feedback and giving this such a close read! I agree it would be helpful to mention some of the other libraries you listed and distinguish some of their goals, I'll make that update before publishing 😄 

Also great that you opened an issue around making `phx-target` configurable! I think what I'll do is link to that issue in the post and give our readers a chance to contribute if anyone is interested. (Obviously feel free to knock out that work at any time though). Linking to the issue will work as a sort of living documentation--at such time as it is addressed and closed, readers who click through to it will be able to learn about the change and understand how to use it. I don't mind using `send_update/2` as a teachable moment or just teasing that readers can learn more about it in the upcoming LiveView book in which we will go deeper on that topic. 

Lastly, thanks for sharing this https://github.com/mindok/contex/blob/master/test/contex_plot_test.exs#L41 re: how to initialize a plot for a barchart. One question--if you use this approach, where/how can you specify bar chart options like `BarChart.colours`?

SophieDeBenedetto:
Addressed your comments @mindok and @srowley! Thanks for you support and feedback--I've added a "Thank You" section to the bottom of the post giving ya'll a shoutout 😄

doomspork (approved):

mindok:
> Lastly, thanks for sharing this https://github.com/mindok/contex/blob/master/test/contex_plot_test.exs#L41 re: how to initialize a plot for a barchart. One question--if you use this approach, where/how can you specify bar chart options like `BarChart.colours`?

Hmmm - good question. Looks like we need to coordinate options better to allow this.

PR #2440, opened 2020-10-14
Add basics of Czech translation

martin0rszulik:
Needs to be complete, but I'm glad I finally did it 😄


gemantzu:
Thank you for your hard work @martin0rszulik !

Can you please remove the lessons you haven't translated? Leaving them like this will make them show as translated when selecting an english lesson and will even mess with the translation report system and show all of them as up to date.

Leave in the PR only the lessons you have actually translated.

Thanks again!

doomspork:
@martin0rszulik @gemantzu what do you want to do with this PR? Now that we've moved to the new format we need to update this PR and point to `master`.

gemantzu:
The problem with this PR is that it has all the English files copied over. We should remove those and keep the rest of the work.

theokeist:
Sorry for long wait (🥔) anyway I would like to somehow continue.. what is the current state as I can see new legacy repo elixirschool:legacy

doomspork:
@theokeist this is still the repo for lessons 👍 The school_house repo is just the Phoenix app that serves these lessons

PR #2452, opened 2020-10-22
Update 2020-10-06-server-side-svg-charts-with-contex-and-liveview.md

alaadahmed:
I think you need to pass the variable `games_with_average_ratings` to `Contex.DataSet.new/1` function


SophieDeBenedetto (approved):
Right you are, thanks!

PR #2471, opened 2020-10-27
first attempt to switch to Github Actions

gemantzu:


gemantzu:
I am pretty sure we don't need all those steps on the deploy part, and I would question if we need the third party helper, but it works like that (and didn't when I tried to do a manual git push on heroku).

Thoughts?

SophieDeBenedetto (dismissed):
🎉 this looks great, so excited to be using GH Actions! I think the third-party action is the way to go

gemantzu:
Thanks for the review @SophieDeBenedetto ! I guess we should wait a couple of days for more people to chime in their thoughts, first time I try something with Github Actions, I feel uneasy yet with them (although it worked very well on my fork)

.travis.yml
```diff
@@ -1,11 +0,0 @@
-language: ruby
-rvm:
-  - 2.5
-cache: bundler
-before_install: gem install bundler:1.16.3
-install:
-- bundle install
-before_script:
- - chmod +x ./script/build.sh
```
marocchino, line 9: We can delete this file.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,32 @@
...
+    - run: gem install bundler
+    - run: bundle config path vendor/bundle
+    - run: bundle install --jobs 4 --retry 3
+    - run: mkdir public
+    - run: bundle exec jekyll build --destination ./public
+    - uses: akhileshns/heroku-deploy@v3.6.8 # This is the action
+      with:
+        heroku_api_key: ${{secrets.HEROKU_API_TOKEN}}
+        heroku_app_name: ${{secrets.HEROKU_APP_NAME}} #Must be unique in Heroku
+        heroku_email: "flamefist@gmail.com"
```
doomspork, line 32: Should we see if we can move this to GitHub Pages at the same time instead of continuing to use Heroku to serve a static site?

gemantzu: Good idea @doomspork . Hopefully, I will try it later tonight.

gemantzu:
I have made an attempt to switch to GithubPages, but at the very least we need [this setup as well](https://jekyllrb.com/docs/continuous-integration/github-actions/#providing-permissions).

doomspork:
@gemantzu with us getting to the point of launching school house, do you think we can close this PR?

gemantzu:
Yes sir!

PR #2486, opened 2020-11-03
update puma to 5.0.4

jaimeiniesta:
I was having trouble installing puma 4 so I checked for an update, here's puma 5 working.


vbrazo (approved):
👍

doomspork (approved):

PR #2506, opened 2020-11-26
Add missing highlighting to code blocks

antedeguemon:
There were some code blocks that missed language specification.

Since there is no default language set for Jekyll, those code blocks ended up not being highlighted.


SophieDeBenedetto (approved):
Thanks for this fix!

PR #2508, opened 2020-12-10
Add some changes for less error prone

romenigld:
I add these changes because  I am doing the tutorial and when I try to run for the first time the `iex -S mix`.
I receive an error and it complains the Earmark was not loaded.
So for a while investigating the problem I realized that the error was because the file `/priv/posts/2020/11-04-exciting-news.md` was empty with no Markdown code.
And for the tutorial be less error prone I do these changes.
I think for a beginner level in elixir it will helps.

### DIFF
--- a/en/lessons/libraries/nimble-publisher.md
+++ b/en/lessons/libraries/nimble-publisher.md
@@ -29,14 +29,20 @@ Now, let's add some posts. We need to start creating a directory that will conta
 /priv/posts/YEAR/MONTH-DAY-ID.md
 ```
 
+So create these folders:
+
+```shell
+mkdir -p priv/posts/2020
+```
+
 For example, we start with these two posts:
 
-```
-/priv/posts/2020/10-28-hello-world.md
-/priv/posts/2020/11-04-exciting-news.md
+```shell
+touch /priv/posts/2020/10-28-hello-world.md
+touch /priv/posts/2020/11-04-exciting-news.md
 ```
 
-A typical blog post will be written in Markdown syntax, with a metadata section on top, and the content below separated by `---`, like this:
+A typical blog post will be written in Markdown syntax, with a metadata section on top, and the content below separated by `---`, like the `/priv/posts/2020/10-28-hello-world.md` file:
 
 ```
 %{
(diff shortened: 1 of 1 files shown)


en/lessons/libraries/nimble-publisher.md
```diff
@@ -29,14 +29,20 @@ Now, let's add some posts. We need to start creating a directory that will conta
 /priv/posts/YEAR/MONTH-DAY-ID.md
 ```
 
-For example, we start with these two posts:
+So create these folders:
```
SophieDeBenedetto, line 32: Should say `create this folder` - singular since there is only one folder

en/lessons/libraries/nimble-publisher.md
```diff
@@ -49,6 +55,20 @@ A typical blog post will be written in Markdown syntax, with a metadata section
 Yes, this is **the post** you've been waiting for.
 ```
 
+and add these content for the `/priv/posts/2020/11-04-exciting-news.md` file:
```
SophieDeBenedetto, line 58: should say `this content`

SophieDeBenedetto (changes requested):
Thanks for your contributions! Left two small comments :)

en/lessons/libraries/nimble-publisher.md
jaimeiniesta: @romenigld thanks but I think that these instructions are too low-level. The reader doesn't need to be told to create each folder, she should already know that in order to create the files, she needs to create the folders that contain them.

jaimeiniesta:
Thanks for your suggestions @romenigld but I don't see this helps in making the tutorial more readable - the reader is supposed to know how to create directories and folders, in my humble opinion.

Thanks for your corrections @SophieDeBenedetto 

I would rather not merge this PR.

en/lessons/libraries/nimble-publisher.md
```diff
@@ -29,14 +29,20 @@ Now, let's add some posts. We need to start creating a directory that will conta
 /priv/posts/YEAR/MONTH-DAY-ID.md
 ```
 
-For example, we start with these two posts:
+So create these folders:
 
+```shell
+mkdir -p priv/posts/2020
```
jaimeiniesta, line 35: This is too low level.

en/lessons/libraries/nimble-publisher.md
```diff
@@ -29,14 +29,20 @@ Now, let's add some posts. We need to start creating a directory that will conta
...
 ```
-/priv/posts/2020/10-28-hello-world.md
-/priv/posts/2020/11-04-exciting-news.md
+
+For example, we start with these two posts:
+
+```shell
+touch /priv/posts/2020/10-28-hello-world.md
+touch /priv/posts/2020/11-04-exciting-news.md
```
jaimeiniesta, line 42: This is too low level as well.

en/lessons/libraries/nimble-publisher.md
jaimeiniesta, line 58: Too low level as well. We don't need to instruct the reader how to do these operations.

PR #2509, opened 2020-12-10
For be less error prone I add these changes.

romenigld:
I add the path to create the view file `blog_view.ex`.

### DIFF
--- a/en/lessons/libraries/nimble-publisher.md
+++ b/en/lessons/libraries/nimble-publisher.md
@@ -221,7 +221,13 @@ end
 
 ### View
 
-Create the view module where you can place the helpers needed to render the view. By now it's just:
+Create the view module where you can place the helpers needed to render the view. First create the file: 
+
+```shell
+touch lib/nimble_school_web/views/blog_view.ex
+```
+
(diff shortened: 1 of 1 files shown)


en/lessons/libraries/nimble-publisher.md
```diff
@@ -221,7 +221,13 @@ end
 
 ### View
 
-Create the view module where you can place the helpers needed to render the view. By now it's just:
+Create the view module where you can place the helpers needed to render the view. First create the file: 
+
```
SophieDeBenedetto, line 225: We could consolidate this wording a little to something like:

`Create a file, lib/nimble_school_web/views/blog_view.ex, and add the following content:`

SophieDeBenedetto (changes requested):

en/lessons/libraries/nimble-publisher.md
jaimeiniesta: Yes, I prefer this wording as the important thing is to create the file, rather than how to create it - it's not important if you `touch` it via shell or just -like me- dare to use a mouse to do it :D

PR #2527, opened 2021-01-19
Update enum.md

zwippie:
I don't know when `:asc/:desc` was introduced, but I think this would be a nice addition without going into the gory details.


doomspork (approved):
I wholeheartedly agree @zwippie this is a good change :+1:

PR #2528, opened 2021-01-19
Update mix.md

zwippie:
Fix incorrect module name.


doomspork (approved):
Good catch @zwippie! In the future you can apply this kind of change across all the translations. I'll go ahead and do that now for this one :+1:

PR #2531, opened 2021-01-21
Add text of　returns the value of the equation

taku333:
Currently, the example using the "Pattern Matching" List does not have the standard output as shown below.
So we fixed it.

https://elixirschool.com/en/lessons/basics/pattern-matching/

```
# Lists
iex> list = [1, 2, 3]
iex> [1, 2, 3] = list
[1, 2, 3]
...
```


doomspork (approved):
Thank you and welcome @taku333 :tada:

PR #2535, opened 2021-01-23
Update mox lesson

thiamsantos:
- Simplify instructions, removing the step to start mox manually
- Add `async: true` to the final example, to make clear that the tests can be run async

### DIFF
--- a/en/lessons/libraries/mox.md
+++ b/en/lessons/libraries/mox.md
@@ -146,7 +146,7 @@ Then you may include `mox` in your dependencies:
 defp deps do
   [
     # ...
-    {:mox, "~> 0.5.2", only: [:test], runtime: false}
+    {:mox, "~> 0.5.2", only: :test}
   ]
 end
 ```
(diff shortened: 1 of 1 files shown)


en/lessons/libraries/mox.md
```diff
@@ -155,22 +155,18 @@ Install it with `mix deps.get`.
 
 Next, modify your `test_helper.exs` so it does 3 things:
```
SophieDeBenedetto, line 156: Now this should read `Next, modify your `test_helper.exs` so it does *2* things:` 😄

SophieDeBenedetto (changes requested):
Thanks for these updates! Just one small request 🚀

en/lessons/libraries/mox.md
thiamsantos: Done. Thanks for catching that @SophieDeBenedetto :)

SophieDeBenedetto (approved):

PR #2538, opened 2021-01-29
Update poolboy.md

dlobo:
Catching an exception thrown by GenServer.call in the transaction function. If we don't do this, a timeout de-commisions that specific worker and pretty soon the system stops responding since no workers are around

### DIFF
--- a/en/lessons/libraries/poolboy.md
+++ b/en/lessons/libraries/poolboy.md
@@ -151,7 +151,14 @@ defmodule PoolboyApp.Test do
     Task.async(fn ->
       :poolboy.transaction(
         :worker,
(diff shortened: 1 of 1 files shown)


doomspork:
Howdy @dlobo! I apologize for the delay getting back to you I missed the notification for this PR by mistake. I think this is a good idea but do you think we should add some comments around this block to clarify why this is here and what it is going?

dlobo:
> Howdy @dlobo! I apologize for the delay getting back to you I missed the notification for this PR by mistake. I think this is a good idea but do you think we should add some comments around this block to clarify why this is here and what it is going?

Added a few comments to the changes made

doomspork (approved):

PR #2557, opened 2021-04-08
Remove reference to postgrex using poolboy

maartenvanvliet:
No longer the case, see https://github.com/elixir-ecto/db_connection/commit/98b23398afc6f8af8e21d598c9b31741ac9b8e6d


SophieDeBenedetto (approved):
Thanks for the update!

PR #2564, opened 2021-04-21
Migration to NimblePublisher format and new structure — Content Only

doomspork:
This is a work-in-progress PR that includes a very large refactoring and migration of the existing content into a new structure and format suitable for NimblePublisher.

Used by https://github.com/elixirschool/school_house and https://beta.elixirschool.com


doomspork:
Dang. This won't work. The way this PR came together has botched the history of all the lessons and I'm not willing to destroy that much history. I'll have to figure out a better approach to this. :cold_sweat:

PR #2571, opened 2021-05-14
Fix an error with chunk_every arity

DrJackilD:
There was an error in Russian translation regarding `chunk_every/2` arity (there was `chunk_every/4` which is incorrect)


lex111 (approved):

lex111:
@DrJackilD thanks!
=

Issue #2496, opened 2020-11-13
Blog Live View With Presence: Usage of broadcast_from doesn't match signature in docs


Exadra37:
In the [Phoenix.PubSub](https://hexdocs.pm/phoenix_pubsub/Phoenix.PubSub.html#broadcast/4) docs report this usage for `broadcast_from/4`:

```elixir
broadcast_from(pubsub, from, topic, message, dispatcher \\ __MODULE__)
broadcast_from(t(), pid(), topic(), message(), dispatcher()) ::
  :ok | {:error, term()}
```

In this [blog post](https://elixirschool.com/blog/live-view-with-presence/) we have this code:

```elixir
PhatWeb.Endpoint.broadcast_from(self(), topic(chat.id), "message", %{chat: chat})
``` 

So, based on the docs I was expecting to see the usage of `self()` as the second parameter, not as the first.

doomspork:
The blog post has been updated. We should probably consider adding a note to blog posts about the version of Elixir, OTP, and the specific library that was used. It'll be difficult to keep this content updated over time as projects evolve.

PR #2575, opened 2021-05-18
Blog Live View With Presence: Usage of broadcast_from doesn't match signature in docs

samantadearaujo:
Issue: https://github.com/elixirschool/elixirschool/issues/2496
I change the order of the parameter to second. 
#2496


doomspork (approved):
Thank you and welcome @samantadearaujo!

Issue #2594, opened 2021-06-06
[FR] Update lesson Basics/Enum


g-belmonte:
Update French version of Basics/Enum from 1.5.0 to 1.7.0

PR #2595, opened 2021-06-06
[FR] Update lesson Basics/Enum

g-belmonte:
Closes https://github.com/elixirschool/elixirschool/issues/2594


doomspork (approved):
Thank you so much for taking the time to update these lessons! It's really appreciated!

What is your twitter handle by the way?

g-belmonte:
My pleasure! @doomspork 

My twitter handle is @BelmonteDev
I joined twitter last year, and I haven't done much there... It seems like a ghost account, but believe, I'm hidden somewhere inside that place 😅

PR #2603, opened 2021-06-13
Trigger a deploy of the School House repo

doomspork:
See https://github.community/t/triggering-by-other-repository/16163 and https://github.com/elixirschool/school_house/pull/61


doomspork:
Tested the cURL command and it works. I believe this file is good to go so I'm going to merge 👍

PR #2615, opened 2021-07-25
removes invalid `frameborder` attribute on Youtube iframes.

jaimeiniesta:
The `frameborder` attribute is not a valid `iframe` attribute and it's not making a visual difference in the site (we could use CSS if needed), let's remove it.

Found in https://rocketvalidator.com/s/328d31c1-0a28-423f-88dc-64c5c0d97590/i/112975746

### DIFF
--- a/_posts/2019-03-18-phoenix-live-view.md
+++ b/_posts/2019-03-18-phoenix-live-view.md
@@ -30,7 +30,7 @@ Let's get LiveView up and running to support a feature that pushes out live upda
 Here's the functionality we're building:
 
 <div class="responsive-embed">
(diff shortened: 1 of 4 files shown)


doomspork (approved):
@jaimeiniesta we will probably want these changes on the branch being used to power SchoolHouse. I'll do a rebase later today and that _should_ address that.

jaimeiniesta:
@doomspork ah, right, is that the `content-only-changes` branch?

doomspork:
Yup! No worries though, I need to keep that synced with this master until we launch. I'll rebase here shortly to also incorporate the various lesson updates that have occurred.

PR #2618, opened 2021-07-27
Update changeset validation notes

adrianbarbic:
Adding the atom `:age` to the code example making it easier to follow that it's talking about the same changeset made above, not creating a second changeset for validation.


doomspork (approved):
This is great @adrianbarbic, thank you!

PR #2630, opened 2021-08-10
add key - no unique example

cassiogiehl:


thiamsantos:
@cassiogiehl Please suggest additions to the text into the english text at `en/lessons/basics/collections.md`. After that we can update all the translations accordingly.

We don't accept changes to the translations that don't exist in the original english text.

PR #2651, opened 2021-09-02
Fix link to elixirbookclub.com

jaimeiniesta:
It looks like the site is not configured for the `www.` URL.


kinson (approved):
Thanks for the fix!

doomspork (approved):

PR #2656, opened 2021-09-25
Update current actions

doomspork:
1. Update the content deployment workflow to use the `master` branch
2. Remove the current `ci` workflow
3. Remove extra files


kinson (approved):
🎉

doomspork:
Sorry @kinson I pushed a comment just after you reviewed, dismissing it.

.github/CODEOWNERS
```diff
@@ -2,38 +2,39 @@
 # Order is important. The last matching pattern takes the most precedence.
 
 # Default owners for everything in the repo.
-*        @elixirschool/developers
+*        @elixirschool/english
 
-en/      @elixirschool/english
+/lessons/en/      @elixirschool/english
```
kinson, line 7: since it was `en/` before, should the new path be `lessons/en/` or do both work?

kinson (approved):

.github/CODEOWNERS
doomspork: I _think_ we just need the latter since the former no longer exists. Thoughts?

kinson: Sorry for the confusion, I was just referring to the leading slash 😅 

So:

`/lessons/en` vs
`lessons/en`

doomspork: Oh! Yeah, that makes sense

kinson: I'm not sure if it makes a difference tbh so if you want to merge this and then check that works!

doomspork:
We should figure out a CI pipeline for this repo later but that doesn't need to block us finishing other things.

.github/CODEOWNERS
```diff
@@ -2,38 +2,38 @@
...
+lessons/en/      @elixirschool/english
 
-es/      @elixirschool/spanish
+lessons/es/      @elixirschool/spanish
 
-fr/      @elixirschool/french
+lessons/fr/      @elixirschool/french
 
-gr/      @elixirschool/greek
+lessons/gr/      @elixirschool/greek
```
kinson, line 13:
```suggestion
lessons/el/      @elixirschool/greek
```

Since we just updated gr -> el

kinson (changes requested):
Whoops missed this on the first pass

doomspork:
@kinson if it’s GTG please merge when you approve 🙏

kinson (approved):

PR #2663, opened 2021-09-30
Fix Excerpt Indentation In Lesson Files

kinson:
# Overview

This pr fixes the indentation used in the lesson excerpt text.

When building the school_house application right now we get a lot of warnings about the issue, and updating these will eliminate those warning messages.

**No text/content was changed in this pr, just spacing adjustments**

An example of the warning I am referring to:
Screen Shot 2021-09-30 at 8 20 29 AM


doomspork (approved):
Thank you! 🎉

PR #2668, opened 2021-10-03
Fix duplicated word

OiTAvO:
Removing duplicated word from this sentence: "Telemetry is an open source suite of libraries which aims to to unify and standardize how the libraries and applications on the BEAM are instrumented and monitored."


doomspork (approved):

doomspork:
Thank you @OiTAvO!

PR #2704, opened 2021-10-24
Update functions.md

george124816:
out of date error


doomspork:
Thank you @george124816! We need to bump the lesson version at the top of the file so the translations know to pull this change in 🙏

doomspork (approved):

PR #2718, opened 2021-11-09
Update CI to build School House

### LINKED ISSUE #2695: Implement a CI tool to ensure School House will build
This recent commit https://github.com/elixirschool/elixirschool/commit/2fb69b7a9edc193d39eff715d59c3a83b9496656 includes files that cannot be compiled and break the build. We need to implement a CI step to ensure this does not happen in the future.

image

doomspork:
This action ensures the content is properly formatted by attempting to build the School House app. If it can't build, we need to identify the formatting issue and resolve it.

Closes #2695

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,32 @@
+---
+name: Continuous Integration
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+  push:
+    branches:
+      - master
+
+jobs:
+  Build:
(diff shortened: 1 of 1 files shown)


kinson:
@doomspork thanks for doing this!! This is way better than the approach I had visualized, no need to update the school_house repo at all!

Do you mind updating a lesson to fail compilation so we can verify the check will fail on GitHub? I believe an easy way to do this is to mess up the metadata at the top of the site by making it an invalid map.

kinson (approved):
💯

PR #2733, opened 2021-11-23
Update 2021-11-22-til-cleaner-queries-with-ecto-map.md

angrycandy:
- correct first sentence wording
- not sure why the last sentence got changed by the github editor; that was not my intention

### DIFF
--- a/posts/2021-11-22-til-cleaner-queries-with-ecto-map.md
+++ b/posts/2021-11-22-til-cleaner-queries-with-ecto-map.md
@@ -13,7 +13,7 @@ Today I learned how to write cleaner `Ecto` select queries with the help of `Ect
 
 # TIL: Cleaner queries with Ecto `map`
 
-When working with large tables, is a common practice to avoid using `SELECT *` to make a beter use of computer resources and database indices.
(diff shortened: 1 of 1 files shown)


posts/2021-11-22-til-cleaner-queries-with-ecto-map.md
```diff
@@ -13,7 +13,7 @@ Today I learned how to write cleaner `Ecto` select queries with the help of `Ect
 
 # TIL: Cleaner queries with Ecto `map`
 
-When working with large tables, is a common practice to avoid using `SELECT *` to make a beter use of computer resources and database indices.
+When working with large tables, a common practice avoids using `SELECT *` to make better use of computer resources and database indices.
```
doomspork, line 16: Thanks @angrycandy! I think we may still be missing something from this sentence 🤔 How does this read:

```suggestion
When working with large tables a common practice is to avoid using `SELECT *` to make better use of database indices and resources.
```

I also rearranged the end of the sentence to emphasize the benefit is to indices. 

cc @SophieDeBenedetto

doomspork (approved):

PR #2736, opened 2021-11-24
[PL] Add translation for lesson "Date and Time"

jangromko:
This PR contains:
1. Polish translation of lesson "Date and Time".
2. Minor improvements for English version of that lesson.

### DIFF
--- a/lessons/en/basics/date_time.md
+++ b/lessons/en/basics/date_time.md
@@ -1,5 +1,5 @@
 %{
-  version: "1.1.1",
+  version: "1.1.2",
   title: "Date and Time",
   excerpt: """
   Working with time in Elixir.
@@ -42,7 +42,7 @@ But there is a catch: as you may have noticed, this struct only contains the tim
 
 ## Date
 
-Contrary to `Time`, a `Date` struct has info about the current date, without any info about current time.
+Contrary to `Time`, a `Date` struct has info about the date, without any info about the time.
 
 ```elixir
 iex> Date.utc_today
@@ -52,7 +52,7 @@ iex> Date.utc_today
 And it has some useful functions to work with dates:
 
 ```elixir
-iex> {:ok, date} = Date.new(2020, 12,12)
+iex> {:ok, date} = Date.new(2020, 12, 12)
 {:ok, ~D[2020-12-12]}
 iex> Date.day_of_week date
 6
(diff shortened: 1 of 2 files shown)


lessons/en/basics/date_time.md
```diff
@@ -42,7 +42,7 @@ But there is a catch: as you may have noticed, this struct only contains the tim
 
 ## Date
 
-Contrary to `Time`, a `Date` struct has info about the current date, without any info about current time.
```
jangromko, line 45: Word "current" may be a bit misleading in this case, because `Date` may contain _any_ date, not only the current.

lessons/en/basics/date_time.md
```diff
@@ -76,7 +76,7 @@ iex(15)> NaiveDateTime.utc_now
 ~N[2029-01-21 19:55:10.008965]
 ```
 
-But it has both the current time and date, so you can play with adding time, for example:
+But it has both the time and date, so you can play with adding time, for example:
```
jangromko, line 79: The same as above — `NaiveDateTime` doesn't have to contain only the _current_ time.

lessons/en/basics/date_time.md
```diff
@@ -100,7 +100,7 @@ iex> DateTime.from_naive(~N[2016-05-24 13:26:08.003], "Etc/UTC")
 
 ## Working with timezones
 
-As we have noted in the previous chapter, by default Elixir does not have any timezone data.
+As we have noted in the previous section, by default Elixir does not have any timezone data.
```
jangromko, line 103: Word "chapter" may be a bit confusing — it may suggest previous lesson. In other places we use "section" and I think it's better.

lessons/en/basics/date_time.md
```diff
@@ -1,5 +1,5 @@
 %{
-  version: "1.1.1",
+  version: "1.1.2",
```
doomspork, line 2: These seem like great changes to include in other languages, let's bump the minor version so it becomes a call out:

```suggestion
  version: "1.2.0",
```

doomspork (changes requested):
This is wonderful @jangromko! One small change which versions but other than that this is good to go. Thank you so much 🎉

lessons/pl/basics/date_time.md
```diff
@@ -0,0 +1,125 @@
+%{
+  version: "1.1.2",
```
doomspork, line 2: We'll want to update this too 
```suggestion
  version: "1.2.0",
```

jangromko:
@doomspork Thanks for the comments! Version numbers have been updated. :)

doomspork (approved):

PR #2741, opened 2021-12-05
add library maintenance with github actions

yordis:


doomspork:
I'm wondering if we should create a whole new section called "Guides" for things like this 🤔

yordis:
@doomspork actually I thought 20,000 about this topic. I have been following strictly https://documentation.divio.com/ lately, and as I was writing this, I couldn't figure out what type of documentation I should write thou.

Is it a Tutorial, How-To, or Explanation?

Also, what would you like it to be as well?

gemantzu (commented):
Looks mostly good to me. @doomspork I can't see from my phone which dir this lives in, but I would assume this should be placed in libraries. Guides sounds like a good idea, but idk if it should be applied in this specific scenario. Let me sleep over it

doomspork:
@yordis would you be willing to revisit this and spin it into a series of blog posts while we figure out where lessons on this could live long term?

doomspork:
Closing for now. Let's circle back to this once we've had a chance to talk through where this fits in with lessons vs blog posts. I'd also like to look at relying less on the strawhat org actions you've built because they abstract away too much and they're no likely the route most companies / projects will take, I'd us to promote the community supported best practice.

Issue #2713, opened 2021-11-05
{% raw %} {% endraw %} in Mnesia code sample


louwers:
https://github.com/elixirschool/elixirschool/blob/master/lessons/en/storage/mnesia.md?plain=1#L279

That doesn't look like Elixir syntax to me. ;-)

doomspork:
Thank you for the report @louwers!

kinson:
Closed in #2746

PR #2746, opened 2021-12-17
Update mnesia.md

monkeypy324:
Remove `{% raw %}` and `{% endraw %}` from `Mnesia.select/2` example.

Seems like an issue was opened already here https://github.com/elixirschool/elixirschool/issues/2713


kinson (approved):
Thanks for raising and fixing this 🚀  (and for your patience)!

PR #2749, opened 2022-01-04
handle_result.({:error}) also returns :ok

drewkitch:


doomspork (approved):
Good catch @drewkitch!

PR #2750, opened 2022-01-10
Update enum.md

benyakirten:
Change headers at Enum using the Capture operator (&), Using the capture operator with an anonymous function, and ### Using the capture operator with a named function so they are not a subset of Enum.uniq


kinson (approved):
Good catch and thanks for the fix! 🚀

PR #2755, opened 2022-01-24
Remove unnecessary `raw` tags in ETS.md

tmr08c:
The `raw` and `endraw` tags are showing up in outputted HTML

Screen Shot 2022-01-24 at 12 12 26 PM

Since none of the other examples include this escaping, I am assuming it's unnecessary.


doomspork (approved):
Thank you @tmr08c!

PR #2759, opened 2022-02-05
[el] Update bypass.md

tenshikiri:
Update to 1.0.1


gemantzu:
@doomspork the original english lesson was bumped in version, while it only had typos as change. maybe we should avoid that in the future?

gemantzu (approved):

PR #2766, opened 2022-02-19
Improve Heirarchy of Headings in Enum Lesson

kinson:
# Overview

This pr updates the structure of the headings used in the enum lesson to match the information hierarchy better. 

Instead of starting with a level heading that matches the name of the lesson, this structure uses an `# Overview` heading instead.

I also nested the functions into a common level one heading of `# Common Functions`.

Let me know if you have any suggestions for the section names.

### DIFF
--- a/lessons/en/basics/enum.md
+++ b/lessons/en/basics/enum.md
@@ -1,5 +1,5 @@
 %{
-  version: "1.8.1",
+  version: "1.9.0",
   title: "Enum",
   excerpt: """
   A set of algorithms for enumerating over enumerables.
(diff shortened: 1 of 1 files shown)


doomspork (approved):

lessons/en/basics/enum.md
```diff
@@ -1,13 +1,13 @@
 %{
-  version: "1.8.0",
+  version: "1.8.1",
```
doomspork, line 2: I think this should be 1.9 since it is a change we’d like to see made in the translations

doomspork (changes requested):

doomspork (approved):

PR #2767, opened 2022-02-20
[JA] add testing/mox v1.0.1

koga1020:
Added mox Japanese lessons.


doomspork (approved):
Thank you @koga1020!!

PR #2781, opened 2022-03-25
Update plug.md to use Config module

miguelbermudez:
Use `Config` module since `Mix.Config` was removed in Elixir 1.9. 
Only reference I could find for supported Elixir versions was 1.5 and above. 
Not sure if this the established way of handling deprecations but I thought it might help someone.


doomspork (changes requested):
Thank you and welcome @miguelbermudez! Apologizes for the delay, could you please update the version of this document (top of the document) to 2.3.0?

miguelbermudez:
@doomspork, no worries. Please let me know if you need anything else.

lessons/en/misc/plug.md
```diff
@@ -336,7 +347,17 @@ defmodule Example.Application do
...
+  # For Elixir < 1.9
   defp cowboy_port, do: Application.get_env(:example, :cowboy_port, 8080)
+  
+  # For Elixir 1.9+
+  defp cowboy_port do
+    case Application.fetch_env(:example, :cowboy_port) do
+      {:ok, value} -> value
+      :error -> 8080
+    end
+  end
```
kinson, line 359: Perhaps I am missing something here, but is there something incorrect with using `Application.get_env/3`? If not, I think it may be confusing to indicate that it should not be used in Elixir 1.9+

lessons/en/misc/plug.md
```diff
@@ -311,12 +311,23 @@ It's considered good practice to make the port configurable by putting it in a c
...
 
+For Elixir versions below 1.9:
+
 ```elixir
 use Mix.Config
 
 config :example, cowboy_port: 8080
 ```
 
+For Elixir version 1.9+ (`Mix.Config` was removed in Elixir 1.9):
```
kinson, line 322:
```suggestion
For Elixir version 1.9+ (`Mix.Config` was deprecated in Elixir 1.9):
```

Correct me if I am wrong, but I believe this was not removed just deprecated with 1.9

kinson (commented):
Thanks for the updates! Just two suggestions/questions 🙂

doomspork:
@miguelbermudez could you review @kinson feedback please? 🙏

lessons/en/misc/plug.md
miguelbermudez: @kinson Apologies, this was a long time ago and I'm not sure what my rationale was at that point. I can't recall why I thought `fetch_env/2` would be better for 1.9+ than `get_env/3`. I'll change this back to use just `get_env/3`.

miguelbermudez:
@doomspork all set!

kinson (approved):

PR #2784, opened 2022-03-29
[PL] Update translation of lesson "Executables"

jangromko:


doomspork (approved):
Thank you @jangromko!

PR #2787, opened 2022-04-02
Identification of data type error

BryanJBryce:
The data structure being passed in is a map not an `Actor` struct


SophieDeBenedetto (approved):
Thanks for the fix!

PR #2790, opened 2022-04-22
Fix a bug in producer-consumer example snippet

shinderohitt:


doomspork (approved):
Thank you and welcome @shinderohitt!

PR #2794, opened 2022-04-28
Pattern Matching should always return something on iex

adolfont:
I have just added "Sean" as the return on iex to:
iex> name = "Sean"


doomspork (approved):
Thank you @adolfont! I will retroactively apply this to other languages and lessons 🎉

adolfont:
I have already asked to change this in the Portuguese version. I should have started here.

PR #2798, opened 2022-05-08
OTP Supervisors: Make correction mentioning 0 as a valid time out + Fix Lists Indentation

eksperimental:
Timeout could be `0`.

### DIFF
--- a/lessons/en/advanced/otp_supervisors.md
+++ b/lessons/en/advanced/otp_supervisors.md
@@ -1,5 +1,5 @@
 %{
-  version: "1.1.1",
+  version: "1.1.2",
   title: "OTP Supervisors",
   excerpt: """
   Supervisors are specialized processes with one purpose: monitoring other processes.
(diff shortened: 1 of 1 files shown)


eksperimental:
Additionally it fixes the list indentation.
I think you should review your guides if you are using this style.

-----------

This is how it looks now,
image

-----------
This is how it looks after the PR

image

eksperimental:
I just found out that shutdown value of `0` is only valid on OPT 24.

kinson (approved):
Thanks for the correction and formatting update 🚀 !

eksperimental:
Do we want to leave 0 as a valid value, since it's only available since OTP 24 and OTP 20-23 are currently supported by the supported versions of Elixir.

kinson:
If you want to open a pr to add a note about that compatibility concern that would be great! Do you mind sharing the reference you are using?

Thanks again for helping out 🚀

eksperimental:
https://github.com/elixir-lang/elixir/pull/11836

Apparently the spec included `0`, but in reality that would not work.
Anyway, the Elixir specs don't include `0` and will not include it until OTP 24 is supported exclusively, so maybe the PR should remove the `0` for the time being.

Issue #2788, opened 2022-04-10
Pipe operator used before it's introduced


snus-kin:
Hi,

Nothing major, just reading through the lessons and the pipe operator is used in examples on the functions page before it's introduced by the next page, kind of just breaks the flow is all in my opinion. 

[Functions Lesson](https://elixirschool.com/en/lessons/basics/functions)

allefgomes:
Yeah! You are right.
I saw that in [guards session](https://elixirschool.com/en/lessons/basics/functions#guards-6) the pipe operator appear and don't have a previews explanation about it.

The sessions that use pipe operator are:
[Guards](https://elixirschool.com/en/lessons/basics/functions#guards-6) 
[Default Arguments](https://elixirschool.com/en/lessons/basics/functions#default-arguments-7)

I think that we have two options:
1. We can put a simple signification like "The pipe operator `|>` is to send the result of function A to function B as the first argument. You will understood more about pipe operator in the next lesson" .
2. The Guards and Default Arguments sessions can be after introduction of pipe operator.

snus-kin:
Personally (and as someone who is not familiar with Elixir) it seems like a fairly important part of the language's syntax. Perhaps something could be mentioned (as to the effect of option 1) in the first chapter? 

From the Pipe Operator page:
"The pipe operator |> passes the result of an expression as the first parameter of another expression." 

This seems like a fairly simple and intuitive explanation that could be included, perhaps then reference there is more detail in the dedicated chapter to it.

allefgomes:
@doomspork can you close this issue?
It was finally at #2799

doomspork:
Done, thank you again @allefgomes!

doomspork:
@allefgomes we should backport this change to the other languages. Would you like to do that? If not I'm happy to do so.

allefgomes:
> @allefgomes we should backport this change to the other languages. Would you like to do that? If not I'm happy to do so.

I can do it in the next week :)

PR #2799, opened 2022-05-11
Pipe operator used before it's introduced

allefgomes:
As mentioned in the #2788 the pipe operator is used in examples on the functions page before it's introduced by the next page.

Here, I create a simple introduction about pipe operator and a link to see about.

### DIFF
--- a/lessons/en/basics/functions.md
+++ b/lessons/en/basics/functions.md
@@ -315,7 +315,8 @@ iex> Greeter.phrase
 ### Guards
 
 We briefly covered guards in the [Control Structures](/en/lessons/basics/control_structures) lesson, now we'll see how we can apply them to named functions.
-Once Elixir has matched a function any existing guards will be tested. In the following examples we will use the [Pipe Operator](/en/lessons/basics/pipe_operator) `|>` that passes the result of an expression as the first parameter of another expression.
+Once Elixir has matched a function any existing guards will be tested.
+In case you didn't see or don't understand the pipe operator `|>`, you must check in this [lesson](/en/lessons/basics/pipe_operator).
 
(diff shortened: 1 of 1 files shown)


lessons/en/basics/functions.md
```diff
@@ -315,7 +315,7 @@ iex> Greeter.phrase
 ### Guards
 
 We briefly covered guards in the [Control Structures](/en/lessons/basics/control_structures) lesson, now we'll see how we can apply them to named functions.
-Once Elixir has matched a function any existing guards will be tested.
+Once Elixir has matched a function any existing guards will be tested. In the following examples we will use the [Pipe Operator](/en/lessons/basics/pipe_operator) `|>` that passes the result of an expression as the first parameter of another expression.
```
herminiotorres, line 318: hi Allef, thank you for your contribution.
IMHO, I think, it would be better if you said something like:

In case you didn't see or don't understand the pipe operator, you must check in this [lesson](/en/lessons/basics/pipe_operator). 

WDYT?

allefgomes: It sounds better. Thank you!

doomspork (changes requested):
Thank you and welcome @allefgomes! In this situation rather than linking to a future page and potentially taking someone away from finishing this lessons is not ideal. Instead, I would rather we update the code examples to not leverage the pipe:


```elixir
  def hello(names) when is_list(names) do
    names = Enum.join(names, ",")
    hello(names)
  end
```

What do you think?

allefgomes:
It's a better solution. Thank you for suggestion.
I will change and update the code.

doomspork (approved):
Thank you for the speedy turn around @allefgomes!

PR #2801, opened 2022-05-17
Updated collections.md to avoid confusing usage of verb "accessing"

mjvandermeulen:
When talking about key value pairs the verb "accessing" generally means accessing the value with the key. It caused me some confusion when I read it. I have removed this word, to avoid this confusion, and replaced it with plain easy to read english.


doomspork (approved):
Thank you and welcome @mjvandermeulen!

PR #2803, opened 2022-05-20
Remove pipe operator before its intruduced

allefgomes:
As commented at #2799, update examples in other languages.

@doomspork :)


doomspork (approved):
Nice work @allefgomes! 🚀

PR #2806, opened 2022-05-24
[ko] Add a translation of "Testing:Mox"

bangalcat:


doomspork (approved):
Thank you @bangalcat! I've merged all of these so they don't sit longer without review. If other contributors want to update these we can open up patch PRs.

Thank you again!

PR #2808, opened 2022-05-31
[PL] Update translation of "ETS"

jangromko:
Updated Polish translation of lesson "Erlang Term Storage".


doomspork (approved):
Thank you @jangromko!

PR #2819, opened 2022-07-13
Update otp_distribution.md

anthony-zhou:
Updated the tutorial config files to use `import Config` and `config_env()` instead of `use Mix.Config` and `Mix.env()`, since `Mix.Config` is [deprecated](https://hexdocs.pm/mix/main/Mix.Config.html).


lessons/en/advanced/otp_distribution.md
```diff
@@ -1,5 +1,5 @@
 %{
-  version: "1.0.2",
+  version: "1.0.3",
```
doomspork, line 2: Let's bump to 1.1.0 to ensure translators add this

```suggestion
  version: "1.1.0",
```

doomspork (approved):
Thank you and welcome @anthony-zhou!

doomspork (approved):

PR #2824, opened 2022-08-01
Fix deploy_content action

SophieDeBenedetto:
This PR updates the deploy_content action to target the correct `main` branch of the `elixirschool/school_house` repo. It was targeting the `master` branch which does not exist.


ojizero:
@SophieDeBenedetto maybe it's worth it adding [`--fail`](https://curl.se/docs/manpage.html#-f) or [`--fail-with-body `](https://curl.se/docs/manpage.html#--fail-with-body) option to cURL as to have it fail the pipeline instead of being silent about the error 🤔

doomspork (approved):

doomspork:
@ojizero can you add that in a follow up PR?

ojizero:
@doomspork sure thing I'll send a follow up PR in a few!

PR #2826, opened 2022-08-01
Fix options to match current cURL version

ojizero:
The version of cURL found in the GitHub runner doesn't support `--fail-with-body` so we should use `--fail` instead.


ojizero:
@doomspork turns out the version of cURL the GitHub runner has doesn't recognize the `--fail-with-body` option, I switched it to use the `--fail` option as it should be supported (tested it in an Ubuntu image locally with the same version of cURL as that documented by GitHub, `7.68.0-1ubuntu2.12`).

doomspork (approved):
Thanks for catching this and following up @ojizero! 🙏

PR #2828, opened 2022-08-01
[PL] Update translation of lesson "Enum" (+ minor change in EN)

jangromko:
1. Polish translation of lesson "Enum" has been updated to reflect the latest EN version (in particular changes from #2766).
2. Added missing newlines after two sub-chapters titles in EN version.


doomspork (approved):
Much appreciated as always @jangromko 🙏

PR #2829, opened 2022-08-01
[PL] Update title & excerpt in "Mix tasks" lesson

jangromko:
Title and excerpt of lesson "Mix tasks" have been updated to respect rules of the Polish language.


doomspork (approved):
🚀

PR #2831, opened 2022-08-11
[FR] Translation of bascis/strings lesson

paulhenri:
Hello,

Translation for the strings lesson, in French. First time doing this so do not retain from asking me to modify things.
Somehow the translation was merged in a previous version of the school but it did not went through the new version.

Side note : If the translation group is active i am willing to translate more English lessons to French.

Paul-H


lessons/fr/basics/strings.md
```diff
@@ -0,0 +1,188 @@
+---
+version: 1.2.0
+title: Chaines
+---
```
doomspork, line 4: We use a different syntax for this header information now. I think this should do the trick and build correctly:

```suggestion
%{
  version: "1.2.0",
  title: "Chaines",
  excerpt: """
  Strings, Char Lists, Graphemes and Codepoints.
  """
}
```

doomspork (changes requested):
Hello and welcome @paulhenri! We only have a very small @elixirschool/french team and they're often very busy but always happy to grow the team with more Elixir fans! 

One small change to get the build to pass due to us migrating to a new format for header information in the markdown files.

PR #2838, opened 2022-08-20
Update mix_tasks.md

chriseyre2000:
Explain the mapping between the module name and the name of the task.

It is documented in Mix.Tasks, but would be clearer here.

### DIFF
--- a/lessons/en/intermediate/mix_tasks.md
+++ b/lessons/en/intermediate/mix_tasks.md
@@ -145,3 +145,5 @@ mix hello             # Simply calls the Hello.say/0 function.
 
 Note: Our code must be compiled before new tasks will appear in the `mix help` output.
 We can do this either by running `mix compile` directly or by running our task as we did with `mix hello`, which will trigger the compilation for us.
+
+The name of the task is whatever is after the Mix.Tasks in the module name. So `MixTasks.MyHelper.Utility` becomes `my_helper.utility`.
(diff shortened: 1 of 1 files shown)


lessons/en/intermediate/mix_tasks.md
```diff
@@ -145,3 +145,5 @@ mix hello             # Simply calls the Hello.say/0 function.
 
 Note: Our code must be compiled before new tasks will appear in the `mix help` output.
 We can do this either by running `mix compile` directly or by running our task as we did with `mix hello`, which will trigger the compilation for us.
+
+The name of the task is whatever is after the Mix.Tasks in the module name. So `MixTasks.MyHelper.Utility` becomes `my_helper.utility`.
```
doomspork, line 149: What do you think about rephrasing this slightly?

```suggestion
It's important to note that task names are derived from the module name, so `MixTasks.MyHelper.Utility` will become `my_helper.utility`.
```

doomspork (changes requested):
Thank you for adding this important note @chriseyre2000! I had one thought on a slight rephrasing but more importantly could you bump the version at the top of the file from `1.2.0` to `1.3.0` so this shows up in the translation report?

chriseyre2000 (commented):
Updated version number.

chriseyre2000:
I have applied your changes.

I only added this as I had got caught out trying to use conventional module
names and ended up with a strange looking mix command.

doomspork (approved):
Thank you @chriseyre2000 🎉

PR #2843, opened 2022-08-26
Update pipe_operator.md coding syntax

joao-dantas:
While reading the docs, I saw that the code snippets on this page were different colors than previous pages. They had a gray color that was hard to read. While looking at the code I saw they were using `shell` syntax instead of `elixir`.
This PR aims to standardize it.


anderkonzen (approved):
Hi @joao-dantas ! Thanks for fixing this! It's always nice to have docs sync'ed and updated.

PR #2868, opened 2022-10-25
Fix Grammer

ChandanChainani:


lessons/en/basics/functions.md
```diff
@@ -194,14 +194,14 @@ iex> fred = %{
 %{name: person_name}
 ```
 
-In `Greeter1.hello/1`, the map we pass (`fred`) is evaluated against our argument (`%{name: person_name}`):
+In `Greeter1.hello/1`, the map we pass (`fred`) is evaluated against function argument (`%{name: person_name}`):
```
doomspork, line 197: I think the previous text was sufficient but if we want to make it clearer we need to correct the grammar in this change:

```suggestion
In `Greeter1.hello/1`, the map we pass (`fred`) is evaluated against our function argument (`%{name: person_name}`):
```

lessons/en/basics/functions.md
```diff
@@ -263,11 +263,11 @@ So now when we call `Greeter2.hello/1`, we can use all of Fred's information:
     iex:15: Greeter2.hello/1
 ```
 
-So we've seen that Elixir pattern-matches at multiple depths because each argument matches against the incoming data independently, leaving us with the variables to call them by inside our function.
+So we've seen that Elixir pattern-matches at multiple depths because each argument matches against the incoming data independently, leaving us with the variables that we can use inside the function.
```
doomspork, line 266: Appreciate the suggestion but not sure this is necessarily correct grammar, how about:

```suggestion
So we've seen that Elixir pattern-matches at multiple depths because each argument matches against the incoming data independently, leaving us with variables we can use inside our function.
```

doomspork (changes requested):
Thank you @ChandanChainani! I have some feedback we should review. Also, please bump the version of the file 🙏

PR #2872, opened 2022-11-12
Update genstage.md

chanphiromsok:
At line 113 It should be GenStage not GenServer


doomspork (changes requested):
Thank you and welcome @chanphiromsok! Could you please bump the version at the top of the file? Today it is "1.1.1" and I think "1.2.0" would be good here 👍 

Alternatively you could apply this change to all the translations impacted.

doomspork (approved):

PR #2879, opened 2022-12-01
Update mnesia.md

ByeongUkChoi:
Remove `{% raw %}` and `{% endraw %}` from `Mnesia.select/2` example.

associated https://github.com/elixirschool/elixirschool/pull/2746


gidskql6671 (approved):
감사합니다~

PR #2883, opened 2022-12-19
Update nimble_publisher.md

vanderhoop:
This pull request updates the index.html.eex template within the `nimble_publisher` lession to use `ul` and `li` elements, as each item in the blog index represents a list item

I considered using an `ol` element due to the default sorting of the blog posts (date, descending), but semantically this didn't feel correct, as users generally don't have to consume blog posts in a particular order.


jaimeiniesta (approved):
That's a nice detail, LGTM 👍

PR #2892, opened 2023-02-09
functions.md added unexpected statement.

abdullahb53:
Normally, we should cover all possible statements as we used on erlang.

### DIFF
--- a/lessons/en/basics/functions.md
+++ b/lessons/en/basics/functions.md
@@ -410,6 +410,8 @@ defmodule Greeter do
   def hello(name, language_code) when is_binary(name) do
     phrase(language_code) <> name
   end
+  
+  def hello(_), do:  "Unexpected statement."
 
   defp phrase("en"), do: "Hello, "
   defp phrase("es"), do: "Hola, "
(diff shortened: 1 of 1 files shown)


doomspork:
Hello and thank you for the suggestion @abdullahb53! Unfortunately, I'm not sure this is the right change for this lesson. Earlier in the lesson we cover `** (FunctionClauseError) no function clause matching in ...` so if we add anything I think it should be a final call out to the aforementioned error if you pass in something like an number.

If we're going to add in a function to catch-all it should have a more substantial call out than just appearing in the final example.

Thoughts?

abdullahb53:
Hello,
While i reading your message, i notice that adding these pattern-matching things is getting off topic. Actually I learned erlang 2 months ago (it took 1 month) and i was forced by the erlang documentation for doing all the capturing (pattern-match). It is important not to deviate from the heart of the matter. So this learning model is less confusing.

You are right.

doomspork:
@abdullahb53 thanks for the reply! I'll close this for now but please do keep up the suggestions, always looking to improve our content for others 🙏

PR #2893, opened 2023-02-11
logger-changed-to-LoggerExample-and-env-deleted

abdullahb53:
"Logger" module name is not working.  We have to change module name Logger to another because "elixir" already have "Logger" module name in itself.

Error : {"init termiLogger - error: nating in do_boot",{undef,[{'Elixir.Logger',flush,[],[]},{'Elixir.{reKermoved_failing_filter,process_level} nel.CLI',run,1,[{file,"lib/kernel/cli.ex"},{line,64}]},{init,start_em,1,[{file,"init.erl"},{line,1220}]},{init,do_boot,3,[{file,"init.erl"},{line,910}]}]}} init terminating in do_boot ({undef,[{Elixir.Logger,flush,[],[]},{Elixir.Kernel.CLI,run,1,[{_},{_}]},{init,start_em,1,[{_},{_}]},{init,do_boot,3,[{_},{_}]}]})

### DIFF
--- a/lessons/en/advanced/metaprogramming.md
+++ b/lessons/en/advanced/metaprogramming.md
@@ -110,23 +110,26 @@ This is different from other languages where there is still the overhead of a fu
 To demonstrate this we'll make a simple logger that can either be enabled or disabled:
 
 ```elixir
-defmodule Logger do
+defmodule LoggerExample do
   defmacro log(msg) do
-    if Application.get_env(:logger, :enabled) do
+
       quote do
-        IO.puts("Logged message: #{unquote(msg)}")
+        IO.puts("[LOGGER]- #{Time.utc_now()} : #{unquote(msg)}")
       end
-    end
+
   end
 end
 
 defmodule Example do
-  require Logger
+  require LoggerExample
 
-  def test do
-    Logger.log("This is a log message")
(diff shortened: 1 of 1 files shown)


doomspork:
Howdy @abdullahb53 👋 How are you trying to run this code in order to get this error?

When running this in `iex` you will get a warning `warning: redefining module Logger` but you should not be getting this kind of error 🤔 

I'm open to changing the module name but I don't believe the other changes here are necessary. Thoughts?

abdullahb53:
Hi @doomspork , i learned from you that similar defmodule name will trigger new module defination. Sorry, I didn't know this. Now, I execute my code and i see normal behaviour that like u said. There is a no problem. Probably, I was not able to catch the right module name. One week before, i didn't use 'iex'. I was used to write c("...") like a erlang shell. and i probably overlooked right file name.
My code was like that
"
defmodule Logger do
  defmacro log(msg) do

      quote do
        IO.puts("[LOGGER]- #{Time.utc_now()} : #{unquote(msg)}")
      end

  end
end

defmodule Example do
  require Logger

  def test(text \\ "DEACTIVE") do
    Logger.log("#{text}")
  end
end


Example.test("This")
"

PR #2895, opened 2023-02-13
Fix grammar

igorgbr:


doomspork:
Hello and welcome @igorgbr! Thanks so much for the update _and_ for taking time to update the version metadata 🙏

doomspork (approved):

PR #2897, opened 2023-02-15
[KO] Fix broken lists in advenced/otp_distribution

39hn:
fix broken lists in lessons/advanced/otp_supervisors


gidskql6671 (approved):
Thank you 👍

PR #2899, opened 2023-02-18
Fixes Plug's tutorial, Unfetched error

joseildofilho:
every time I'm getting a Unfetched error, so I need to fetch by hand


doomspork:
Howdy @joseildofilho! I do not believe this change is necessary and there may another component missing from your project, is your code on GitHub somewhere?

In this same lesson, when creating the router, we use `plug Plug.Parsers, parsers: [:urlencoded, :multipart]` which handles `fetch_query_params/2` for you per the docs here: https://hexdocs.pm/plug/Plug.Parsers.html.

Can you take another look at your code to see if the aforementioned parser plug is present?

PR #2910, opened 2023-03-13
Update bypass.md

msutkowski:
Small grammar corrections for readability


doomspork (approved):
Thank and welcome @msutkowski! We look forward to many more contributions from you as you dive deep into the world of Elixir 🎉

PR #2927, opened 2023-05-20
Update collections.md

soneyworld:
Maybe inside the Maps description it should be "Maps" instead of "Listen"


doomspork (approved):
Thank you and welcome @soneyworld!

PR #2929, opened 2023-05-21
Fix maps called lists in collections.md

hc42:
Small fix in maps definition where maps are named lists.


doomspork:
Thank you @hc42! Apologizes for the delay in getting back to you. This has been fixed in an earlier PR that has been merged. As such I'm closing this but please don't let that stop future contributions!

PR #2931, opened 2023-05-30
Update associations.md to use primary_key: false on migration

emarchak:
I was following this tutorial, but when trying to populate the associations I kept running into a postgresql error complaining about a `column id not null violation`

I didn't realize what was going on until I looked at the database, and the recommended migration adds an `id` column to the `movies_actors` table, which I was leaving null during the `put_assoc/3` call!

Easy fix here, as other n00bs might run into this.


gf3:
nice find!

doomspork (approved):
Good find indeed! Thanks @emarchak 🙏

PR #2932, opened 2023-06-02
[FA] Translated Enum

a5hk:


doomspork (approved):
Thank you @a5hk! Hope you are well

PR #2937, opened 2023-06-25
Update sigils.md with new multi-character syntax from 1.15

skogsbrus:
This commit adds information and a new section on multi-character sigils, supported in Elixir 1.15.

Minor changes are made to the definition of sigils to support this new syntax and a new section is added to provide an example.

Slight corrections are also made to the iex output in some examples, tested with Elixir 1.14 and 1.15.


doomspork:
@skogsbrus welcome! This is awesome 🤯 Would you mind bumping the version at the top of the file so the change is flagged for the translators?

skogsbrus:
@doomspork Thanks for the tip, done. Assuming this is a minor version bump.

doomspork (approved):

PR #2939, opened 2023-07-11
Update nimble_publisher.md to phoenix version 1.7

scubass:


lessons/en/misc/nimble_publisher.md
```diff
@@ -1,5 +1,5 @@
 %{
-  version: "1.0.2",
+  version: "2.0.0",
```
doomspork, line 2: @scubass can provide some additional context on why we need this version change?

scubass: Phoenix changed a lot and is not compatible with the other version code, so I thought that the version should reflect the major update to the library.

doomspork: This version reflects the _lesson's_ version to notify translators when changes need to be made.

doomspork:
This change doesn't appear to be necessary so I'm closing this for now. The lesson versions reflect the version of the lesson content. We shouldn't need to update that without updating the content itself.

PR #2940, opened 2023-08-10
fix: split code block in English functions lesson

takclark:


takclark:
NVM that's the output

PR #2942, opened 2023-09-02
update Rocket Validator post

jaimeiniesta:
Rocket Validator now has a different [Stats](https://rocketvalidator.com/blog/stats) section that no longer has the feature of shared domain stats that was referenced in this blog post.

This PR removes references and links to the old stats so we don't get a 404.


doomspork (approved):

doomspork:
Thanks for circling back to update! Hope you're doing well @jaimeiniesta

PR #2945, opened 2023-09-12
Remove input count from iex example

Markcanfly:
In the NaiveDateTime example, the code snippet includes the input count for iex. This serves no purpose, and might be confusing.

### DIFF
--- a/lessons/bn/basics/date_time.md
+++ b/lessons/bn/basics/date_time.md
@@ -1,5 +1,5 @@
 %{
-  version: "1.1.1",
+  version: "1.1.2",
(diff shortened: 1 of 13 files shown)


anderkonzen:
Hi @Markcanfly ! Thanks for that! I was checking other languages and it seems they also present the same problem. Would you be so kind to fix this to other languages as well? This way we keep all of them in sync.

julienXX (approved):

anderkonzen (approved):

ByeongUkChoi (approved):

erickgnavar (approved):

PR #2953, opened 2023-10-03
[EN] Update mix_tasks.md to fix the formatting

marcsvll:


gemantzu (commented):
@doomspork do we bump versions in this case?

doomspork (approved):
@gemantzu patch is unnecessary but is fine since it doesn't impact translations

PR #2962, opened 2023-10-23
fix: spelling of map

m4hi2:


wasi0013 (approved):
Thanks, for the fix! @m4hi2 bhai.

PR #2968, opened 2023-11-14
Patch/fix error message

Kuroki-g:
`42 and true` in Boolean section no longer throws `ArgumentError`.

Tested in the following environment.

- Erlang/OTP 26 [erts-14.1.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]
- Interactive Elixir (1.15.7) 

```
iex(1)> 42 and true
** (BadBooleanError) expected a boolean on left-side of "and", got: 42
    iex:1: (file)
```

### DIFF
--- a/lessons/en/basics/basics.md
+++ b/lessons/en/basics/basics.md
@@ -1,5 +1,5 @@
 %{
-  version: "1.3.0",
+  version: "1.3.1",
   title: "Basics",
   excerpt: """
   Getting started, basic data types, and basic operations.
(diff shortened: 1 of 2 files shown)


lessons/en/basics/basics.md
```diff
@@ -1,5 +1,5 @@
 %{
-  version: "1.3.0",
+  version: "1.3.1",
```
doomspork, line 2:
```suggestion
  version: "1.4.0",
```

lessons/ja/basics/basics.md
```diff
@@ -1,5 +1,5 @@
 %{
-  version: "1.3.0",
+  version: "1.3.1",
```
doomspork, line 2:
```suggestion
  version: "1.4.0",
```

doomspork (changes requested):
Welcome and thank you @Kuroki-g! This is great, thank you for making this change.

I think we should bump these to 1.4 so it triggers a notice for other maintainers.

PR #2974, opened 2024-01-03
Update documentation.md

noworry-b-still:
Version : `1.2.0`

```elixir
iex> h Greeter
Greeter was not compiled with docs
```

- Add explanation/notes for above issue that is encountered during following cases : 
   - if `h greeter` is executed right after module compilation in `iex`
   - if `c/1` is used


doomspork (commented):
Thank and welcome @lord-cosmos! Could you bump the version at the top from `1.1.1` to `1.2.0`?

lord-cosmos:
I am new to contributing. Can you please let me know what do you mean by bumping the version? 
I added a version to the PR comment. 

Thanks.

doomspork (approved):

PR #2976, opened 2024-01-14
Update pattern_matching.md

thiagonebuloni:
There's a typo in line 59 that makes the sentence meaningless.

The correct translation of the sentence "When we pin a variable *we* match on the existing value rather than rebinding to a new one."

...is the one suggested in the commit:
"Quando fixamos uma variável, *nós* a associamos ao valor existente ao invés de reassociar a um novo valor."


anderkonzen (approved):
SGTM! Thanks @thiagonebuloni !

PR #2981, opened 2024-03-21
Update pattern_matching.md

codecharlan:
Simplify understanding of pinning in Function Definitions.


doomspork (approved):
Thank you and welcome @codecharlan! I updated the min version of the lesson because this change feels important enough for translators to pick it up.

PR #2983, opened 2024-03-28
Fix typo in modules.md

noamsauerutley:
The proposed change drops an unneeded apostrophe to convert "let's" to the grammatically correct "lets"


doomspork (approved):
Thank you @noamsauerutley!

PR #2987, opened 2024-05-31
Update modules.md

aqdd:
I'm a newbee, just i think it should be `as:`, not `:as`, please forgive me if I'm wrong.

### DIFF
--- a/lessons/en/basics/modules.md
+++ b/lessons/en/basics/modules.md
@@ -167,7 +167,7 @@ defmodule Example do
 end
 ```
 
-If there's a conflict between two aliases or we just wish to alias to a different name entirely, we can use the `:as` option:
+If there's a conflict between two aliases or we just wish to alias to a different name entirely, we can use the `as:` option:
 
 ```elixir
 defmodule Example do


kinson (commented):
Depending on how you read this, I think either option `:as` or `as:` could work. The reason that the author presumably wrote this with `:as` is that it is used in a [Keyword list](https://hexdocs.pm/elixir/1.16.3/Keyword.html) where technically, each element is composed of a two element tuple (e.g. `[{:as, Hi}]`). But, Elixir also provides a special syntax for Keyword lists that lets us make this more concise (e.g. `[as: Hi]`). Ultimately, the `:as` here is still an atom though.

So, with that said, I would probably keep this sentence as is.

Here is the documentation explaining that syntax: https://hexdocs.pm/elixir/1.16.3/Keyword.html#module-examples

kinson:
And to be 100% clear, I don't mean to discourage sincere feedback, we always appreciate a keen eye reading through the guides! :octocat: 

If you, or someone else, disagrees with my conclusion feel free to re-open this pr and we can discuss further.

PR #2994, opened 2024-06-09
Update pattern_matching.md

yesuko:
Fix pattern matching and demonstrate successful match with pin operator

- Updated the example to correctly demonstrate the use of the pin operator (^) in pattern matching.
- Initially, the pattern `{x, ^x} = {2, 1}` failed with a `MatchError` as expected.
- Added a successful pattern match example `{x, ^x} = {2, 2}` to show the correct usage.
- Confirmed that after a successful match, the variable `x` retains the value `2`.

This commit clarifies the behavior of the pin operator in pattern matching, highlighting both an expected failure and a successful match.

### DIFF
--- a/lessons/en/basics/pattern_matching.md
+++ b/lessons/en/basics/pattern_matching.md
@@ -68,7 +68,9 @@ iex> x = 1
 iex> ^x = 2
 ** (MatchError) no match of right hand side value: 2
 iex> {x, ^x} = {2, 1}
-{2, 1}
+** (MatchError) no match of right hand side value: {2, 1}
+iex> {x, ^x} = {2, 2}
+{2, 2}
 iex> x
 2
 ```


lessons/en/basics/pattern_matching.md
```diff
@@ -68,7 +68,9 @@ iex> x = 1
 iex> ^x = 2
 ** (MatchError) no match of right hand side value: 2
 iex> {x, ^x} = {2, 1}
-{2, 1}
+** (MatchError) no match of right hand side value: {2, 1}
```
kinson, line 71: In this example, `{x, ^x} = {2, 1}`, the second `^x` only acts as a pin, so the result of this is re-binding the first element of the tuple `{2, 1}`, which is why `x` evaluates to 2 in the next statement. 

In short, the example as is works as intended (removing the `MatchError` example for brevity):

```elixir
iex> x = 1
1
iex> {x, ^x} = {2, 1}
{2, 1}
```

kinson (commented):
Thanks for the close eye on these examples 🚀 
Let me know if you have questions about my comment.

lessons/en/basics/pattern_matching.md
yesuko: ok, got it. Thanks

PR #2995, opened 2024-06-11
Correct a typo

danadaldos:
Here you would want to use "its" (which is possessive) rather than the contraction of "it is".


kinson (approved):
### 🔍 👀 

Good catch!

PR #3022, opened 2024-10-29
fix: clarify wording in control structures lesson

rudiejd:
I found this wording to be a bit confusing - I couldn't figure out whether "it" was referring to `case` or `else`. Tried to clarify a bit


kinson (approved):
🚀  Thanks for the clarification!

PR #3023, opened 2024-11-07
Tweak text for atom-maps special syntaxes

riffraff:
I had to read the text a couple times cause I thought there was a missing word. I find this a bit easier to understand but YMMV :)


kinson (approved):
Thanks for the contribution 🫶 — I agree that the sentences are easier to read with these changes!

@doomspork should we bump the patch version of the lesson for this change?

doomspork:
@kenspirit normally but the patch version is just for record keeping really, since it doesn't do anything with the translator report.

PR #3028, opened 2024-12-14
Update otp_distribution.md

krishna-404:
I received this error without the `-S mix`

```
** (EXIT from #PID<0.116.0>) shell process exited with reason: an exception was raised:
    ** (UndefinedFunctionError) function Kate.say_name/0 is undefined (module Kate is not available)
        Kate.say_name()

20:46:06.031 [error] Process #PID<13355.114.0> on node :ate@localhost raised an exception
** (UndefinedFunctionError) function Kate.say_name/0 is undefined (module Kate is not available)
    Kate.say_name()
```

Here is what I got from Claude

```
You need to start both nodes with iex -S mix to load the application code. You started the second node with just iex --sname, which doesn't load your project.
```

I am new to Elixir & am not sure this is completely right. Please ignore if thats the case.

### DIFF
--- a/lessons/en/advanced/otp_distribution.md
+++ b/lessons/en/advanced/otp_distribution.md
@@ -19,14 +19,14 @@ A node is any Erlang runtime system that has been given a name.
 We can start a node by opening up `iex` session and naming it:
 
 ```bash
-iex --sname alex@localhost
+iex --sname alex@localhost -S mix
 iex(alex@localhost)>
 ```
 
 Let's open up another node in another terminal window:
 
 ```bash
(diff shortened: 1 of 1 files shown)


anderkonzen:
Hi @krishna-404 !

The error you see is likely because you are trying to run the `iex` command inside a mix project. When you run `iex --sname alex@localhost`, for example, in a folder that is not a mix project it should work as expected.

I'll close this PR as this is not applicable. I hope you're enjoying the ElixirSchool content, and thanks for the interest in the project.

PR #3030, opened 2025-01-02
Update control_structures.md for better case/2 clarity

bommar-lu:
Added variable assignment before case/2 examples. As a first time Elixir learner, I was confused by the use of the tuple and introduction to _, as it was unclear to me that "case {value} do" was using the example value (i.e. {:ok, "Hello World"}, :even) to pattern match. I added a variable assignment before as this is generally as I see case being useful

### DIFF
--- a/lessons/en/basics/control_structures.md
+++ b/lessons/en/basics/control_structures.md
@@ -1,5 +1,5 @@
 %{
-  version: "1.2.0",
+  version: "1.1.3",
   title: "Control Structures",
   excerpt: """
   In this lesson we will look at the control structures available to us in Elixir.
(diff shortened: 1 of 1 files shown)


lessons/en/basics/control_structures.md
```diff
@@ -1,5 +1,5 @@
 %{
-  version: "1.1.2",
+  version: "1.2.0",
```
doomspork, line 2:
```suggestion
  version: "1.1.3",
```

doomspork (approved):
Thank you @bommar-lu! I'm doing to tweak the version # due to how that works with translations but this looks great, thank you

doomspork (approved):

PR #3036, opened 2025-02-04
Update enum.md

jl33-ai:
**Change**
Introduce `map` before `each` for more easy to follow ordering.

**Why?**
While this is an extreme nitpick, I did find the original ordering slightly unintuitive. 

When I read _"It may be necessary to iterate over a collection without producing a new value"_, this led me to waste a few seconds searching for the classic `map` function (this is what the majority of developers be familiar with). 
Since `each` is a less common function, I assumed that I had accidentally missed the `map` function.


kinson (commented):
@doomspork is this worth bumping the version for?

PR #3037, opened 2025-02-27
docs: add more info to explain reduce without given accumulator

mrsofiane:


kinson (approved):
Nice! This is a help improvement to the explanation! Thanks for taking the time to contribute it 🫶

PR #3038, opened 2025-03-06
Update modules.md

randallk:
fix typo


kinson (approved):
Good catch! Thanks for the pr 💥

PR #3044, opened 2025-05-09
Update otp_distribution.md

julianvargasalvarez:
The documentation says "# app/chat.ex" but it should be "# lib/chat.ex"


kinson (approved):
Thanks for the correction 🚀

PR #3050, opened 2025-06-25
Update basic.md

PiSoup:
Updated example of integer division to one which differs from standard float division.


doomspork (approved):
Thank you @PiSoup! We should apply this to the all the translations as well. Would you like to open a separate PR for that or have me handle it?

doomspork:
Done in #3051! Thanks again @PiSoup

PR #3052, opened 2025-06-27
Clear confusion.

acburigo:
I believe this will clear some confusion to what's happening here.

### DIFF
--- a/lessons/en/basics/functions.md
+++ b/lessons/en/basics/functions.md
@@ -402,9 +402,9 @@ defmodule Greeter do
   def hello(names, language_code \\ "en")
 
   def hello(names, language_code) when is_list(names) do
-    binary_names = Enum.join(names, ", ")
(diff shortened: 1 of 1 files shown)


lessons/en/basics/functions.md
```diff
@@ -402,9 +402,9 @@ defmodule Greeter do
   def hello(names, language_code \\ "en")
 
   def hello(names, language_code) when is_list(names) do
-    names = Enum.join(names, ", ")
+    binary_names = Enum.join(names, ", ")
```
doomspork, line 405: I think this is a good change overall to avoid reassigning `names` but I'm not sure `binary_names` is actually much clearer, maybe we can go with something a bit more verbose but exact? `comma_separated_names`? `joined_names`?

doomspork (approved):
Thank you @acburigo!

PR #3061, opened 2025-08-15
Update link to strict comparison

0bvim:
- Fix path in all languages for `strict comparison` redirection.


0bvim:
> This is fantastic, thank you @0bvim! Let me pull just to make sure things work as expected. Then we should see about applying throughout the lessons.

I see that have another links that may be broken too, but I focus on solve this ones first!
I checked sometimes to ensure that it's working properly, because names are quite different through languages!

PR #3067, opened 2025-11-01
Add better Map.put description to collections lesson

### LINKED ISSUE #3016: Can we add a block explaining what happens when we use `put` in a map for an existing key?
https://elixirschool.com/en/lessons/basics/collections

I saw that there is a code block explaining what happens when you try to update a map with a non-existing key, and why a put operator should be used instead. (See image below)

I would like to propose that we also add a caveat saying what happens when `put` is used with an existing key.
Does it raise an error? Does it process the code just as an update statement would? Let's clarify.

P.S. reference to context

image

brain-geek:
I guess this fixes #3016 .

### DIFF
--- a/lessons/en/basics/collections.md
+++ b/lessons/en/basics/collections.md
@@ -140,24 +140,26 @@ For these reasons, keyword lists are most commonly used to pass options to funct
 
 In Elixir, maps are the "go-to" key-value store.
 Unlike keyword lists, they allow keys of any type and are un-ordered.
-You can define a map with the `%{}` syntax:
+You can define a map with the `%{}` syntax, using defined keys or variables:
 
 ```elixir
 iex> map = %{:foo => "bar", "hello" => :world}
 %{:foo => "bar", "hello" => :world}
-iex> map[:foo]
-"bar"
-iex> map["hello"]
-:world
-```
-
-As of Elixir 1.2, variables are allowed as map keys:
-
-```elixir
 iex> key = "hello"
 "hello"
 iex> %{key => "world"}
 %{"hello" => "world"}
+```
+
+There are also multiple ways of accessing map values:
+
+```elixir
+iex> map[:foo]
+"bar"
+iex> map["hello"]
+:world
+iex> Map.get(map, :foo)
+"bar"
 ```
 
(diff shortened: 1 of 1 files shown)


lessons/en/basics/collections.md
```diff
@@ -196,7 +196,7 @@ iex> %{map | foo: "baz"}
 
 **Note**: this syntax only works for updating a key that already exists in the map! If the key does not exist, a `KeyError` will be raised.
 
-To create a new key, instead use [`Map.put/3`](https://hexdocs.pm/elixir/Map.html#put/3)
+To create a new key, instead use [`Map.put/3`](https://hexdocs.pm/elixir/Map.html#put/3). They are equivalent, however the [`Map.put/3`](https://hexdocs.pm/elixir/Map.html#put/3) adds the new key even if it did not exist in the original map. As seen in this example:
```
doomspork: If I'm understanding the issue correctly they're looking for something like this:

```suggestion
To add a new key we can instead use [`Map.put/3`](https://hexdocs.pm/elixir/Map.html#put/3), this adds a new key if one does not exist and updates the record if a value exists at that key already.  We can see these behaviors demonstrated in our example:
```

Then maybe we update the example:

```elixir
iex> map = %{hello: "world"}
%{hello: "world"}
# Try updating our map with a new `:foo` key using the `|` method
iex> %{map | foo: "baz"}
** (KeyError) key :foo not found in: %{hello: "world"}
    (stdlib) :maps.update(:foo, "baz", %{hello: "world"})
    (stdlib) erl_eval.erl:259: anonymous fn/2 in :erl_eval.expr/5
    (stdlib) lists.erl:1263: :lists.foldl/3
# Use `Map.put/3` to add our new key and value
iex> Map.put(map, :foo, "baz")
%{foo: "baz", hello: "world"}
# Use `Map.put/3` to update our key
iex> Map.put(map, :foo, "bar")
%{foo: "bar", hello: "world"}
```

Thoughts @brain-geek @kinson?

brain-geek:
After rereading the lesson, I've also added two more changes:
- dropping the deprecated behavior with duplicate keys (in current elixir it's a warning `warning: key :foo will be overridden in map`)
- added `Map.get` example and reshuffled order of it with map definitions

PR #3091, opened 2026-03-31
Update dependencies in mix.exs for Elixir project

williycole:
Was going through this module and ran into some issues. Thought it good to update these accordingly.


doomspork (approved):
This is great @williycole! Would you be open to seeing if any other translations need these updates as well? That can be a separate PR.

williycole:
Absolutely @doomspork, I'll keep an eye out.

PR #3099, opened 2026-07-11
[ID] Update for all "Basics" section

widianto:


doomspork (commented):
Thank you @widianto!!

